PR #6, opened 2014-03-25
Allow users to provide own circuitbreakstate implementation / inject circuit state

lund:
Hi

I'd like to be able to use a circuitbreaker that is able to persist its state somewhere.
Thought this was the most framework'ish approach.

What do you think, does it follow the path that you would like Polly to take?

/martin

### DIFF
--- a/src/Polly.Net35/CircuitBreakerSyntax.cs
+++ b/src/Polly.Net35/CircuitBreakerSyntax.cs
@@ -28,10 +28,8 @@ namespace Polly
         /// <exception cref="System.ArgumentOutOfRangeException">exceptionsAllowedBeforeBreaking;Value must be greater than zero.</exception>
         public static Policy CircuitBreaker(this PolicyBuilder policyBuilder, int exceptionsAllowedBeforeBreaking, TimeSpan durationOfBreak)
         {
-            if (exceptionsAllowedBeforeBreaking <= 0) throw new ArgumentOutOfRangeException("exceptionsAllowedBeforeBreaking", "Value must be greater than zero.");
-
             var policyState = new CircuitBreakerState(exceptionsAllowedBeforeBreaking, durationOfBreak);
-            return new Policy(action => CircuitBreakerPolicy.Implementation(action, policyBuilder.ExceptionPredicates, policyState));
+            return CircuitBreaker(policyBuilder, exceptionsAllowedBeforeBreaking, policyState);
         }
 
         /// <summary>
--- a/src/Polly.Net35/CircuitBreakerSyntax.cs
+++ b/src/Polly.Net35/CircuitBreakerSyntax.cs
@@ -33,5 +33,23 @@ namespace Polly
             var policyState = new CircuitBreakerState(exceptionsAllowedBeforeBreaking, durationOfBreak);
             return new Policy(action => CircuitBreakerPolicy.Implementation(action, policyBuilder.ExceptionPredicates, policyState));
         }
+
+        /// <summary>
+        /// <para> Builds a <see cref="Policy"/> that will function like a Circuit Breaker.</para>
+        /// <para>The circuit will break after <paramref name="exceptionsAllowedBeforeBreaking"/>
+        /// exceptions that are handled by this policy are raised.</para>
+        /// </summary>
+        /// <param name="policyBuilder">The policy builder.</param>
+        /// <param name="exceptionsAllowedBeforeBreaking">The number of exceptions that are allowed before opening the circuit.</param>
+        /// <param name="circuitBreakerState">Use to provide own implementation of the circuitbreaker state.</param>
+        /// <returns>The policy instance.</returns>
+        /// <remarks>(see "Release It!" by Michael T. Nygard fi)</remarks>
+        /// <exception cref="System.ArgumentOutOfRangeException">exceptionsAllowedBeforeBreaking;Value must be greater than zero.</exception>
+        public static Policy CircuitBreaker(this PolicyBuilder policyBuilder, int exceptionsAllowedBeforeBreaking, ICircuitBreakerState circuitBreakerState)
+        {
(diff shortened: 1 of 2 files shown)


src/Polly.Net35/CircuitBreakerSyntax.cs
```diff
@@ -28,10 +28,26 @@ public static class CircuitBreakerSyntax
...
+        /// <para>The circuit will break after <paramref name="exceptionsAllowedBeforeBreaking"/>
+        /// exceptions that are handled by this policy are raised.</para>
+        /// </summary>
+        /// <param name="policyBuilder">The policy builder.</param>
+        /// <param name="exceptionsAllowedBeforeBreaking">The number of exceptions that are allowed before opening the circuit.</param>
+        /// <param name="circuitBreakerState">Use to provide own implementation of the circuitbreaker state.</param>
+        /// <returns>The policy instance.</returns>
+        /// <remarks>(see "Release It!" by Michael T. Nygard fi)</remarks>
+        /// <exception cref="System.ArgumentOutOfRangeException">exceptionsAllowedBeforeBreaking;Value must be greater than zero.</exception>
+        public static Policy CircuitBreaker(this PolicyBuilder policyBuilder, int exceptionsAllowedBeforeBreaking, ICircuitBreakerState circuitBreakerState)
```
bcuff: Maybe `exceptionsAllowedBeforeBreaking` should be a parameter to the `circuitBreakerState` implementation. You may want an implementation that requires a success ratio for example. In that case the exception count would be meaningless.

bcuff:
:+1: I hope this gets merge. I would like this exposed as well.

bunceg:
Ditto +1

reisenberger:
@lund @bunceg @bcuff / for anyone reading this PR at a later date:

While Polly has not at this point gone the route of allowing people to inject circuit-breaker state (see also the discussion in #41 on merits/de-merits of this), 

+ more manual control over the Circuit-Breaker has been provided in Polly v4.0; 
+ and the [Roadmap](https://github.com/App-vNext/Polly/wiki/Polly-Roadmap) proposes future Circuit-Breaker types which break on different statistics.

joelhulen:
Bouncing off of what @reisenberger, if you guys can review the latest changes and look at the roadmap for proposed circuit breaker modifications and comment on [this sticky issue](https://github.com/App-vNext/Polly/issues/90), that would help open up the dialog for others in context to the current state of Polly.

reisenberger:
Hi @bunceg, coming to the discussion from the other thread:

>I take on board all the comments re: negotiation between services and what if connection to the 
>shared state handler goes down etc. but I'm taking the pessimistic view that if a service the 
>breaker is protecting is dead then it's more likely that the service is really dead rather than the 
>alternative that the service is alive but one instance can't seem to talk to it.

>Besides, IMO, that can be some kind of "quorum" logic - two instances, two need to vote dead etc. 
>More complex, but this can be part of the state handler to be developed by the poor soul (i.e. me!) 
>who desires this kind of functionality...

Re the scenario:

>>Two instances in their own processes, each has an instance of circuit breaker. Instance A trips 
>>at 00:00 and lasts for 5 minutes.. Due to random load balancing, most requests just happen to 
>> go to instance A. However, by chance, at 04:50 instance B gets some requests and also trips. 
>>it's 5 minute trip timer now starts. However, the tripped service is back at 05:00. The instance 
>>A breaker is cleared at 05:01 but the instance B breaker is not able to be cleared until 09:50,
>> 4+ minutes after when it should be able to serve requests.

The manual controls `.Isolate()` and `.Reset()` added in Polly v4 (designed to allow greater control over isolation of downstream systems) could be used.  Whenever your co-ordination system (by whatever logic deemed quorate) decides that the breaker on instance B can be cleared, it could invoke `.Reset()` on the instance B breaker straight away - no need to wait.

Similarly, on the breaking side of the process, if A (plus however many other services deemed quorate) determine that some subsystem should be marked unavailable for all, you could (at present) call `.Isolate()` on breakers used elsewhere over that subsystem, and later `.Reset()` them when your co-ordination system deems it is safe to do so.

Per previous discussion (#41), there are arguments for and against maintaining such a cross-process-breaker-state co-ordination system, but your line of reasoning (effectively to trust the co-ordination between local services more than the link to a certain downstream service) makes a reasonable argument for this.  And, clearly the approach outlined above is not as straightforward as simply injecting state (tho this would also have other implications).  _Interested to hear other community views, and keeping this open for review_.

reisenberger:
@bunceg On a practical level also, I would like to resolve the future direction for the circuit-breaker further (frequency-based? proportion-based? hystrix-like?) (ref #41 and #90) before making the internal state public eg to be injected.  The internal representation of circuit-breaker state has already evolved over the last few months, and may evolve yet further as new circuit-breaker variants are introduced.  I wouldn't want to make the internal state public and 'lock us in' at this point, and/or provide an API which we then only break for you later.  

This is not to say that Polly might not at some future date expose the internal breaker state for injection - can see how this would help your use case and possibly others.  Just that it's not next up on the roadmap.  Sharing this wider roadmap view to keep you informed (hope useful).  Hope that the v4 features are in the meantime useful to your use case.

bunceg:
Hi,

Happy to wait for a wider discussion/direction as this makes sense if some hystrix logic may be ported. Circuit breaker is probably the most useful, (and most complex :) of the polly features) so it makes sense to think this through carefully.

For my immediate need, I am working around the issue anyway now by putting my own coordinator in front of the Polly handler (i.e. coordinator think it's tripped just throw a BrokenCircuit without even checking Polly). Isolate and Reset work ok as a "master switch" but not really in my scenario.

reisenberger:
Closing this old PR as this will not be the approach to adding this to Polly.  On-going discussion of allowing circuit-breakers to use external state is in #215 .  Referencing this from #215, for anyone who wants to read the discussion here.

PR #8, opened 2014-03-28
Added Noop policy method

NielsKlosterAndreassen:
In a large application it is sometimes nice to be able to pass exception policies around. When doing so, it is necessary to have a policy equivalent to not wrapping the call in a policy. 

This is my first public pull request ever, so comments and critique is welcome.

### DIFF
--- a/src/Polly.Net35/Policy.cs
+++ b/src/Polly.Net35/Policy.cs
@@ -68,5 +68,14 @@ namespace Polly
 
             return new PolicyBuilder(predicate);
         }
+
+        /// <summary>
+        /// Builds a void policy without behaviour.
+        /// </summary>
+        /// <returns>A policy instance</returns>
+        public static Policy Empty()
+        {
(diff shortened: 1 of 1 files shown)


bchavez:
:+1: for this

bchavez:
This seems to work just as well:

```
    public static class Safely
    {
        public static PolicyBuilder HandleAny()
        {
            return Policy.Handle<Exception>();
        }
    }
```
:+1:

reisenberger:
@NielsKlosterAndreassen @mattgwagner  The AppvNext team have now taken on the stewardship of Polly.  Although this PR is old, just updating you that this (being able to specify no-retry as a `Policy`) seems a reasonable idea, that several people have asked for.  The retry-count-of-zero approach is preferred for this and we should be able to merge this in due course.  See #37 if you want to follow / help implement this!

reisenberger:
Closing this PR in favour of #37

PR #16, opened 2014-12-09
Add handle by typeof

feanz:
Add method to allow you to handle policy by Type so you can pass in typeof().  This will make it easier for consuming libraries to configure what exceptions they are want to handle at run time based on config.

I've added some test coverage but did not want to repeat all tests for both generic and non generic methods.

### DIFF
--- a/src/Polly.Net35/OrSyntax.cs
+++ b/src/Polly.Net35/OrSyntax.cs
@@ -11,11 +11,40 @@ namespace Polly
         /// Specifies the type of exception that this policy can handle.
         /// </summary>
         /// <typeparam name="TException">The type of the exception to handle.</typeparam>
-        /// <param name="policyBuilder">The current builder to chain off.</param>
+        /// <param name="policyBuilder"></param>
         /// <returns>The PolicyBuilder instance.</returns>
         public static PolicyBuilder Or<TException>(this PolicyBuilder policyBuilder) where TException : Exception
         {
             ExceptionPredicate predicate = exception => exception is TException;
+            policyBuilder.ExceptionPredicates.Add(predicate);
+            return policyBuilder;
+        }
+
+        /// <summary>
+        /// Specifies the type of exception that this policy can handle.
+        /// </summary>
+        /// <param name="policyBuilder">The current builder to chain off.</param>
+        /// <param name="expectedException">The type of the exception to handle.</param>
+        /// <returns>The PolicyBuilder instance.</returns>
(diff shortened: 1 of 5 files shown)


feanz:
@michael-wolfenden was there an issue with this PR?

CreepyGnome:
I know this is old, @reisenberger ... Just curious what happened to this request? You still do not have this feature and I and maybe others could really use it. 

Basically to provide signatures like these:

```csharp
public static PolicyBuilder Handle(Type exception);
public static PolicyBuilder Handle(Type exception, Func<Exception, bool> exceptionPredicate);
public PolicyBuilder Or(Type exception);
public PolicyBuilder Or(Type exception, Func<Exception, bool> exceptionPredicate);
```

_(Also extend this idea to related methods)_

This will allow the ability to dynamically at runtime build policies. Currently, since there are only generic forms the types have to be known at compile-time.

**A related note:**
What is confusing is you have a doc titled "Why does Polly offer both non-generic and generic policies?" https://github.com/App-vNext/Polly/wiki/Non-generic-and-generic-policies

This doc states that "Polly offers non-generic policies" However in that example, it shows 3 generic methods on Policy.  It talks about RetryPolicy and CircuitBreakerPolicy, etc that extend this base Policy as being generic. However, they are not. They still only have generic handle methods and provide nothing new really over the pays Policy in most cases. (eg: RetryPolicy.

This document seems to be misleading as there is no way to define a policy with calling methods that have generic definitions.

reisenberger:
@CreepyGnome I would suggest solving this using some short extension methods in your own code:

```csharp
public static class PolicyExtensions
{
    public static PolicyBuilder Handle(Type exceptionType) => Policy.Handle<Exception>(e => exceptionType.IsInstanceOfType(e));
    public static PolicyBuilder Handle(Type exceptionType, Func<Exception, bool> exceptionPredicate) => Policy.Handle<Exception>(e => exceptionType.IsInstanceOfType(e) && exceptionPredicate(e));
    /// (and similarly for other variants)
}
```

CreepyGnome:
@reisenberger I started down that path and abandoned it but at this time I forget what blocked it as in project time that was a sprint ago and we moved on. I do not remember trying to incorporate the Type.IsInstanceOfType call into it though. Since I worked around it for now, when it comes back up I can revisit it and see if that is what I was missing.

Thanks for the insight. I do hope you add in officially non-generic Handle methods as that would be great all around. ;)

Issue #34, opened 2015-05-28
Policy ExecuteAsync method hangs if .Result is used.


hacko-bede:
Hey there, first of all thanks for the great work that you have been made

We created a execution policy solution based on Polly and we face following problem. When we call Task<T>.Result the thread hangs. We made an investigation over the code and the conclusion is that the problem is in the implementation of Polly. We Fork the project and change the implementation of `PoicyAsync.cs` on line 45.
Original:
```C#
 await _asyncExceptionPolicy(async () => { result = await action().NotOnCapturedContext(); });
```
With this one:
```C#
 await _asyncExceptionPolicy(async () => { result = await action().NotOnCapturedContext(); }).NotOnCapturedContext();
```
And this fixed the issue for us.

Cheers,
Hacko

michael-wolfenden:
Hey Hacko,

Thanks for the issue, your suggestion is a good one and I'll add it when I get a chance.

I'm curious though about why your calling `.Result`. This means your blocking on the asynchronous call which can cause deadlocks (http://blog.stephencleary.com/2012/07/dont-block-on-async-code.html).

The solution is to not mix synchronous and asynchronous code although I know in some situations this is avoidable.

Are you able to post some sample code so I can get a better context on what your trying to achieve?

hacko-bede:
Hey again,

Well short answer we don't have a choice. For now part of our system is still now async ready and we should use .Result.

Thanks for your fast response and collaboration.

Cheers

PR #36, opened 2015-06-02
Added NotOnCapturedContext in PolicyAsync

hacko-bede:
This is done in order not to block threads then someone calls .Result


michael-wolfenden:
Much appreciated. Will merge, update the release notes and push out a new nuget package.

This closes #34.

PR #39, opened 2015-07-01
Added AspNet50 Support with a new class library project

ThreeSevenths:
Added ASP.net 5 project to the solution with support for the dnx and dnxcore frameworks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,7 @@ Backup*/
 UpgradeLog*.XML
 
 build_artifacts
+
+#ASPNet5 Project related stuff
+artifacts
+*.lock.json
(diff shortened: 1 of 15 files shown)


ericis:
*bump*  @ThreeSevenths AppVeyor doesn't support .net 5 projects yet, right?  Should this sit in a branch until they're out of beta?

michael-wolfenden:
Thanks for the contribution, but I want to hold out on adding ASP.NET 5 support until it goes RTM as the platform seems to be in a lot of flux at the moment.

ThreeSevenths:
@ericis - It failed on my machine as well. Problem seems to be msbuild not liking the new targets for aspnet5 by default. Hopefully they will fix with RTM.

@michael-wolfenden - Good call.

PR #40, opened 2015-07-06
Post execute functionalities

ThomasMentzel:
We started using Polly but required "post execution steps" if retry did not succeed. Basically we have a "Save()" method which either throws a "RepositoryException" or any other exception. On a repository exception we would like to retry the save (e.g. a network issue happened) and if this fail we would like to show a message box. On another exception (like a schema exception) we would like to keep the rethrow mechanism. The application became unusable and a "hard crash" is preferred.

This pull request adds a separate execute method 'ExecuteAnd' which returns a HandledPolicy which might contain an exception. The HandledPolicy can be used to create custom extension methods (like we did with the message box or already has some predefined generic methods. The ExecuteAnd method still throws an exception if the exception is not registered as "expected to retry".

Due to the fact that we need and use this functionality, we hope to get this added to polly. Otherwise we need to keep the fork and create our own "PollyPost" package.


michael-wolfenden:
Thomas,

Thanks for your contribution. My ai mis always to keep the surface area of the api as small as possible.

Are you able to implement this via extensions methods rather than modifying the existing solution?

For example 

```csharp
public static HandledPolicy ExecuteAnd(this Policy policy, Action action) {
	try
	{
		policy.Execute(action);
		return new HandledPolicy(null);
	}
	catch (Exception ex)
	{
		return new HandledPolicy(ex);
	}
}
```

ThomasMentzel:
I tried several ways to get this implemented via extension methods but I did not found any solution. My problem was the definition of the exceptions to handle. For the "post processing" I need the exception predicates to keep the "throw if not handled" mechanism. Because if I do a retry on a "ConnectionException" and after 10 retries, the connect did not work, Execute() throws the exception. I need to throw the exception if it was not handled and post-handle the exception if fulfills the retry predicates.

I tried to keep the footprint as minimal as possible but unfortunately I needed to forward the exception predicates. https://github.com/michael-wolfenden/Polly/pull/40/files#diff-9c3ec5de43339ca9d19197ca30b8ef17R78 . So I made the policy contain the predicated to access these after execute.

michael-wolfenden:
@ThomasMentzel I'm going to merge your request, however I am going to make a few changes. To keep the surface area of the api small, I am going to remove the extension methods as you can add them in your own project. I will also make some naming changes. Hopefully I can get a new nuget package out today, if not in the next couple of days. Thanks again for your contribution.

PR #43, opened 2015-08-10
Fix typo in readme

benjamin-hodgson:


michael-wolfenden:
Thanks @benjamin-hodgson

PR #48, opened 2015-10-21
Support for .NET 4.6

manastalukdar:
Support for .NET 4.6: (1) Added projects. (2) Updated nuspec and readme text.


michael-wolfenden:
@manastalukdar Thanks for the pull request however given that no 4.6 features are used there is no no benefit in providing a NET 4.6 library and hence maintaining another project.

Issue #46, opened 2015-08-24
Add ability to cancel execution


nedstoyanov:
It may be handy to have ability to cancel execution for example, via a `CancelationToken`

reisenberger:
@nedstoyanov In #56 I have another issue that might move things towards cancellation support (tho it's quite a recast of the Polly retry architecture).

We also think cancellation support would be a valuable addition. Exponential back-off retry (eg after 1, 2, 4 seconds; or even 5, 10, 30) is wonderful, especially in a machine-to-machine/service-to-service environment, where eventual consistency is acceptable, and eventual success (even after quite generous delays) is far preferable to faults to be manually troubleshot later.  However, the lack of cancellation support makes it hard to conduct a graceful shutdown of such retries.

reisenberger:
@joelhulen Some starting thoughts / a possible spec around cancellation support. Keen to hear others' views.

[1] All types of asynchronous retry policy should support cancellation.  In respect of retries, cancellation should prevent further retries commencing, if the delegate being `ExecuteAsync()`d throws.  For `WaitAndRetryAsync()`, cancellation should also interrupt the `Task.Delay`.  All retry policies should also honour cancelled-before-first-try (I believe this is .NET framework approach - don't even start a `Task` if it's cancellation token is found cancelled, on entry)

[2] Should synchronous retry policies support a `CancellationToken`?  `CancellationToken` exists for .NET4, so cancellation could conceivably be added for .NET4 versions of _synchronous_ retry policies.  However, would lead to divergence in the synchronous API between .NET3.5 and .NET4+. Perhaps conceptually simpler (including for understanding of the product) to limit/align cancellation support to async?

[3] Specifying that a retry policy should observe a cancellation token, probably forms part of the retry overload signature?, eg

`var cancellableRetryPolicy = Policy.Handle<WhateverException>.Retry(3, myCancellationToken)`

[4] `action`s being `ExecuteAsync()`d obviously often support a `CancellationToken` too.  But there's possibly no need for Polly to get involved with that (eg with extra `ExecuteAsync()` overloads), can leave cancellation tokens to be captured by implicit closure, to reduce Polly interface surface, eg:

```
CancellationToken cancellationToken = ... 
var retryPolicy = Policy.Handle<Whatever>.WaitAndRetryAsync(new[] {
                   1.Seconds(),
                   2.Seconds(),
                   4.Seconds()
                }, cancellationToken);
[...]

PR #53, opened 2015-11-16
Async sleep + continue on captured ctx. Fixes #49

### LINKED ISSUE #49: Running async actions on captured synchronization context
Runnng Polly inside MS Orleans grain is [not](https://github.com/dotnet/orleans/pull/875#issuecomment-145917957) possible, since Orleans uses custom TaskScheduler and due to automatically added `ConfigureAwait(false)` action escapes it and runs on different thread, thus breaking single-threaded guarantee provided by Orleans. 

Is that possible to add an overload for `RetryAsync\WaitRetryAsync` that will allow to control that behavior? 

P.S. I'll be happy to send a PR but I have no idea what feature/project type my VS15 installation is missing, that I cannot load Polly.Shared projects. It tell me that `The imported project "C:\Program Files (x86)\MSBuild\Microsoft\WindowsXaml\v14.0\8.1\Microsoft.Windows.UI.Xaml.CSharp.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.  C:\Program Files (x86)\MSBuild\Microsoft\WindowsXaml\v14.0\Microsoft.Windows.UI.Xaml.CSharp.targets` :worried: 

Is that ExpressionDesigner that need to be installed?

yevhen:


yevhen:
@michael-wolfenden any updates on this?

reisenberger:
@yevhen Looking at your PR, I see (only now...) that you have also tackled async sleep as `await Task.Delay` as part of this PR on captured ctx, same as I have raised in #56.  I have done it slightly differently, but same underlying principle.  Will be happy with whatever Polly community want to integrate. Think this is an important fix.

yevhen:
@reisenberger ye, think the same!

src/Polly.Shared/RetrySyntaxAsync.cs
```diff
@@ -19,6 +19,17 @@ public static partial class RetrySyntax
...
+
+        /// <summary>
+        /// Builds a <see cref="Policy"/> that will retry once.
+        /// </summary>
+        /// <param name="policyBuilder">The policy builder.</param>
+        /// <param name="continueOnCapturedContext">Whether to continue on a captured synchronization context.</param>
+        /// <returns>The policy instance.</returns>
+        public static Policy RetryAsync(this PolicyBuilder policyBuilder, bool continueOnCapturedContext)
+        {
             return policyBuilder.RetryAsync(1);
```
reisenberger: @yevhen This overload `RetryAsync(this PolicyBuilder policyBuilder, bool continueOnCapturedContext)` doesn't make use of the supplied `continueOnCapturedContext` value.  Does it want a small correction to:
`return policyBuilder.RetryAsync(1, continueOnCapturedContext);` ?

(Apologies @yevhen @joelhulen for intervening but hope useful - I spotted this while doing some exploratory investigation of how to add cancellation support!)

joelhulen: @reisenberger No apology needed! It takes volunteers like you to help spot these things and keep progress marching forward. All help is gladly accepted :)

yevhen: @joelhulen +1 @reisenberger nice catch! I was trying really hard to not introduce breaking changes and got lost with all those pesky overloads. Many thanks and you don't need to apologize for that. It's great to have another pair of eyes! :eyes:

reisenberger:
@yevhen @joelhulen On the subject of all those overloads (agree), it might be worth some forethought - if we are going to introduce both `bool continueOnCapturedContext`, and `CancellationToken cancellationToken` (ref #46) - on what priority order we want the parameters to run in.

In the overloads in this #53, @yevhen @joelhulen (see snip below) ... 

image

... I like the order in `RetryAsync(...)` and `RetryForeverAsync(...)` (it seems logical that `bool continueOnCapturedContext` comes after how many retries but before `Action onRetry` ... ).  But, for `WaitAndRetryAsync(...)` I would suggest `bool continueOnCapturedContext` should come _after_ `Func<int, TimeSpan> sleepDurationProvider` or `IEnumerable<TimeSpan> sleepDurations`.  The how-long-to-delay-for is core to expressing `WaitAndRetryAsync(...)` and should come right after the `(`.  This would also place overload parameters in a consistent order for all retry overloads:

```
how many times / when to retry,
whether to continue on captured context,
action on retry
```

If/when introducing `CancellationToken` could then be expanded to:

```
how many times / when to retry,
cancellation token,
whether to continue on captured context,
action on retry
```

(or cancellation token / captured context in the other order)

yevhen:
Very good suggestion. I will update PR.

re order of CancellationToken/ContinueOnCapturedContext I think any order is fine.

The more important is to figure out what to do with all those overloads? If we introduce another set of overloads (permutation with `CancellationToken`) we may end up with unmaintainable mess. I think that number of parameters is already far behind the number of parameters acceptable for Clean Code :stuck_out_tongue_winking_eye: 

What if we introduce something like Parameter Object? Alternatively we can try to employ automation, say T4 template which will auto-generate overloads with all that currently copy-n-pasted xml documentation.

reisenberger:
@yevhen Thanks for this great contribution; we're looking to get this merged soon.  

Re overloads I can see three options, not sure if any quite come together (tho if anyone can see a bright way through this conundrum, please shout)

**Optional parameters**: Would be ideal (my preferred) if it weren't for the caller/call-ee problem in how this is implemented in C#.  IE if we later have to change signature, can cause run-time failure if people just drop in new DLL rather than recompile against.  Believe why most of .NET framework doesn't use optional params.

**T4**: Slight concern about introducing something lesser-known which may raise the barrier of entry for other contributors.  Also, does this just help with documentation (eg xml comments)?

**Parameter object**: Problem I can see is there's already the optional `onRetry` parameter.  If we move that into a parameter object, it'd be a breaking change :cry:.  If we don't move it in to a parameter object, some of the optional parameters will be in a parameter object and some not  :disappointed: .

Also, `.RetryAsync(int, new RetryAsyncOptions(CancellationToken, bool), Action)` not necessarily more concise when calling than `.RetryAsync(int, CancellationToken, bool, Action)`
.
Open to views, but it may be the overloads are just something we're going to have to live with, rather like `Task.Factory.StartNew()` in the framework

reisenberger:
@yevhen Thanks again for this important PR!  _Unless_ anyone has brighter idea on overloads, team AppvNext are ready to accept/merge.  

Are you able to update the PR with the following minor adjustments? - then it will probably be good to go!

+ resolve conflicts by rebasing against latest master
+ fix overload not using `bool continueOnCapturedContext` mentioned few days ago
+ adjust `bool continueOnCapturedContext` parameter position per discussion few days ago and a few more line notes I will add
+ update changelog.MD to describe change (but not GitVersionConfig.yaml; @joelhulen will when pushing to Nuget)

Thank you!

src/Polly.Shared/CircuitBreaker/CircuitBreakerPolicyAsync.cs
```diff
@@ -10,7 +10,7 @@ namespace Polly.CircuitBreaker
 {
     internal partial class CircuitBreakerPolicy
     {
-        internal static async Task ImplementationAsync(Func<Task> action, IEnumerable<ExceptionPredicate> shouldRetryPredicates, ICircuitBreakerState breakerState)
+        internal static async Task ImplementationAsync(bool continueOnCapturedContext, Func<Task> action, IEnumerable<ExceptionPredicate> shouldRetryPredicates, ICircuitBreakerState breakerState)
```
reisenberger: Request move `bool continueOnCapturedContext` to be last parameter, in `ImplementationAsync`

yevhen:
If you're ready to publish new Nuget package then I'm ready to update this
in few days )))

2015-12-10 1:26 GMT+02:00 reisenberger <notifications@github.com>:

src/Polly.Shared/CircuitBreakerSyntaxAsync.cs
```diff
@@ -25,16 +25,17 @@ public static class CircuitBreakerSyntaxAsync
         /// <param name="policyBuilder">The policy builder.</param>
         /// <param name="exceptionsAllowedBeforeBreaking">The number of exceptions that are allowed before opening the circuit.</param>
         /// <param name="durationOfBreak">The duration the circuit will stay open before resetting.</param>
+        /// <param name="continueOnCapturedContext">Whether to continue on a captured synchronization context.</param>
         /// <returns>The policy instance.</returns>
         /// <remarks>(see "Release It!" by Michael T. Nygard fi)</remarks>
         /// <exception cref="System.ArgumentOutOfRangeException">exceptionsAllowedBeforeBreaking;Value must be greater than zero.</exception>
-        public static Policy CircuitBreakerAsync(this PolicyBuilder policyBuilder, int exceptionsAllowedBeforeBreaking, TimeSpan durationOfBreak)
+        public static Policy CircuitBreakerAsync(this PolicyBuilder policyBuilder, int exceptionsAllowedBeforeBreaking, TimeSpan durationOfBreak, bool continueOnCapturedContext = false)
```
reisenberger: Request add overload, not optional parameter, because of optional-params-compiled-at-callee-not-caller problem in c# implementation of optional parameters

src/Polly.Shared/PolicyAsync.cs
```diff
@@ -28,10 +28,21 @@ internal Policy(Func<Func<Task>, Task> asyncExceptionPolicy, IEnumerable<Excepti
...
+            return ExecuteAsync(false, action);
+        }
+
+        /// <summary>
+        ///     Executes the specified asynchronous action within the policy.
+        /// </summary>
+        /// <param name="continueOnCapturedContext">Whether to continue on a captured synchronization context.</param>
+        /// <param name="action">The action to perform.</param>
+        [DebuggerStepThrough]
+        public Task ExecuteAsync(bool continueOnCapturedContext, Func<Task> action)
```
reisenberger: Request move `bool continueOnCapturedContext` to be last parameter, in `ExecuteAsync()`

src/Polly.Shared/PolicyAsync.cs
```diff
@@ -39,14 +50,25 @@ public Task ExecuteAsync(Func<Task> action)
...
+            return ExecuteAndCaptureAsync(false, action);
+        }
+
+        /// <summary>
+        ///     Executes the specified asynchronous action within the policy and returns the captured result.
+        /// </summary>
+        /// <param name="continueOnCapturedContext">Whether to continue on a captured synchronization context.</param>
+        /// <param name="action">The action to perform.</param>
+        [DebuggerStepThrough]
+        public async Task<PolicyResult> ExecuteAndCaptureAsync(bool continueOnCapturedContext, Func<Task> action)
```
reisenberger: Request move `bool continueOnCapturedContext` to be last parameter, in `ExecuteAndCaptureAsync()`

src/Polly.Shared/PolicyAsync.cs
```diff
@@ -62,24 +84,62 @@ public Task ExecuteAsync(Func<Task> action)
...
+
+        /// <summary>
+        ///     Executes the specified asynchronous action within the policy and returns the result.
+        /// </summary>
+        /// <typeparam name="TResult">The type of the result.</typeparam>
+        /// <param name="continueOnCapturedContext">Whether to continue on a captured synchronization context.</param>
+        /// <param name="action">The action to perform.</param>
+        /// <returns>The value returned by the action</returns>
+        [DebuggerStepThrough]
+        public async Task<TResult> ExecuteAsync<TResult>(bool continueOnCapturedContext, Func<Task<TResult>> action)
```
reisenberger: Request move `bool continueOnCapturedContext` to be last parameter, in `ExecuteAsync()`

src/Polly.Shared/PolicyAsync.cs
```diff
@@ -62,24 +84,62 @@ public Task ExecuteAsync(Func<Task> action)
...
+        /// </summary>
+        /// <typeparam name="TResult">The type of the result.</typeparam>
+        /// <param name="continueOnCapturedContext">Whether to continue on a captured synchronization context.</param>
+        /// <param name="action">The action to perform.</param>
+        /// <returns>
+        /// The value returned by the action
+        /// </returns>
+        /// <exception cref="System.InvalidOperationException">Please use the asynchronous RetryAsync, RetryForeverAsync, WaitAndRetryAsync or CircuitBreakerAsync methods when calling the asynchronous Execute method.</exception>
+        [DebuggerStepThrough]
+        public async Task<PolicyResult<TResult>> ExecuteAndCaptureAsync<TResult>(bool continueOnCapturedContext, Func<Task<TResult>> action)
```
reisenberger: Request move `bool continueOnCapturedContext` to be last parameter

joelhulen:
@yevhen Once I merge your PR (after your changes), then I'll test locally and publish the new NuGet package once verified.

src/Polly.Shared/RetrySyntaxAsync.cs
```diff
@@ -119,9 +198,23 @@ public static Policy RetryForeverAsync(this PolicyBuilder policyBuilder, Action<
...
+        /// <summary>
+        ///     Builds a <see cref="Policy" /> that will wait and retry as many times as there are provided
+        ///     <paramref name="sleepDurations" />
+        ///     On each retry, the duration to wait is the current <paramref name="sleepDurations" /> item.
+        /// </summary>
+        /// <param name="policyBuilder">The policy builder.</param>
+        /// <param name="continueOnCapturedContext">Whether to continue on a captured synchronization context.</param>
+        /// <param name="sleepDurations">The sleep durations to wait for on each retry.</param>
+        /// <returns>The policy instance.</returns>
+        public static Policy WaitAndRetryAsync(this PolicyBuilder policyBuilder, bool continueOnCapturedContext, IEnumerable<TimeSpan> sleepDurations)
```
reisenberger: Per previous discussion, move `bool continueOnCapturedContext` to after `sleepDurations`

src/Polly.Shared/RetrySyntaxAsync.cs
```diff
@@ -135,9 +228,24 @@ public static Policy WaitAndRetryAsync(this PolicyBuilder policyBuilder, IEnumer
...
+        ///     Builds a <see cref="Policy"/> that will wait and retry <paramref name="retryCount"/> times.
+        ///     On each retry, the duration to wait is calculated by calling <paramref name="sleepDurationProvider"/> with
+        ///     the current retry attempt allowing an exponentially increasing wait time (exponential backoff).
+        /// </summary>
+        /// <param name="policyBuilder">The policy builder.</param>
+        /// <param name="retryCount">The retry count.</param>
+        /// <param name="continueOnCapturedContext">Whether to continue on a captured synchronization context.</param>
+        /// <param name="sleepDurationProvider">The function that provides the duration to wait for for a particular retry attempt.</param>
+        /// <returns>The policy instance.</returns>
+        public static Policy WaitAndRetryAsync(this PolicyBuilder policyBuilder, int retryCount, bool continueOnCapturedContext, Func<int, TimeSpan> sleepDurationProvider)
```
reisenberger: Per previous discussion, move `bool continueOnCapturedContext` to after `sleepDurations`

src/Polly.Shared/RetrySyntaxAsync.cs
```diff
@@ -157,8 +265,30 @@ public static Policy WaitAndRetryAsync(this PolicyBuilder policyBuilder, int ret
...
+        /// <param name="sleepDurationProvider">The function that provides the duration to wait for for a particular retry attempt.</param>
+        /// <param name="onRetry">The action to call on each retry.</param>
+        /// <returns>The policy instance.</returns>
+        /// <exception cref="System.ArgumentOutOfRangeException">retryCount;Value must be greater than zero.</exception>
+        /// <exception cref="System.ArgumentNullException">
+        ///     timeSpanProvider
+        ///     or
+        ///     onRetry
+        /// </exception>
+        public static Policy WaitAndRetryAsync(this PolicyBuilder policyBuilder, int retryCount, bool continueOnCapturedContext, Func<int, TimeSpan> sleepDurationProvider, Action<Exception, TimeSpan> onRetry)
```
reisenberger: Per previous discussion, move `bool continueOnCapturedContext` to after `sleepDurations`

src/Polly.Shared/RetrySyntaxAsync.cs
```diff
@@ -193,8 +323,28 @@ public static Policy WaitAndRetryAsync(this PolicyBuilder policyBuilder, int ret
...
+        /// <param name="continueOnCapturedContext">Whether to continue on a captured synchronization context.</param>
+        /// <param name="sleepDurations">The sleep durations to wait for on each retry.</param>
+        /// <param name="onRetry">The action to call on each retry.</param>
+        /// <returns>The policy instance.</returns>
+        /// <exception cref="System.ArgumentNullException">
+        ///     sleepDurations
+        ///     or
+        ///     onRetry
+        /// </exception>
+        public static Policy WaitAndRetryAsync(this PolicyBuilder policyBuilder, bool continueOnCapturedContext, IEnumerable<TimeSpan> sleepDurations, Action<Exception, TimeSpan> onRetry)
```
reisenberger: (you guessed it - move bool to after `sleepDurations` before `onRetry`)

src/Polly.Shared/Retry/RetryPolicyStateWithCountAsync.cs
```diff
@@ -0,0 +1,19 @@
+﻿#if SUPPORTS_ASYNC
+
+using System;
+using System.Threading.Tasks;
+
+namespace Polly.Retry
+{
+    internal partial class RetryPolicyStateWithCount : IRetryPolicyState
+    {
+        static readonly Task<bool> Done = Task.FromResult(true);
```
reisenberger: Remove `Done` if unused?

src/Polly.Shared/Retry/RetryPolicyStateWithSleepAsync.cs
```diff
@@ -0,0 +1,27 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Polly.Utilities;
+
+namespace Polly.Retry
+{
+    internal partial class RetryPolicyStateWithSleep : IRetryPolicyState
+    {
+        static readonly Task<bool> Done = Task.FromResult(true);
```
reisenberger: Remove `Done` if unused?

reisenberger:
( @yevhen / all, apologies, and thanks, for bearing with flurry of line notes ) ( as you rightly say, yevhen, large number of overloads ...;)  

But purpose: If we fix parameter order appropriately now, the cancellation support #46 can be built right around this without any breaking changes to the API between releases

joelhulen:
@yevhen: We're ready to include your PR into version 3.0! This along with cancellation features @reisenberger is working on will really help make Polly even more mature. Thank you so much for helping us push the library forward.

If you are ready, please fix the parameter order, as outlined in detail by @reisenberger. Once that's done, we can merge it in.

yevhen:
Sorry, guys. I'll definitely do this till EOW. At the moment I'm incredibly busy, preparing for 2 public presentations this week.

joelhulen:
Thanks, @yevhen! I hope you know I wasn't trying to pressure you. It's a crazy a time of year for all of us.

yevhen:
No worries ))

2015-12-16 21:18 GMT+02:00 Joel Hulen <notifications@github.com>:

joelhulen:
@yevhen: Hope you had a great holiday weekend :)  Just wondering if you've had a chance to fix up those last few things? We're hoping to release a new version with your changes by year's end, if possible.

Also, just so you're aware, you'll need to rebase your local repo with the latest changes we incorporated yesterday. Thanks!

yevhen:
Hey, just recovered from all those holidays ))

Going to work on this tonight. Hope to finish it tomorrow. 

Stay tuned :)

joelhulen:
Looking forward to it! Thanks :)

reisenberger:
Hey both @yevhen @joelhulen .  While on this, I spotted a couple of awaits that possibly weren't yet configured with the new (great) bool you're adding yevhen (ie what context the continuation runs on).  I think it was the `await`s on the calls to `CanRetryAsync()` and `SleepAsync()`.  Do you agree that we want to control the continuation context there also? (ie propogate configureawait() control via the new bool to those calls also?) In order to configure the awaits _everywhere_, ie to make the context after the awaits _completely_ deterministic throughout the Polly library.  Or does it not matter?

Could there be a risk, for example, that a retry runs on a different context from the original try? (which might not help you MS Orleans scenario, yevhen?), if we weren't controlling  `CanRetryAsync()` and `SleepAsync()`?

I found a great plugin for ReSharper that helped me find these: https://github.com/aelij/ConfigureAwaitChecker - but you can find/install it via the PlugIns library on ReSharper of VStudio.
(I was using the same plug-in when exploring issues in the possible cancellation code)

Shout if I'm not making sense or you think this is wrong - in haste with this comment as I need to be elsewhere!

yevhen:
makes total sense for me. Any other places or those are the only one that need to be patched?

joelhulen:
Makes sense to me, too. I think that we want to add ConfigureAwaits(false) everywhere within the library, as you suggested. Keeping track of the original context is also important.

reisenberger:
@yevhen @joelhulen Super: it was my assumption that we would want to be completely deterministic about this across the library, but wanted to check with you guys.  

@yevhen : those two:
```
if (!(await policyState.CanRetryAsync(ex)))
/
await SystemClock.SleepAsync(currentTimeSpan);
```
... appear to the only two not currently controlling the synchonization context (`.ConfigureAwait()`) via the new `bool continueOnCapturedContext`.

yevhen:
guys, what do you have set for Git's core.autocrlf and what version of git do you have installed? I'm observing absolutely weird behavior with files being modified just after the clone, despite having core.autocrlf=false. I'm using Git v1.9.5.

reisenberger:
I've seen cr/lf oddness too but don't know the setting - over to @joelhulen ? - would be good to iron this out.

reisenberger:
Apologies - more thoughts coming on this:

Because of the way the codebase has evolved (see end of post), we now have / are working towards a number of constructs in the codebase like / effectively equivalent to:  [+]

```
if (continueOnCapturedContext)
   await action();
else
   await action().NotOnCapturedContext();
```

or

```
if (onCapturedContext)
   return await task;
else
   return await task.NotOnCapturedContext();
```

Can these not just be shortened to ```await task.ConfigureAwait(continueOnCapturedContext)``` ? 

(It could also have the advantage of explicitly configuring the await either way - it probably makes no practical difference in the current .NET implementation, but may be clearer to subsequent code readers what is going on.)


The history of this in the Polly library was:  (from memory- could be wrong on some detail) 

+ Async implementation was initially added, without (I think; cld be wrong) code controlling continuation context, ie plain `await Task` 
+ Contributors then logically requested the addition of `.ConfigureAwait(false)`, as is standard for library code, particularly if library code shouldn't care about sycnhronisation context.  This was abbreviated to nice extension methods `.NotOnCapturedContext(this Task task)` etc.
+ But since the point of Polly is to run delegates (wrapped in some way), we might sometimes need to control the synchonisation context the delegates run on - hence yevhen's contribution.  Hence we can see why built on the existing ```.NotOnCapturedContext()```, this leads to patterns like [+] above, but I'm wondering if we can just do the much more direct ```await task.ConfigureAwait(continueOnCapturedContext);``` ?

(with apologies again for late additional realisation ... what do you guys think?)

yevhen:
@reisenberger ye, NotOnCapturedContext() doesn't make much sense anymore. I'll incorporate this realization ;)

yevhen:
I do believe problems with line endings are due to use of .gitattributes file. I observed similar problems on my projects when this file was included. It needs to be deleted from source. It only creates more problems.

yevhen:
I've added all change requests as separate commits so it should be easy to review. If you want I can squash everything into single commit.

yevhen:
Hey, build failed for some weird reasons. Please kick a rebuild.

joelhulen:
@yevhen there were some build failures yesterday due to connectivity issues between appveyor and NuGet. I'll kick off another build.

joelhulen:
Build succeeded :)

yevhen:
Awesome! =)

yevhen:
Ah, I forgot to modify CHANGELOG. Tomorrow then.

joelhulen:
I can modify the CHANGELOG if you haven't already.

joelhulen:
@reisenberger @yevhen: I just posted instructions on fixing the line endings to our [wiki site](https://github.com/App-vNext/Polly/wiki/Stifle-the-excess-file-changes%3A-Fix-your-line-endings%21). See if these steps help. If not, feel free to update the page :)

reisenberger:
@yevhen Thanks for your exemplary clear presentation! (a standard for us to follow :smile: ).  Have been a good second pair of :eyes: and @joelhulen  looks good to go to me!  Many thanks yevhen for tracking all that detail...

yevhen:
That would be great, cause I'm out of city till 10th of January 😄

yevhen:
Thanks guys! It was a pleasure working with you on this PR. Keep it up!

reisenberger:
Thanks @yevhen .  Cancellation support (#46) coming shortly (I've been working on it in background), which should round out the async support in Polly nicely. :wink:

yevhen:
Awesome! That would be great addition!

Issue #59, opened 2015-12-04
Polly NuSpec is missing dependencies


onovotny:
.NET Core requires packages to specify their dependencies. It should be pretty easy to add as I've written a tool that helps.

https://github.com/onovotny/ReferenceGenerator

The docs are there but just add a ref to your project and it'll update the nuspec on build to include the right dependencies.

joelhulen:
@onovotny I won't be able to update this until this evening. If you feel so inclined, please update the NuSpec file, and submit a PR. That should help get this resolved sooner.

joelhulen:
@onovotny Appreciate the PR :)

One thing I just noticed is, even though the build was successful in AppVeyor, there was an error thrown by cake:

========================================
__AddDotNetReferencesToNuspecFile
========================================
Executing task: __AddDotNetReferencesToNuspecFile
C:\projects\polly\tools\Cake\Cake.exe : error RG001: project.lock.json is missing
At line:1 char:1
+ C:\projects\polly\tools\Cake\Cake.exe "build.cake" -target="Default" -configurat ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (error RG001: pr...json is missing:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
Finished executing task: __AddDotNetReferencesToNuspecFile


========================================
There was a new task added by your reference generator, I believe, that the above error refers to: https://github.com/App-vNext/Polly/blob/master/build.cake#L185

Are you familiar with the error above about "project.lock.json" missing?

onovotny:
yeah... you need to call nuget restore in the appveyor script first as that lock file should never be checked in.

Here's an example I use in Humanizer: 
https://github.com/Humanizr/Humanizer/blob/dev/appveyor.yml#L7-L12

Also, I think that cake task can be removed as the build will handle it the way it's setup as part of building the pcl csproj.

onovotny:
Ugh, this whole thing could have been a giant brain-fart/missing stuff in nuget.org. It appears to have been fine originally :(

Very sorry for the false alarm and no hard feelings if you want to revert my commits.

PR #60, opened 2015-12-04
Use NuSpec refgen to put correct dotnet targets

onovotny:
Addresses #59

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ TestResults
 *.suo
 *.user
 *.sln.docstates
+.vs/
 
 # Build results
 [Dd]ebug/
@@ -110,3 +111,7 @@ UpgradeLog*.XML
 artifacts
 build
 tools
+
+*.lock.json
+*.nuget.targets
+*.nuget.props
--- a/src/Polly.Pcl/Polly.Pcl.csproj
+++ b/src/Polly.Pcl/Polly.Pcl.csproj
@@ -38,6 +38,11 @@
     <AssemblyOriginatorKeyFile>..\Polly.Net45\Polly.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
   <ItemGroup>
+    <NuSpecFile Include="$(SolutionDir)Polly.nuspec">
+      <Visible>False</Visible>
+    </NuSpecFile>
+  </ItemGroup>
+  <ItemGroup>
     <Compile Include="..\GlobalAssemblyInfo.cs">
       <Link>Properties\GlobalAssemblyInfo.cs</Link>
     </Compile>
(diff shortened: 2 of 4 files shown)


onovotny:
Not sure about the cake build stuff, but this works...

joelhulen:
This is perfect. Thanks!

michael-wolfenden:
@onovotny I'm a bit confused. I thought I was updating the nuspec file as part of the build script using the ReferenceGenerator cli. For clarification was this incorrect?. 

In this pull request it looks like you've manually specified the dependendies in the nuspec file, but my understanding was that this is the whole point of ReferenceGenerator ie to automatically generate these so if you add a new dependency to your project, it will correctly be added to the nuspec.

Any clarification would be much appreciated.

onovotny:
So there's two issues, both bugs I think. The main issue is that the tool wasn't picking up the dependencies correctly. If you noticed, the dotnet section was blank, when in fact it shouldn't have been.

In theory, you should be able to run the tool without having it as a reference (though I'm always baffled by why people would want that as you're less shielded from cmd line arguments changes that are handled by having a matching targets file).

1. The tool looks for the presence of a `packages.config` or a `project.json` and got confused. As the scenarios I'd previously tested relied on one of those files being there, I'm guessing there was an issue with the tool running externally and not finding one.

2. There was a bug where `project.json` usage for a PCL (instead of net46/dnx/uwp, etc), didn't properly pick up the built-in system references. I've fixed that in v1.4 to work properly.

I'm still not 100% sure of behavior without a project.json or packages.config file.

As to why the nuspec was checked in with it, it simply seems easier than always fighting git as the tool updates the nuspec on every build. If there are no changes, then git doesn't care either. It's entirely possible to remove that dotnet section and have it created on the CI server.

Does that answer your questions?

onovotny:
Hmm...re-checking the source, I stand corrected. It *should* handle having no packages.config present. I'd have to recheck in the debugger what was going on as it should have worked as-is (but didn't).

onovotny:
cross posting my comment from the other thread:

Ugh, this whole thing could have been a giant brain-fart looking at nuget.org. It appears to have been fine originally :(

Very sorry for the false alarm and no hard feelings if you want to revert my commits. If you do revert, the one thing you'll want to keep is the `build.ps1` update to the NuGet download location as that'll pull in v3. the existing url is frozen at v2.

michael-wolfenden:
No problem @onovotny, happens to the best of us. Thanks for clarifying though and thanks in general. Your posts have be invaluable in helping me understand the new nuget monikers and net core.

joelhulen:
Agreed, this has been very helpful to me as well, in better understanding the NuGet dependency intricacies. The question now is whether to roll back these changes and keep the new NuGet url?

onovotny:
In the end, I'd have to ask what benefit do you see from manually calling the cli from the cake script as opposed to having the targets do it on build?

There's no right or wrong answer, I just figure that in the unlikely event the params change, the targets will always be right. I pretty much assumed the cli would be more for building on mono or some other usage that people come up with.

Your call either way.

BTW, there seems to be an extra packages.config in the .pcl folder that has a GitVersion ref in it. That file should prob be removed and if you need the ref there, moved into the project.json (should you keep it).

PR #61, opened 2015-12-04
Use Nuget v3 and remove refgen from cake (handled by build)

onovotny:
This should address #59 outstanding issues


joelhulen:
You, sir, are a gentleman and a scholar!

PR #62, opened 2015-12-05
Allow reset of circuit-breaker policy

DennisNerush:
Sometimes you want the ability to reset the policy and start over (tests or application logic).
First I've add the circuitBreakerState to the Policy instance.
Then I've exposed a "Reset" method on the Policy instance which simply calls The state's reset method and by that reset the policy. I've added a test to support that it works.


reisenberger:
@DennisNerush [aside: I'm just another user very interested in Polly, not core team]

What happens if the Policy is a retry policy type, and `.Reset()` is called? IE

```csharp
Policy retryPolicy = Policy.Handle<Exception>().Retry();
retryPolicy.Reset();
```

@joelhulen Perhaps `Policy` abstract base class with derived types `CircuitBreakerPolicy` and `RetryPolicy`, if desired, to accommodate this?

reisenberger:
@joelhulen There could be other benefits to explicitly separating out policies in to separate `CircuitBreakerPolicy` and `RetryPolicy` types.  Example: In our microservices implementation, for certain calls we allow callers to pass in the retry policy or circuit-breaker policy (but specifically one or the other) they want to use.  Because the Polly library at present does not distinguish the types, we cannot use `Policy` as a parameter type (people could pass in a circuit-breaker where a retry is expected...).  
It's ended up in us having to wrap `Policy` in marker classes equivalent to what I've suggested (`CircuitBreakerPolicy` and `RetryPolicy`), these classes enforce the kind and re-duplicate the Polly interface for it but only pass all calls through to a relevant inner `Polly.Policy` - long-winded.  

Having separate `CircuitBreakerPolicy` and `RetryPolicy` types  could allow library users to specify to expect / pass around instance of eg `RetryPolicy` type only, as well as build functionality specific only to either the `CircuitBreakerPolicy` and `RetryPolicy` type.

If you want, I can open this as a separate issue # for tracking.

DennisNerush:
I can separate them and ask for another pull request. Would you want it?

reisenberger:
@DennisNerush I've noted a few more potential complications with the separating out in #63 (although I think it's a good idea).  Stepping back now to let @joelhulen and team consider way forward, and/or @michael-wolfenden clarify.

DennisNerush:
Tell me how I can help :)

reisenberger:
Thanks @DennisNerush.  (Apologies for delay in replying - AppvNext have asked me to join the project team on this and we've been reviewing the various issues and requests.)  

Suggest hold on any new PR for now, as we intend first (per #63) to have a broader think about where/how abstract base classes or interfaces might be introduced.  That is planned to follow fixes/enhancements to async retry (53, 46).  

However, your idea to be able to reset circuit breaker state fits right in with other requests for manual control over the circuit breaker (eg ability to break it manually, for manual override; then likewise reset it). We plan to look at circuit breaker when the stuff mentioned above is in (may take until the new year), but it seems like a sensible feature, so: watch this space!

reisenberger:
@DennisNerush We are making a PR that will implement your request here - thanks for the initial idea.

DennisNerush:
Greast!

joelhulen:
This has now been rolled into version 4.0.

Issue #78, opened 2016-02-03
ContextualPolicy is missing ExecuteAndCapture() support


reisenberger:
`ContextualPolicy` does not have `.ExecuteAndCapture(...)` support.  One can (at present) only call `.Execute()` on the `ContextualPolicy`

reisenberger:
May relate to #72; also noticed while investigating #63

Suggest we fix this after #63 .

PR #82, opened 2016-02-17
ContextualPolicy: add async support; bind more tightly with (extend) Policy

### LINKED ISSUE #64: Cannot combine ContextualPolicy and Policy; [DONE: document nesting Polly statements]
Hi,

I've just battled through a rather complex policy statement. I think it would help as an example on the readme and/or samples project.
Moreover I think I've found a gap in the API or rather something rather non-obvious that would be worth helping either via a new method/extension.

In particular we have a need to wrap Policy statements - a retryforever (that logs error) wrapping a retry and backoff (that logs warning). An extra complexity is that somethimes we want that behaviour and sometimes we want another policy. I know.

So our policy:

```
private Polly.Policy SetRetryPolicy(int retries, int retryWaitInMilliseconds, bool isFromApi)
{
	if (!isFromApi)
	{ 	
		return
			Polly
				.Policy
				.Handle<Exception>()
				.RetryForever(exception =>
				{
					_log.ErrorFormat("An erros with Exception detail " + exception);
				})
				.Execute(() =>
					Polly
						.Policy
						.Handle<Exception>()
						.WaitAndRetry(
							retries,
							retryAttempt => TimeSpan.FromMilliseconds(retryWaitInMilliseconds*Math.Pow(2, retryAttempt)),
							(exception, timeSpan) =>
							{
								_log.WarnFormat("An erros with Exception detail " + exception);
							}));
	}
	return
		Polly
			.Policy
			.Handle<AggregateException>()
			.Retry(retryWaitInMilliseconds, (exception, i) =>
			{
				_connectionSingleton.Stop();
				_log.ErrorFormat("An erros with Exception detail " + exception);
			});
}
```

The issue I'd like to point attention to is I *was* passing to the WaitAndRetry method above the following:
```
(exception, timeSpan, *context*) => ...
```
.... this has the effect of creating a ContextualPolicy, whereas the second Policy was a Polly.Policy. 

- There is no way to combine a ContextualPolicy and Policy
- It's *very* easy to miss this in amongst the various overloads.

I love how Polly works but the lack of examples covering wrapping policies really stymied me. Any help / guidance for others could help.

### LINKED ISSUE #81: ContextualPolicy has no async support
ContextualPolicy currently supports `.Execute()`, but not `.ExecuteAsync()`
(compare issue #78)

PR for this coming shortly.

reisenberger:
Adds async support to `ContextualPolicy` - closes #81 .  

(Lays ground also for later addressing 78 : Add `ExecuteAndCapture/Async()` support for `ContextualPolicy`)

Closes #64 (cannot combine `ContextualPolicy` and `Policy`).

Tidies by moving certain async tests which were in sync test classes, into async test classes.

### DIFF
--- a/src/Polly.Shared/Context.cs
+++ b/src/Polly.Shared/Context.cs
@@ -13,7 +13,7 @@ namespace Polly
     /// </summary>
     public class Context : ReadOnlyDictionary<string, object>
     {
-        internal static Context Empty = new Context(new Dictionary<string, object>());
+        internal static readonly Context Empty = new Context(new Dictionary<string, object>());
 
         internal Context(IDictionary<string, object> values) : base(values)
         {
--- a/src/Polly.Shared/ContextualPolicy.cs
+++ b/src/Polly.Shared/ContextualPolicy.cs
@@ -8,15 +8,11 @@ namespace Polly
     /// Transient exception handling policies that can be applied to delegates.
     /// These policies can be called with arbitrary context data.
     /// </summary>
-    public class ContextualPolicy
+    public class ContextualPolicy : Policy
     {
-         private readonly Action<Action, Context> _exceptionPolicy;
-
-         internal ContextualPolicy(Action<Action, Context> exceptionPolicy)
+        internal ContextualPolicy(Action<Action, Context> exceptionPolicy, IEnumerable<ExceptionPredicate> exceptionPredicates) 
+            : base(exceptionPolicy, exceptionPredicates)
         {
-            if (exceptionPolicy == null) throw new ArgumentNullException("exceptionPolicy");
-            
-            _exceptionPolicy = exceptionPolicy;
         }
 
          /// <summary>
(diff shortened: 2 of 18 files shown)


reisenberger:
**COMMUNITY ALERT  Upcoming edge-case breaking change** _only_ if currently combining `ContextualPolicy` and `Policy` in certain very specific ways.  

Read on if you think this could affect you - though steps have been taken to minimise impact / ensure only extreme edge cases may be affected.

**BACKGROUND**

The existing class structure of Polly had `ContextualPolicy` and `Policy` with no relation.  Arising from this:

1. People wanting to combine both `ContextualPolicy` and `Policy` faced barriers doing so (#64)
- `ContextualPolicy` was generally not tied in to a number of features provided for `Policy`; eg no async support (#81), no `.ExecuteAndCapture()` support (#78)
- The lack of relation between `ContextualPolicy` and `Policy` meant it would be hard to split Circuit-Breaker from Retry (in order to add deeper functionality specific to circuit-breaker - requested in several PRs/issues) ... while still maintaining support for both `ContextualPolicy` and `Policy` in both circuit-breaker and retry (which we did not want to sacrifice).

**Conclusion**

We propose `ContextualPolicy : Policy`.

This resolves all aforementioned issues and provides a platform to grow `ContextualPolicy` with full support in both retry and circuit-breaker.  However, it was identified this _could be a breaking change in certain very limited scenarios_.

**The limited scenario affected by a breaking change**

The only usages that could experience a breaking change would be code which reasons based on the current split `Policy` / `ContextualPolicy` structure in one particular way.

- _only_ if you have noticed the problem @penderi raised in #64 
- _only_ if you who have chosen to solve it by trying to capture either policy type into an item of type `object` (because there was no other choice)
- and _only further_ if you have code reasoning which this `object` represents, with code such as:

```
if (myCouldBeEitherType is Policy) { // then it can’t be a ContextualPolicy } 
```
Code which reasons in the _above manner only_ may give unexpected results after this PR is released.

Code facing the same dilemma which reasons in the opposite direction, taking the more specific case, would be fine:

```
if (myCouldBeEitherType is ContextualPolicy) { // do something specific to ContextualPolicys }
``` 

***FOR CLARITY - ALL MAINSTREAM USAGES OF POLLY UNAFFECTED***

For clarity, all mainstream uses of both `Policy` and `ContextualPolicy` will be unaffected and experience no breaking change:

```
var myPolicy = Policy.Handle<Whatever>.WhateverPolicy(my, arguments) // unaffected …
Policy myPolicy = Policy.Handle<Whatever>.WhateverPolicy(my, arguments) // unaffected  …
var myPolicy = Policy.Handle<Whatever>.WhateverPolicy(my, arguments, context) // unaffected …
ContextualPolicy p = Policy.Handle<Whatever>.Whatever(my, arguments, context) // unaffected …
```

PR #108, opened 2016-05-21
Support async onRetry delegates on async retry policies

### LINKED ISSUE #107: Supplying an async `onRetry` delegate compiles but compiler assignment to async void lambda causes delegate completion not to be awaited
I have an issue where I need to call an async method from the Retry block that changes authorisation criteria within an object before executing the query again.

  ```
          query = bookingQuery;

            try
            {
                bookings = await Policy
                    .Handle<ApiException>()
                    .RetryAsync(2, async (ex, retry) =>
                        {
                            await Task.Run(async () => await RefreshApiAuthentication(retry));
                            System.Diagnostics.Debugger.Break();
                        })
                    .ExecuteAsync(async () => await GetBookingsApiCall());
                
            }
            catch (Exception ex)
            {
                System.Diagnostics.Debugger.Break();
            }
```

```
        private async Task RefreshApiAuthentication(int retryCount)
        {
            switch (retryCount)
            {
                case 1:
                    query.Authorisation = await GetRefreshAuthorisation(query.Authorisation, query.ClubId);
                    System.Diagnostics.Debugger.Break();
                    break;
                case 2: 
                    query.Authorisation = await GetRefreshAuthorisation(query.Authorisation, query.ClubId);
                    System.Diagnostics.Debugger.Break();
                    break;
                default:
                    break;
            }
        }
```

This problem I have is that in the RefreshApiAuthentication method once the async method is called the ExecuteAsync method is called by Polly. Is there a method of calling async methods from the Policy RetryAsync method that would await before executing the retry?

Great library by the way.

reisenberger:
Fixes #107 . 

@joelhulen Lodging this PR for review at your leisure.

We can also await @PhilipAnthonyMurray 's testing against the real-world scenario he originally raised.  However, the unit tests also cleanly demonstrated the issue and prove the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.2.1
+
+- Allowed async onRetry delegates to async retry policies.  Thanks to [@reisenberger](https://github.com/reisenberger)
+
+
 ## 4.2.0
 
 - Add AdvancedCircuitBreaker.  Thanks to [@reisenberger](https://github.com/reisenberger) and [@kristianhald](https://github.com/kristianhald)
--- a/GitVersionConfig.yaml
+++ b/GitVersionConfig.yaml
@@ -1,1 +1,1 @@
-next-version: 4.2.0
+next-version: 4.2.1
--- a/README.md
+++ b/README.md
@@ -208,6 +208,8 @@ Policy
     });
 ```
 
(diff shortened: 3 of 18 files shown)


reisenberger:
Also fixes #109

joelhulen:
@reisenberger Thanks so much for this pull request! I plan on merging it tomorrow.

Issue #90, opened 2016-03-07
Future Polly Roadmap - **community feedback sought**


reisenberger:
App-vNext has [published a Roadmap](https://github.com/App-vNext/Polly/wiki/Polly-Roadmap) for Polly.  

**_We are actively seeking community feedback on the roadmap._**

Which aspects of the roadmap would be a priority for you, and which not?

Do you have views on how the features in the roadmap should be implemented?

Are there features you'd like to see, that are not on the roadmap?

What is your view on the longer-term possibility, to grow Polly towards a more general resiliency framework for .NET? (as Hystrix is for Java) – for detail see [Roadmap](https://github.com/App-vNext/Polly/wiki/Polly-Roadmap).  .  

_Please comment on the Roadmap under this placeholder issue._

mfjerome:
About your last point: 

> Several of our teams working on .NET projects have recommended Polly as being useful in building microservice-based systems. It encourages the fluent expression of transient exception-handling policies and the Circuit Breaker pattern, including policies such as Retry, Retry Forever and Wait and Retry. Similar libraries already exist in other languages (Hystrix for Java for example), and Polly is a welcome addition from the .NET community. 

Taken from ThoughWorks Technology Radar. There is clearly a lack of good offers in the .NET ecosystem for general resiliency framework and the spotlight is already on Polly as a Hystrix(Java) alternative so why not jump on the occasion =)

bunceg:
On the open issue #6, I appreciate it wasn't the path the team wanted to go at the time, but I really can't see any other clean way the issue below...

> Two instances in their own processes, each has an instance of circuit breaker. Instance A trips at 00:00 and lasts for 5 minutes.. Due to random load balancing, *most* requests just happen to go to instance A. However, by chance, at 04:50 instance B gets some requests and also trips. it's 5 minute trip timer now starts. However, the tripped service is back at 05:00. The instance A breaker is cleared at 05:01 but the instance B breaker is not able to be cleared until 09:50 - 4+ minutes after when it should be able to serve requests.

Without some kind of shared state, and injecting this state manager into the circuit breaker, I don't see how any of the admittedly very useful updates so far, helps with this scenario.

reisenberger:
Hi @bunceg, many thanks for the input!  Noted that you are still looking for the best solution to shared circuit-breaker state across processes with Polly, and/or a feature to inject state into a circuit.  _Further views on this welcome from across the community..._

[...]

PR #127, opened 2016-06-20
Dont lock on hot path reads if we can help it

christopherbahr:
I may be missing something here but I think we can get away without the locking in `OnActionPreExecute` since enum reads are atomic. I don't think this actually protects us from anything.

I also added a check in the reader so if we aren't going to do any writing we can do the read without locking. 

The only edge case this introduces is if we make our read, the circuit is not open, someone sets it to open, and we return open without ever checking whether we should go to half open (which we almost certainly should not do that soon after opening anyway but it's net necessarily prevented). 

Before these changes were made the reader would have gotten a lock first though and the write wouldn't have executed until we were done anyway so there's not actually behavior difference there.

### DIFF
--- a/src/Polly.Shared/CircuitBreaker/CircuitStateController.cs
+++ b/src/Polly.Shared/CircuitBreaker/CircuitStateController.cs
@@ -31,6 +31,10 @@ namespace Polly.CircuitBreaker
         {
             get
             {
+                if (_circuitState != CircuitState.Open)
+                {
+                    return _circuitState;
+                }
                 using (TimedLock.Lock(_lock))
                 {
                     if (_circuitState == CircuitState.Open && !IsInAutomatedBreak_NeedsLock)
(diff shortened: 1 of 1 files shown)


reisenberger:
@christopherbahr Thanks!  On a quick read this looks good - I'd like to have a little more time to think over the edge case/s, so give me couple days on that ... (agree there is the edge case you mentioned; and it probably pans out as you say)... Optimisations always welcome, though, so thanks!

I also have some CircuitBreaker timing tets (testing 100000 calls through the breaker) which we put together when performance-testing implementation options for the new `AdvancedCircuitBreaker`: it produces stats for well-behaving underlying systems, frequently faulting systems, etc ... so I'll run the change through the performance-tester too and see what we get!

Your optimisation also depends on the switch expression only being evaluated once ... I'd eat my hat if that wasn't the way it was implemented :smile: (and think remember reading that somewhere), but shout if you know different! 

Thanks!

christopherbahr:
The spec says the first step in evaluating a `switch` statement is 
"The switch expression is evaluated and converted to the governing type."
http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-334.pdf page 232

That reads to me like it will store the outcome of the switch expression rather than recalculating it. (I've seen someone verify this by looking at the IL but the compiler is allowed to do whatever it wants as long as it follows the spec so I'm not sure we want to rely on that). 

That said if you're worried about it I can just copy the value into a local variable and then switch based on that. 

Let me know how the performance testing goes. It was actually during performance/load testing something that depends on Polly that I came across this.

reisenberger:
@christopherbahr Many thanks for this PR!  Squeezed in some time for performance-testing after the recent major feature work, and agree that this is a valuable optimisation to the hot path.  [This commit](https://github.com/App-vNext/Polly/pull/130/commits/3b32ec5e6fbeb77b76c6e36b584c5127a9f48509) contains an equivalent change to the new `CircuitStateController`, and we expect this to be released as part of #130 in the next few days.  I am not proposing we do any special workaround for the `switch` statement; I agree with your assessment of the edge case in your original post.
 
The next phase of Polly (have you seen the [Roadmap](https://github.com/App-vNext/Polly/wiki/Polly-Roadmap)?) is very much intended to support high volume / low latency scenarios, with new policies such as cache, fallback, and parallelism-throttle for bulkhead isolation or (if user requires) load-shedding.  If you are working in a high-throughput/low latency environment, now is a great time to provide any feedback you have or feature requests under the related [feedback issue](https://github.com/App-vNext/Polly/issues/90).  
 
Also continuing to develop the performance test bed for Polly: if you are able to share the kind of throughputs you are dealing with, whether sync or async, etc (post here or email me at software[at]reisenberger.net if you prefer), it might be interesting to put your kind of numbers through the test bed?
 
Thanks

joelhulen:
These changes were incorporated into PR #130 

Thanks for your great contribution!

Issue #129, opened 2016-06-27
Proposal: Extend Polly with Policy<TResult> strongly-typed generic policies


reisenberger:
The current `Policy` class in Polly is non-generic. For executing functions which return a result, there are generic _methods_ `.Execute<TResult>(Func<TResult>)` (and async equivalents) on the non-generic class.

### Drivers for change

Two future developments on the Polly roadmap suggest support for generic, strongly-typed policies `Policy<TResult>` could be beneficial:

**Issue #14 Allow policies to handle return values**:  If a Policy is configured to handle return values [ eg `.HandleResult<HttpResponse>(r => statusesWorthRetrying.Contains(r.StatusCode))` ], it only makes sense for these policies to execute delegates returning the relevant return type (`HttpResponse` in this case).  Executing a func or action returning some other return type couldn't be interpreted in the terms of the policy definition.  Strongly-typed generic policies would enforce this well.

The **hystrix-like policies `Fallback` and `Cache`** likewise only make sense for delegates returning a given `TResult` type.  And combining such policies into a **`Pipeline`** likewise only makes conceptual sense if the pipelined policies return the same `TResult` type.  Again, this is much more logically enforceable with strongly-typed generic policies.

### Implementation considerations

**Backwards compatibility** is a given: The new `Policy<TResult>` forms should be introduced without affecting the existing forms.

**Syntax familiarity**: Any new syntax should be familiar and compose fluidly with the existing syntax.  eg 

```csharp
int[] httpStatusCodesWorthRetrying = {408, 500, 502, 503, 504}; 
HttpResponse result = Policy
  .Handle<HttpException>()
  .OrResult<HttpResponse>(r => httpStatusCodesWorthRetrying.Contains(r.StatusCode))
  .Retry(...)
  .Execute<HttpResponse>( /* some Func<...,HttpResponse> */ )
```

**Common underlying implementation**: A common underlying implementation for the generic and non-generic variants of each Policy would aid development and maintenance.  Enhancements and fixes would automatically apply to both.

reisenberger:
Delivered in #130.  On Nuget as v.4.30

PR #130, opened 2016-06-27
Allow policies to handle return results / unblock forward roadmap

reisenberger:
This PR addresses long-standing request #14 (allow policies to handle return results) (also requested under Scott Hanselman's blog).  

It also implements #129, which in turn opens the way for a number of the Hystrix-like features on the Polly [Roadmap](https://github.com/App-vNext/Polly/wiki/Polly-Roadmap), such a Fallback, Cache and fluidly combining policies into pipelines or wraps.

The implementation is code-complete and I am thus opening it @joelhulen for cross-review within our team as usual.

Before merging we (I) also need to re-base, **update/provide the documentation** and update the relevant Polly.nuspec files etc.


pvmraghunandan:
Very useful in scenarios of REST (or) having WCF service that returns custom exception if its external components throttles. Any idea on when this will be available as public?

reisenberger:
@pvmraghunandan This is now both code-complete and documentation-complete.  It is just undergoing our governance process of within-team code-review (by another AppvNext team member) before being merged and released.  I would expect it to be released within the next few days.

joelhulen:
@pvmraghunandan The latest NuGet packages have been published!

pvmraghunandan:
Thanks for updating. Could see new nuget version.

pvmraghunandan:
As per current implementation, we are expecting list of result predicates and the return type is tightly coupled with policy object. If we want to have one policy object for an WCF service, the implementation is forcing us to use object type and do typecast and all sort of stuff and also couldnt write generic code.. The code is:

if (shouldHandleResultPredicates.Any(predicate => predicate(delegateOutcome.Result)))

Do we ever have multiple predicates practically? Shouldnt we update logic to accept predicate during Execute() .Any feedback so that can start updating code.

reisenberger:
@pvmraghunandan I will try to get back to you on this within the coming day.

reisenberger:
hi @pvmraghunandan. It’s a while since I looked at some WCF code, so could I make sure I’ve understood the problem properly? (please correct me if I’m wrong). 
 
Is the problem that you might have some WCF service offering a number of functions returning different types, and you want to apply some common fault-handling policies (eg retries) across the calls returning different types?  eg Something like the below _schematically_ (I’m omitting WCF attributes for brevity)
 
```csharp
public class MyFooService : IMyFooService // a service exposed across WCF to some client somewhere
{
    public int Foo1(Guid id) { ... }
    public string Foo2(Guid id) { ... }
}
```
 
For example, perhaps at the caller you want to define some common RetryPolicy for `MyFooService` across these two calls with return different types? (int, string).  And then maybe some extra value-based fault-handling specific to each call? (like, null or 0 should also be considered failures and retried). 
 
If this is not the class of problem, could you post a simple code example?
 
Also: are you seeking to apply the Polly policies on the service side (when handling the requests, perhaps because that has other vulnerable downstream dependencies), or on the client side? (when calling the service), or both?

Thanks

reisenberger:
@pvmraghunandan Thank you for highlighting the question of applying a single policy to methods of different return types (such as possibly in a WCF service), with the strongly-typed `Policy<TResult>` form.

The move #129 to adopt strongly-typed policies `Policy<TResult>` was influenced by factors on the [wider resilience roadmap](https://github.com/App-vNext/Polly/wiki/Polly-Roadmap#a-wider-resilience-framework): `CachePolicy` and `FallbackPolicy` in particular (and combining them in a `PolicyWrap`) only really make sense in strongly-typed forms.

If we stuck with only a non-generic `Policy` with generic `Execute<TResult>()` methods, and we did this:

>Shouldnt we update logic to accept predicate during Execute()?

we would also have to move cache and fallback functionality into the `Execute<TResult>()` methods.  The cache and fallback functionality would have to be reduced (eg no `onFallback` delegate), and we would quickly have quite complex `.Execute()` overloads like:

`.Execute<TResult>(Func<TResult> executionFunc, Func<TResult, bool> shouldHandleResult, IResultCacheProvider<TResult> cache, Func<TResult> fallbackFunc)`

If we then wanted to offer overloads in variants including/not including `cache`, `fallbackFunc` and `shouldHandleResult` (not everybody wants all all the time), quickly the number of overloads multiplies - on top of the existing already large number of overloads (see async forms).  This development direction for Polly felt unsustainable.

Strategies for dealing with the large number of overloads almost all involve moving instead to a Hystrix-like `Command` class and setting properties to configure it, but that would inevitably be strongly-typed `Command<TResult>` too.

reisenberger:
@pvmraghunandan To use policies of a common characteristic across calls returning different types without resorting to `object` and casting, one option is to add your own extension methods to Polly's `PolicyBuilder<TResult>` class.  (Polly's fluent policy-construction syntax depends on a public `PolicyBuilder` class, but this isn't usually seen because of the way the syntax encourages constructing a whole policy in one statement.)

Extension method examples:

```csharp
public static PolicyBuilder<TResult> WithStandardExceptionHandling<TResult>(this PolicyBuilder<TResult> builder)
{
    return builder
        .Or<ExceptionA>()
        .Or<ExceptionB>(); // etc - more detailed predicates as needed
}

public static RetryPolicy<TResult> WithStandardRetryPolicy<TResult>(this PolicyBuilder<TResult> builder)
{
     return builder
        .Retry(1); // or whatever
}
```

Then you can concisely construct policies with similar characteristics for different `TResult`-types, and with different specific handling for the `TResult`.

```csharp

var intVariantOfPolicy = Policy
    .HandleResult<int>(i => i < 0)
    .WithStandardExceptionHandling()
    .WithStandardRetryPolicy();

var stringVariantOfPolicy = Policy
    .HandleResult<String>(s => s == null || s.Length < 2)
    .WithStandardExceptionHandling()
    .WithStandardRetryPolicy();
```

Does this help with your WCF case?

pvmraghunandan:
@reisenberger  Sorry for late reply. Little bit occupied with other stuff.  We are looking for below.

"WCF service offering a number of functions returning different types, and we want to use same policy for all (Policy for Service) but the result handling is different".

 Today we can't achieve the common policy because if we want to apply result handling and when we say .OrResult(), the return type is PolicyBuilder<T>.

Also We have onRetry and other delegates also as independant because we want to tie up with our logging engine where activityId and other related information is different for each call. With passing onRetry during object creation, we cant achieve this.

Hystrix was one consideration but creating command classes for each type that we are going to handle might not be better way is what we feel.

reisenberger:
Hi @pvmraghunandan 

>Today we can't achieve the common policy because if we want to apply result handling 
>and when we say .OrResult(), the return type is PolicyBuilder

`PolicyBuilder<TResult>` presumably? (A non-generic `PolicyBuilder` occurs for the non-result-handling policies.)  See if my earlier suggestions about extension methods chaining on `PolicyBuilder<TResult>`  help.

>We have onRetry and other delegates also as independant because we want to tie up with our 
>logging engine where activityId and other related information is different for each call. 
>With passing onRetry during object creation, we cant achieve this

This can be achieved in existing Polly.  Configure the policy with an `onRetry` taking a `Context`.  `.Execute()` on the policy with an `.Execute()` (or similar) overload taking a `Context`.  The context passed in to the execute call gets passed to the onRetry.

pvmraghunandan:
Hi @reisenberger 

Thanks for prompt reply. For onRetry() will give a try once. 

But didn't get on one policy that returns different results and handling predicates differently. Basically Result Type is unknown at compile time so today we are casting back and using is/as operators to achieve this.

reisenberger:
@pvmraghunandan No problem.  Re:

>Result Type is unknown at compile time so today we are casting back 
>and using is/as operators to achieve this.

If `TResult` is unknown at compile time but is known at run-time, you could consider using [MakeGenericMethod](https://msdn.microsoft.com/en-us/library/system.reflection.methodinfo.makegenericmethod(v=vs.110).aspx) and [MakeGenericType](https://msdn.microsoft.com/en-us/library/system.type.makegenerictype(v=vs.110).aspx) ? (Since in Polly the `TResult`-policies are constructed using static method calls rather than `new`-ing, that would be `MakeGenericMethod` more than `MakeGenericType`).  

These can effectively invoke closed/bound generic methods or instantiate closed/bound generic types from their unbound equivalents at runtime.  It can be quite powerful, including working with nested generics.  And if, once ResultType is known at runtime, you expect to see the same ResultType/set of ResultTypes over and over, you can cache the results of the reflection rather than repeat it each time.

(All this may or may not make things simpler, depending on what you are looking to achieve!)

pvmraghunandan:
@reisenberger Completely agree on caching reflection and all. Infact we are doing in little different way so that we avoid much reflection. But why cant we expect delegate as an Execute() parameter so that its up to caller to send whatever delegate he need. In our case especially we deal with multiple external systems and each has multiple methods. Some will send specific error codes that indicates their transient failure and some will not. So its upto implementation (In our case adapter) to send result handling delegate as i am already invoking policyBuilder with type as specific type.

reisenberger:
Hi @pvmraghunandan . Thank you for the explanation.  Re:

>But why cant we expect delegate as an Execute() parameter 
>so that its up to caller to send whatever delegate he need. 

The driver here @pvmraghunandan is coherence of the library and its API.  Since which exceptions to handle was already specified in Polly during the Policy configuration stage - not at execution - when we came to add handling results, it was considered more coherent for the library and API if the handling of results was specified in the same place.  I am not keen (unless there is significant community feedback) to split the specification of what to handle - exceptions and results - into two _different_ places - configuration and execution - as this seems overall like a less coherent product and API.   In essence we are guided/constrained by the product's history.

There could of course be more total rewrites - moving specifying both exception-handling and result-handling to execution-time - but this would be a significant change in API/approach and a significant rewrite - anything is possible, but it would have to be strongly motivated!   

The existing Polly team are of course limited in their time, but always open to views: Do other people have views on the Policy construction API?

reisenberger:
@pvmraghunandan Given this:

>In our case especially we deal with multiple external systems and each has multiple methods. 
>Some will send specific error codes that indicates their transient failure and some will not. 
>So its upto implementation (In our case adapter) to send result handling delegate as i am 
>already invoking policyBuilder with type as specific type.

Given the current construction of the Polly API, I would then recommend the solution from my [previous comment](https://github.com/App-vNext/Polly/pull/130#issuecomment-232279874): Use extension methods so that you can state common policy characteristics in a central place for the service - or pass them in as a configuration func - but configure the variant parts (result-handling) as individually as you want for each endpoint.  This is (for sure) not the same as having a single policy for the service on which you can vary result-predicates at execution time, but gets you closer.

PR #147, opened 2016-08-04
updated readme on thread safety

### LINKED ISSUE #146: document thread safety
I am about to be sharing a policy across several threads. The readme does not describe anything in relation to thread safety. Looking at the code, it seems to be thread safe. But I think it would be great if the readme clearly could state that the code is supposed to be thread safe - or whether that is a job for the user of the library.

kbilsted:
close #146

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Polly
 
-Polly is a .NET 3.5 / 4.0 / 4.5 / PCL (Profile 259) library that allows developers to express transient exception- and fault-handling policies such as Retry, Retry Forever, Wait and Retry or Circuit Breaker in a fluent manner.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -423,6 +423,14 @@ policyResult.Result - if executing a func, the result if the call succeeded or t
 */
 ```
 
+
+# Thread safety
+
+The existing policies (retry and circuit-breaker) are fully thread-safe, and the intention is that the forthcoming wider resilience policies will be thread-safe too. 
+
+While the internal operation of the policy is thread-safe, but this does not magically make delegates you execute through the policy thread-safe: if delegates you execute through the policy are not thread-safe, they remain not thread-safe.
```
reisenberger: Thanks for this @kbilsted . Suggest just remove 'but' from this sentence for grammatical correctness.

@joelhulen looks like a good clarification from @kbilsted to merge!

kbilsted:
done

joelhulen:
Thanks, @kbilsted!

PR #157, opened 2016-09-24
Performance refinement - reduce allocations

reisenberger:
Reduce allocations by sharing single static allocation for policies where some predicates may be irrelevant

### DIFF
--- a/src/Polly.Shared/AdvancedCircuitBreakerSyntax.cs
+++ b/src/Polly.Shared/AdvancedCircuitBreakerSyntax.cs
@@ -215,7 +215,7 @@ namespace Polly
                     context,
                     cancellationToken,
                     policyBuilder.ExceptionPredicates, 
-                    Enumerable.Empty<ResultPredicate<EmptyStruct>>(), 
+                    PredicateHelper<EmptyStruct>.EmptyResultPredicates, 
                     breakerController), 
                 policyBuilder.ExceptionPredicates, 
                 breakerController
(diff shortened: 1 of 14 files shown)


src/Polly.Shared/AdvancedCircuitBreakerSyntax.cs
```diff
@@ -215,7 +215,7 @@ public static CircuitBreakerPolicy AdvancedCircuitBreaker(this PolicyBuilder pol
                     context,
                     cancellationToken,
                     policyBuilder.ExceptionPredicates, 
-                    Enumerable.Empty<ResultPredicate<EmptyStruct>>(), 
+                    PredicateHelper<EmptyStruct>.EmptyResultPredicates,
```
WilkaH: `Enumerable.Empty` already [uses a single, shared, instance](http://referencesource.microsoft.com/#System.Core/System/Linq/Enumerable.cs,1162). i.e.

```
public static IEnumerable<TResult> Empty<TResult>() 
{
	return EmptyEnumerable<TResult>.Instance;
}
```


which uses 

```
    internal class EmptyEnumerable<TElement>
    {
        public static readonly TElement[] Instance = new TElement[0];
    }
```

reisenberger:
Good catch @WilkaH !  And exemplifies something I'm often saying to people: "check the reference source!" (should've practised what I preached on this one :wink: ).  (We'll probably reverse out that `PredicateHelper<>` later, as doesn't aid performance or readability.)

PR #160, opened 2016-10-06
Bulkhead policy (sync/async; TResult and non-generic)

reisenberger:
Add Bulkhead policy and multi-threaded, highly-parallelized, simultaneous-load specs


joelhulen:
Brilliant work, Dylan! Thank you for your contribution :)

PR #175, opened 2016-11-05
Tidy the v5.0-alpha build

reisenberger:
Various tidying to the v5.0-alpha build:

+ Correctly state dependencies for .NETStandard1.0 target
+ Remove NuSpec.ReferenceGenerator
+ Remove GitVersionTask from .NetStandard10 project (it is not compatible)
+ Tidy-out duplicate xunit references
+ Update project description


reisenberger:
@joelhulen Please do not merge this yet: further build tidying commits may follow.

(xunit reference changes reverted due to differing cake-script behaviour between local runs and appveyor; no change to actual test functionality)

src/Polly.nuspec
```diff
@@ -141,16 +145,13 @@
...
-        <dependency id="System.Runtime.Extensions" version="4.0.0" />
-        <dependency id="System.Threading" version="4.0.0" />
-        <dependency id="System.Threading.Tasks" version="4.0.0" />
+      <group targetFramework="netstandard1.0">
+        <dependency id="System.Diagnostics.Debug" version="4.0.11" />
+        <dependency id="System.Linq" version="4.1.0" />
+        <dependency id="System.Runtime.Extensions" version="4.1.0" />
+        <dependency id="System.Threading" version="4.0.11" />
       </group>
+      <group targetFramework="dotnet"/>
```
twsouthwick: This is a deprecated target framework. Is there a reason you want to include it?

src/Polly.nuspec
```diff
@@ -141,16 +145,13 @@
...
-        <dependency id="System.Reflection" version="4.0.0" />
-        <dependency id="System.Collections" version="4.0.0" />
-        <dependency id="System.Diagnostics.Debug" version="4.0.0" />
-        <dependency id="System.Linq" version="4.0.0" />
-        <dependency id="System.Runtime" version="4.0.0" />
-        <dependency id="System.Runtime.Extensions" version="4.0.0" />
-        <dependency id="System.Threading" version="4.0.0" />
-        <dependency id="System.Threading.Tasks" version="4.0.0" />
+      <group targetFramework="netstandard1.0">
+        <dependency id="System.Diagnostics.Debug" version="4.0.11" />
```
twsouthwick: Instead of listing the specific assemblies, the recommendation is to reference .NET Standard instead of the individual dependencies (see https://docs.microsoft.com/en-us/dotnet/articles/standard/library)

src/Polly.nuspec
reisenberger: Thanks @twsouthwick .  Deprecated tfm removed.

src/Polly.nuspec
reisenberger: Thanks @twsouthwick .  Changed nuspec to state only the  "NETStandard.Library": "1.6.0" metapackage as dependency.

twsouthwick: Yeah, that was the original recommendation, but it's since proven to be more problematic to reference specific libraries. Hence, the recommendation is to now reference the whole standard.

src/Polly.Pcl/project.json
```diff
@@ -1,12 +1,7 @@
...
-        }
-    },
-    "frameworks": {
-        ".NETPortable,Version=v4.5,Profile=Profile259": { }
-    },
-    "supports": {}
-}
+  "dependencies": {},
+  "frameworks": {
+    ".NETPortable,Version=v4.5,Profile=Profile259": {}
```
twsouthwick: This profile is supported by .NET Standard 1.0, so it's not necessary any more if you don't want to maintain it separately (see https://docs.microsoft.com/en-us/dotnet/articles/standard/library)

reisenberger: Thanks @twsouthwick .  Yes, we're removing the PCL259 target in favour of the .NET Standard 1.0 target, as part of shortly-forthcoming Polly v5.0 RTM.  We figured last summer this was the way to go, but were deep in the many new features for Polly v5.0, and unsure at that time which would be the lowest .NET Standard we'd be able to target.  

(Looking forward to tidying up the solution further when Visual Studio "15" / 2017 is out at RTM / with stable project migration. I see from the [Visual 2017 RC release notes](https://www.visualstudio.com/en-us/news/releasenotes/vs2017-relnotes) couple days ago that the migration tooling is still at preview.)

Thanks @twsouthwick for all your input and support with this.

twsouthwick: Sounds good. Let me know if you have any questions about .NET Standard/VS 2017. btw I'm part of a team in the .NET/VS group whose goals include helping community projects adopt (and adapt) to the changes in the .NET ecosystem and help where needed. I recently learned of this project and have started incorporating it into one of our internal projects :)

reisenberger: :+1: Thanks @twsouthwick .  Will definitely revert to you if we have any qs when the VS2017 tooling comes out.  Great to have support direct from the MS teams.

Issue #174, opened 2016-11-04
Update NuSpec to say .NET Standard


twsouthwick:
Looks like #172 updated the build to include .NET Standard, but the NuSpec still says dotnet. This should be updated to say .NET Standard instead.

reisenberger:
You are absolutely correct @twsouthwick, thanks for picking this up.  Will be fixed in #175

PR #195, opened 2017-01-03
Removed all non-netstandard monikers from nuspec

DixonDs:
It seems that according to https://docs.nuget.org/ndocs/schema/target-frameworks and https://docs.microsoft.com/en-us/dotnet/articles/standard/library all those frameworks in nuspec should be supported by netstandard1.0

### DIFF
--- a/src/Polly.nuspec
+++ b/src/Polly.nuspec
@@ -147,14 +147,6 @@
      - Allow arbitrary data to be passed to policy execution
    </releaseNotes>
     <dependencies>
-      <group targetFramework="net" />
-      <group targetFramework="win" />
-      <group targetFramework="wp" />
-      <group targetFramework="wpa" />
-      <group targetFramework="MonoAndroid" />
-      <group targetFramework="MonoTouch" />
-      <group targetFramework="Xamarin.iOS" />
-      <group targetFramework="portable-net45+win8+wpa81+wp8" />
       <group targetFramework="netstandard1.0">
         <dependency id="NETStandard.Library" version="1.6.0" />
       </group>


reisenberger:
:+1: Thanks @DixonDs

reisenberger:
@DixonDs When I reviewed this PR, I had a doubt over removing the empty dependency group `<group targetFramework="net" />` :  Would removing this from the nuspec mean that when the Polly nuget package was installed into a project targeting the .NET Framework, nuget would then attempt to install the .NETStandard1.6 libraries as a dependency?  (This should not be necessary for a .NET-Framework-targeting project, as the Framework includes all the relevant Base Class Libraries.)

Testing suggests removing `<group targetFramework="net" />` is causing an issue: installing the resulting nuget package in a .NET4.5.x-framework-targeting DLL, seems to be making nuget install additional (unnecessary for .NET4.5) direct references to the various "System.*" packages and dlls.  Same for .NET4.6.x.  So we plan to re-instate the `<group targetFramework="net" />`.  

Does this make sense, or do you think we're missing anything?

(For background: the Polly.nupkg package does include separate compiles for .NET4.x, separate from the `.NETStandard1.0` target, for those still targeting the .NET framework directly.)

(/cc @twsouthwick Any comment on this or on the changes in the original PR welcome.)

DixonDs:
I think that is intention of .NetStandard - by using metapackage "NETStandard.Library", you can be sure that the same APIs are used across all frameworks.

If you still plan to reinstate "net" moniker, I would suggest to use specific monikers like "net45" etc, so that you target specific framework versions. Currently it shows ".NETFramework 0.0" on nuget page: https://www.nuget.org/packages/Polly/5.0.2-v5-0-alpha0001

twsouthwick:
@reisenberger  Those `System.*` packages you're seeing are what are called *type-forwarders*. This means they allow you to have binary compatibility; ie you expect `System.Collections.Generic.List<T>` to be in `System.Collections`, but on .NET 4.5 it's actually in mscorlib, so the `System.Collections.dll` just forwards the type to the correct assembly. However, on .NET Core, it is in `System.Collections` so it actually has the implementation there. Other implementations may have it in a different assembly, but the `System.*` libraries help ensure binary compatibility across implementations.

As @DixonDs pointed out, if you want a build for a specific .NET version, such as .NET 4.5, adding `net` won't work correctly, but using `net45` will. This way NuGet will know which target you are specifying it for. I did the same thing for a recent project, where I kept .NET 4.5 support and added .NET Standard 1.3 using the net45 and netstandard1.3 monikers (see https://github.com/OfficeDev/Open-XML-SDK).

Which targets are you planning to support with the library? If there's a list of those, we can help determine what the minimum set of targets should be to include in the nuget package.

reisenberger:
Thanks @twsouthwick for deep explanation.  Makes sense about _type-forwarders_, in terms of a library supporting .NET45 via a .NETStandard1.0 binary.  (Think some of the differing suggestions in this thread have stemmed from confusion over whether we are pursuing one single binary Polly.dll for both .NET45 and .NETStandard1.0, or separate binaries - currently separate.)

Re whether (**/cc @joelhulen**) to provide .NET4.5 support through a single (same) binary as .NETStandard1.0: _Advantage_: single binary.  _Disadvantage_: Install into a .NET4.5-targeting solution is much noisier than sticking with supporting .NET4.5 through a dedicated Polly.dll (as we currently do).  

In trials just now against a couple of in-house .Net4.5 / .Net4.6.1 projects which use Polly, installing a Polly.dll supporting .NET4.5 via a .NETStandard1.0 binary added (per project) approx 45 extra package references in the `packages.config` and 5-20 extra `System.*.dll` references in each `.csproj`.  I guess @twsouthwick this is the expected behaviour with the type-forwarders(?), but it makes for a much noisier/more interventionist install, for those users still targeting the .NET Framework?  

Any other thinking we're missing on this?

_(Thanks both @DixonDs and @twsouthwick for your contributions!)_

reisenberger:
@twsouthwick Our supported targets for Polly v5.0 RTM are:

(summary /cc @joelhulen for whether you have view on rationalising any of the nuget packaging) 

**.NET40 with Async:** 

+ Can't be supported via a .NET Standard binary
+ Will be via dependencies `<group targetFramework="net40"><dependency id="Microsoft.Bcl.Async" version="1.0.168" /><dependency id="Nito.Async" version="4.0.1-pre" /></group>`
+ Delivered (for historical reasons) as a separate nuget package https://www.nuget.org/packages/Polly.Net40Async/ . We could merge it into the main Polly package.  The separate package however quite usefully tells us (via its separate nuget stats) the extent of the .NET4.0 userbase (about 3%), which might guide future support decisions.

**Net Standard 1.0:**

+ For all portable support, .NET Core, future platforms
+ Dependencies (per #174) as `<group targetFramework="netstandard1.0"><dependency id="NETStandard.Library" version="1.6.0" /></group>`

**.NET4.5+ :**    

+ Dependencies currently declared as `<group targetFramework="net"/>` but definitely better as `<group targetFramework="net45"/>` (thanks @DixonDs for this)   
+ Currently delivered as a separate binary Polly.dll in main Polly nuget package, with Net Standard (eg https://www.nuget.org/packages/Polly/5.0.2-v5-0-alpha0001)

So whether to merge .NET4.5 into .NetStandard seems the main question ...

twsouthwick:
This plan sounds fine - the noisiness is unfortunately part of .NET Standard 1.0 right now. .NET Standard 2.0 reduces all the small, fine-grained packages into a single assembly (not sure if it's settled, but something like `netstandard.dll`) that then forwards where it needs to. For now, however, having three targets (`net40`, `net45`, and `netstandard1.0`) sounds like a good plan, especially since you already have the builds. 

I would suggest, for the ecosystem, to just have one package. The main reason is, if someone relies on your package for a library that supports multiple targets (including net40), they don't need to determine which package to use. Your use case is interesting, so I've opened an issue with the NuGet team to see if there is a way to get at that when it's in the same package (see https://github.com/NuGet/NuGetGallery/issues/3450)

joelhulen:
I would like to merge everything into a single NuGet package, but perhaps we could hold off to reduce confusion for the ~3% of the .NET40Async users and to continue to track its usage. It would be great to see what, if anything, the NuGet team comes up with regarding tracking capabilities of multiple targets within a package.

reisenberger:
Thanks @twsouthwick @DixonDs for all your useful input.

twsouthwick:
It doesn't look like this information is currently available as it is gathered from the client side. However, to get the best of both worlds, you can keep the two separate, but add a `net40` entry in the main package that just depends on the `net40` specific package.

PR #198, opened 2017-01-03
Merge v5.0-alpha to master, for v5.0.3 RTM

### LINKED ISSUE #80: FEATURE: Fallback policy
As I understand it today, if I use Polly to setup a circuit breaker in my policy and that the circuit does break, I will receive a BrokenCircuitException. As a user of the framework, I need to handle this exception and write some code to specify how I want to react to this event. I would prefer to give to Polly that code and let it execute it if the circuit is broken. It would be better performance-wise, and much cleaner.

```
Policy
  .Handle<DivideByZeroException>()
  .CircuitBreaker(2, TimeSpan.FromMinutes(1))
  .WhenBroken(
  () => throw new BrokenCircuitException(); //Or another fallback method specific to my app.
)
```

### LINKED ISSUE #137: FEATURE: Timeout policy
# Proposal: Timeout policy
## Purpose

To provide a timeout for goverened actions.  Acts both as a user-experience enhancing and resilience strategy.
## Scope
- A **non-exception handling** **resilience** policy.
- Support all sync and async variants
- Support non-result-returning invocations (`Action`) and result-returning (`Func<TResult>`)
- Support `Execute(...)` and `ExecuteAndCapture(...)` variants.  
- Add cancellation support for sync variants.
## Configuration syntax

``` cssharp
TimeoutPolicy timeoutPolicy = Policy
  .Timeout(TimeSpan timeout);

TimeoutPolicy<TResult> timeoutPolicy = Policy<TResult>
  .Timeout(TimeSpan timeout);
```

(and similar async forms)
## Execution principles and syntax

The first intention of a timeout is to return back to the caller within a defined period, on the assumption the underlying operation appears not likely to succeed.

A second (but not secondary!) consideration is, in co-operation with the timed-out operation, to release resources used by the timed-out operation.  _Not_ actively releasing timed-out resources could lead to a build up of unreleased resources - especially in circumstances where throughput means new similar requests would immediately take the place of the timed-out ones. 
#### Adding synchronous delegate execution with cancellation

For similar reasons, Polly will not unilaterally terminate/abort threads of timed-out synchronous actions: doing so could leave resources unreleased.  Polly should only engage in co-operative cancellation.  This means we need to add new overloads of the sync `.Execute()` delegates taking a `CancellationToken`:

``` cssharp
// TResult sync form, synchronous with cancellation 
TResult result = timeoutPolicy
  .Execute(Func<CancellationToken, TResult> action, CancellationToken token);

// void sync form with cancellation
timeoutPolicy
  .Execute(Action<CancellationToken> action, CancellationToken token);
```

These (and similar) new overloads will have to be supported by _all_ policies, in order for cancellation on sync calls to be able to take effect across a `PolicyWrap`.
#### Asynchronous delegate execution with cancellation

Cancellation support is already in place for the async forms: 

```
// TResult async form 
TResult result = timeoutPolicy
  .ExecuteAsync(Func<Task<CancellationToken, TResult>> executionFunc, CancellationToken token);

// void async form
timeoutPolicy
  .ExecuteAsync(Func<CancellationToken, Task> executionFunc, CancellationToken token);
```
#### Decision: Enforce cancellation usage?

It would be an option to _enforce_ the use of overloads taking `CancellationToken`s when a timeout policy is in use.  Having the `TimeoutPolicy` throw an `InvalidOperationException` on overloads of `Execute/Async()` not taking a `CancellationToken`, while an aggressive approach, would make clear Polly's recommendation.   

There will be circumstances where users have no control over invoked code, where cancellation support in the invoked delegates is not possible.  If overloads taking a `CancellationToken` were enforced, such code could still be executed by ignoring the `CancellationToken` (but it would possibly make for ugly-reading code).

Views?
## Operation
- Use `CreateLinkedTokenSource()` to combine the caller's cancellation token with a timeout cancellation token such that the user delegate can be cancelled by either. 
- Invoke `executionFunc`.  
- If the configured timeout elapses before `executionFunc` returns:
  - cancellation token cancels to allow `executionFunc` to terminate quickly / release resources.
  - throw an `ExecutionTimeout : ExecutionRejected` exception.
- If `executionFunc` raises any kind of fault: rethrow / pass fault result to caller.  
# Comments?

Comments?  Alternative suggestions?  Extra considerations to bear in mind?  

Have scenarios to share where you'd use this? (could inform development)

### LINKED ISSUE #138: FEATURE: Bulkhead isolation policy
# Proposal: Throttle policy
## Purpose
- To limit the number of concurrent parallel invocations through the policy.  
- To isolate groups of actions (bulkhead isolation) such that the impact of faulting calls is isolated to the resource-pool to which it is limited, and other threads/pools/capacity remain to continue serving other calls.
## Scope
- A **non-exception handling** **resilience** strategy.
- Support all sync and async variants
- Support non-result-returning invocations (`Action`) and result-returning (`Func<TResult>`)
- Support `Execute(...)` and `ExecuteAndCapture(...)` variants.  
#### Configuration syntax

``` cssharp
ThrottlePolicy throttlePolicy = Policy
  .ThrottleAsync(int maxParallelization[, int maxQueuedActions]);

ThrottlePolicy<TResult> throttlePolicy = Policy<TResult>
  .ThrottleAsync(int maxParallelization[, int maxQueuedActions]);
```
#### Execution syntax

```
// TResult async form 
TResult result = throttlePolicy
  .ExecuteAsync(Func<Task<CancellationToken, TResult>> executionFunc, CancellationToken token);

// void async form
throttlePolicy
  .ExecuteAsync(Func<CancellationToken, Task> executionFunc, CancellationToken token);
```
#### Implementation
-  a pair of `SemaphoreSlim`s
#### Operation
- Attempt to acquire a semaphore limited to maxParallelization + maxQueuedActions with no delay.
  - Throw `ThrottleQueueFullException : ExecutionRejectedException` if cannot acquire semaphore  immediately.
- Attempt to acquire a semaphore limited to maxParallelization (**UPDATE**: timeout from any wrapping timeout policy can be automatically handled by the `CancellationToken` passed in)
  - **EDIT** wait until can acquire semaphore (unless timed out)
- (once semaphore acquired) Execute `executionFunc`
- If `executionFunc` raises any kind of fault: rethrow / pass fault result to caller.  
#### Implementation and Operation - Semaphore form
- As for async.  
- **EDIT**: Edited to remove `TaskScheduler`/`ThreadPool` proposal, which will not be taken forward for v5.0.
## Naming

Possible names could include `Throttle`, `ParallelThrottle`, `MaxConcurrency`, `Isolate`?  Typical (_but not the only_) usage is bulkhead isolation; the effective operation (the undisputable thing) is to limit parallelization to a maximum concurrency.
# Comments?

Comments?  Alternative suggestions?  Extra considerations to bear in mind?  

Scenarios where you'd use this? (please share now - could inform development!)

### LINKED ISSUE #139: FEATURE: Add KEYs to policies and executions
# Proposal: Add ids/keys to policies and executions

Ids or Keys could be useful in Polly for a number of functions, and at a number of levels of granularity.
## PolicyWrapKey

| Key type | PolicyWrapKey |
| --- | --- |
| Description / Purpose | A unique key for each Policy Wrap.  While PolicyKey varies with each policy in a wrap, PolicyWrapKey remains constant over the wrap. |
| Primary use | Retrieve a Policy Wrap from Registry? |
| Additional use | Include in emitted events/stats, to indicate policywrap in use |
| Default if not set | short-type-name+some-guid-part?; null if execution not part of a pipeline |
| Behaviour in PolicyWrap | Retains own value |
| How would it be specified? | extension method `.WithPolicyWrapKey(string policyWrapKey)` on `PolicyWrap`, but set-once-only enforced |
| How should it be exposed? | Defined as `PolicyWrap.PolicyWrapKey`?; exposed to execution as `Context.PolicyWrapKey` |
## PolicyKey

| Key type | Policy Key |
| --- | --- |
| Description / Purpose | A unique key for an individual configured Policy instance |
| Primary use | Retrieve a policy from registry |
| Additional use | Include in emitted events/stats, to indicate policy in use |
| Default if not set | short-type-name+some-guid-part ? |
| Behaviour in PolicyWrap | Retains own value |
| How would it be specified? | extension method `.WithPolicyKey(string policyKey)` on `Policy`, but set-once-only enforced |
| How should it be exposed? | Defined as `Policy.PolicyKey`? Exposed to execution as `Context.PolicyKey` |
## ExecutionKey

| Key type | ExecutionKey |
| --- | --- |
| Description / Purpose | A unique key for each _call site usage_ of a policy.  (A single policy instance may be re-used in multiple code locations (see [wiki](https://github.com/App-vNext/Polly/wiki/Some-policy-patterns))). |
| Primary use | Used by `CachePolicy` as the cache key for the given execution |
| Additional use | Include in emitted events/stats, to indicate _where_ policy in use |
| Default if not set | null? |
| Behaviour in PolicyWrap | Takes value passed by any preceding policy in the wrap |
| How would it be specified? | by passing a `new Context(executionKey)` with the Execute call |
| How should it be exposed? | `Context.ExecutionKey` |
## ExecutionGuid

| Key type | ExecutionGuid |
| --- | --- |
| Description / Purpose | A unique GUID for each _invocation_ through a policy |
| Primary use | Distinguish individual calls through the policy, so that stats-interpreting code can distinguish calls in multi-threaded environment |
| Additional use | . |
| Default if not set | Guid.NewGuid() |
| Behaviour in PolicyWrap | Takes value passed by any preceding policy in the wrap |
| How would it be specified? | not for the coder to specify - generated randomly by each call to Execute |
| How should it be exposed? | `Context.ExecutionGuid` |
## Implementation

Every single call through a Polly policy carries an instance of `Context` with it: this will carry the values of the key for each call.  
# Comments?

Comments?  Alternative suggestions?  Extra considerations to bear in mind?  Do we need any other keys?  What do comparable products provide in their stats?

### LINKED ISSUE #140: FEATURE: Policy Wrap ( / pipeline)
# Proposal: Policy Wrap (was: Pipeline)
## Purpose

To provide the ability to chain a series of policies together.
## Scope
- A meta-policy.
- All sync and async variants
- When result-returning policies placed in a wrap, all policies must return the same `TResult`
- _Or_ void-returning (`Action`-executing) policies may be wrapped; all polices in the wrap must return `void`.
- Individual policies should be able to be used in more than one wrap. 

---
# Original proposal: Policy Pipeline

The [roadmap](https://github.com/App-vNext/Polly/wiki/Polly-Roadmap) originally proposed the following syntax:

```
// Possible syntax
Policy
  .Pipeline(params IEnumerable[] policies)
  .Execute(...) // Executes the action through the pipeline of policies, each policy wrapping the next.  
```

Some issues   were:
- The term **pipeline** very much connotes a one-way flow.  In exception-handling this precisely _isn't_ the case: exceptions thrown would then be passed 'backwards back through the pipeline' (? - ugly broken metaphor).
- The  syntax isn't conducive to flexibly composing pipelines by functional composition.
# Revised proposal: Policy Wrap

The revision proposes the name **Policy Wrap**:
- The term **wrap** more clearly connotes outer and inner layers, and the idea of _wrapping_ a function with another (what is actually happenning).
- It's the classic 'onion-layers' metaphor, with outer policies executing next-inner policies, executing next-inner, until getting to execute the user delegate (at the 'core' of the onion). 
- Easier to grok what happens with exceptions: they get thrown back outwards through the layers. 
- The term wrap plays better with a flexibly-composed, functionally-composed policy - see below. 
## Syntax proposal

Both static and instance configuration methods are now proposed:

```
// Static configuration syntax as before:
PolicyWrap policyWrap = Policy.Wrap(fallback, cache, retry, circuitBreaker, timeout, throttle); 
policyWrap.Execute(myAction);

// Instance syntax
Policy retry = ...       // some retry policy
Policy breaker = ...     // some breaker
Policy timeout = ...     // some timeout

retry.Wrap(breaker).Execute(myAction); // Inline wrap usage showing simple functional composition

retry.Wrap(breaker).Wrap(throttle).Execute(myAction); // Inline wrap usage chaining three policies
retry.Wrap(breaker.Wrap(throttle)).Execute(myAction); // (functionally equivalent to preceding line)

// Possible use of instance syntax configuring a policy for later use
PolicyWrap myWrap = fallback.Wrap(cache).Wrap(retry).Wrap(breaker).Wrap(timeout).Wrap(throttle);
PolicyWrap myWrap = fallback.Wrap(cache.Wrap(retry.Wrap(breaker.Wrap(timeout.Wrap(throttle))))); // (functionally equivalent to preceding)


```
#### Value of functional composition

With the original 'one-shot' static syntax, it was not intuitive how to create a pipeline that was a slight variation on another.  It could lead to questions about 'how to [consistently] insert something into the middle of an IEnumerable/ICollection'.  Mutable pipelines also potentially have issues in multi-threaded, concurrent scenarios.

The instance syntax of `.Wrap(...)` allows more flexible functional-composition of a PolicyWrap:

```
Policy myWrap = ...            // some existing wrap or policy - maybe even passed in to this code from elsewhere
bool useCache = ...            // some value, maybe passed in
CachePolicy cachePolicy = ...  // from somewhere
myWrap = useCache ? cachePolicy.Wrap(myWrap) : myWrap;
// etc
```

The functional-composition paradigm matches LINQ and Rx: successively composing functional transformations on the function-so-far, before actual execution.
## Operation
- Executes the supplied user delegate through the layers or wrap: the outermost policy executes the next inner, which executes the next inner, etc, until the innermost policy executes the user delegate.
- Exceptions bubble back outwards (until handled) through the layers.
- The `Context` instance carries `Key` information unifying the call: see Keys proposal #139 for details.
- Care needed around implementations for `ExecuteAndCapture()`.  Probably only the outermost policy in the wrap should `ExecuteAndCapture()`; others should `Execute()` -and-throw.  
## Naming

Any alternative naming suggestions?  
# Comments?

Comments?  Alternative suggestions?  Extra considerations to bear in mind?    

Have scenarios to share where you'd use this? (could inform development)

### LINKED ISSUE #142: Polly v4.x.x will be the last to support .NET3.5
Polly v4.x.x will be the last to support .NET3.5.  Polly v5.0.0 and onwards will not offer .NET3.5 versions.

The new resilience features in progress for Polly v5 require features (`TaskScheduler`, `SemaphoreSlim`, `CancellationToken`) not available in .NET3.5.  In particular, a Timeout policy #137 supporting co-operative cancellation by `CancellationToken` for _synchronous_ delegates - and the ability to use such a Timeout policy wrapping others within a `PolicyWrap` - will require that _all_ policies support `CancellationToken` in their synchronous implementations.  

In case of the need for bug-fixes to v4.x.x versions used by .NET3.5 users, a separate v4.x.x or PollyNet35 branch can be maintained in GitHub, and separate `Polly.Net35` packages issued, if necessary.

Comments/questions welcome below.

### LINKED ISSUE #148: Provide all .NET40 support through the .NET40Async package from Polly v5.0 onwards
.NET4.0 support for Polly is currently available in two packages:  (and their corresponding signed versions)

[1] via the [main Polly package](https://www.nuget.org/packages/Polly/) (with no async support)
[2] via the [Polly.Net40Async package](https://www.nuget.org/packages/Polly.Net40Async/) (with async support, via a dependency on [Microsoft.Bcl.Async](https://www.nuget.org/packages/Microsoft.Bcl.Async/))

A simplification would be to provide all .NET40 support through [2] the Polly.Net40Async package.

With [the discontinuation of .NET35 support from Polly v5.0](https://github.com/App-vNext/Polly/issues/142), .NET4.0 in the main Polly package becomes the only remaining non-async target.  Removing the .NET4.0 no-async offering would simplify the Polly codebase by removing all forking around async.

A disadvantage would be for .NET4.0 users not using async features, who would be forced to take a dependency on Microsoft.Bcl.Async.  The preferred implementation for certain features of Polly v5.0 may however also draw on `Task` features only found for .NET4.0 on `TaskEx` (such as the `Delay(...)` method), which may also drive Polly for .NET4.0 towards a Bcl.Async dependency from v5.0.

_Do .NET4.0 users have a view on this?  Is a Microsoft.Bcl.Async dependency an issue for you?  Or is it such a commonplace that it makes no difference?_

### LINKED ISSUE #159: Make the assembly CLS-Compliant
Is there reason that Polly isn't marked as CLS-Compliant?

For further info on CLS-Complaince, see https://msdn.microsoft.com/en-us/library/bhc3fa7f.aspx

### LINKED ISSUE #164: Polly v5.0 is now in alpha on Nuget
Per the [project board](https://github.com/App-vNext/Polly/projects/1), [PolicyWrap](https://github.com/App-vNext/Polly/issues/140), [ExecutionKeys](https://github.com/App-vNext/Polly/issues/139), [Fallback policy](https://github.com/App-vNext/Polly/issues/80), [Bulkhead isolation](https://github.com/App-vNext/Polly/issues/138) and [Timeout policy](https://github.com/App-vNext/Polly/issues/137) are now all in alpha, delivered into the [v5.0-alpha branch](https://github.com/App-vNext/Polly/tree/v5.0-alpha).

This issue signals our intention to make an alpha-release of this material to Nuget in week-beginning 24 October (or earlier).

Main pre-requisite: documentation of the new policies (underway) in both [readme](https://github.com/App-vNext/Polly/blob/v5.0-alpha/README.md) and [wiki](https://github.com/App-vNext/Polly/wiki).

### LINKED ISSUE #174: Update NuSpec to say .NET Standard

### LINKED ISSUE #176: NetStandard1.0 dll in v5.0-alpha nupkg [Polly5.0.2] is stamped 0.0.0.0
The [GitVersionTask](https://github.com/GitTools/GitVersion) tool we use to stamp assemblies with version number does not yet support assemblies targeting .NET Standard.  
 
+ Explicitly stated as not compatible when nuget-installing it to such a project.  
+ [Some related discussion](https://github.com/GitTools/GitVersion/issues/647)
 
**This means that Polly.dll in our .NETStandard-targeting Nuget packages is currently version-numbered 0.0.0.0**.  (Nuget package number is correct; AssemblyVersion info on Polly.dll (the dll targeting NetStandard 1.0 _only_) reads 0.0.0.0)
 
Options:
 
[1] Continue to live with this
[2] Keep `[assembly: AssemblyVersion("5.0.2.0")]` in `Polly.NetStandard10\AssemblyInfo.cs` manually in sync with `GitVersionConfig.yaml` (two sources of truth) (until GitVersionTask fix this)
[3] **[preferred for now?]** Add some automated regex replace step in the build, to copy the version number from `GitVersionConfig.yaml` (single source of truth) to `Polly.NetStandard10\AssemblyInfo.cs` (until GitVersionTask fix this)
[4] Switch to https://github.com/SimpleGitVersion/SGV-Net.  Claims to handle this scenario, but has more limited following than GitVersionTask.
 
Clearly, with the whole project.json / csproj tooling in flux, we can expect to have to change this again - any solution is temporary.

GitVersionTask folks are (probably wisely) holding off fixing, until the Net Standard project formats/tooling settle.

### LINKED ISSUE #179: Async Bulkhead implementation is wrongly blocking (.NET40 only)
As stated [here](https://github.com/App-vNext/Polly/issues/138#issuecomment-252042050), the async Bulkhead implementation is currently blocking (while waiting to acquire an execution slot in the bulkhead) for .NET40 only.  

This is due to `System.Threading.SemaphoreSlim` not offering a `WaitAsync(...)` method in .NET4.0

(reference issue: PR coming to fix this in next 24-48 hours)

### LINKED ISSUE #188: MINOR BREAKING CHANGE: Polly policies should no longer cancel-after-delegate-completion
# Scenario 

+ A policy executes a delegate which **does not observe cancellation**.  
+ During execution of the user delegate, **cancellation is signalled** (but the delegate does not observe it).  
+ The delegate runs to completion returning a success result (to the policy); or runs to completion returning a fault result (to the policy); or throws some other exception.

## Existing behaviour (Polly 3.0.0 to 5.0.2 alpha) 

Rather than return the result of the completed user delegate, policies aggressively throw `OperationCanceledException` (after delegate completion) for the cancellation which the user delegate did not observe.  

## New behaviour (Polly v5.0.3 RTM onwards)

Return the result of the completed user delegate, or propagate the exception it raised, rather than override with a cancellation the user delegate did not honour.

### Why?

Subsequent code may take critical decisions based on whether the user delegate completed.  Aggressively _throwing-for-cancellation-after-success_ (the current behaviour) misrepresents what happened, in the sense that it communicates the user delegate did not succeed/complete, when in fact it did.  

For parity, Polly will also now return/propagate the fail result rather than aggressively throw-for-cancellation-after-failure.

Both revised stances more truly reflect that the executed delegate did not observe cancellation.  
 
## Comments?

We do our best to avoid breaking changes in Polly (none in fifteen releases), and to document clearly when they occur. This minor case, driven by user feedback/experience, does seem like a good candidate for change.  Comments welcome tho?

reisenberger:
Merge v5.0-alpha to master, to create v5.0.3 RTM

+ Closes #80 : Feature: `Fallback` policy
+ Closes #137: Feature: `Timeout` policy
+ Closes #138 : Feature: `Bulkhead` isolation policy
+ Closes #139 : Add execution keys
+ Closes #140 : Feature: `PolicyWrap`
+ Closes #142 : Discontinue .NET3.5 support
+ Closes #148 : Move .NET40 support to .NET40async package
+ Closes #159 : Label Polly.dll as CLS-compliant (.NET40 version excepted)
+ Closes #164 : v5.0-alpha announcement
+ Closes #174 : Reference NetStandard correctly in NuSpec
+ Closes #176 : Semver-stamp NetStandard dll correctly
+ Closes #179 : Fix async Bulkhead NET40 
+ Closes #188 : Refine cancellation


joelhulen:
Merged. This is a big one! Thanks for everything you do, @reisenberger 💯

PR #207, opened 2017-01-14
Removed InvalidOperationException from initial _lastOutcome of CircuitStateController

lakario:
CircuitStateController.._lastOucome is now initialized with a default…(TResult), instead of an InvalidOperationException

### DIFF
--- a/src/Polly.Shared/CircuitBreaker/CircuitStateController.cs
+++ b/src/Polly.Shared/CircuitBreaker/CircuitStateController.cs
@@ -114,7 +114,7 @@ namespace Polly.CircuitBreaker
         protected void ResetInternal_NeedsLock(Context context)
         {
             _blockedTill = DateTime.MinValue;
-            _lastOutcome = new DelegateResult<TResult>(new InvalidOperationException("This exception should never be thrown"));
+            _lastOutcome = new DelegateResult<TResult>(default(TResult));
 
             CircuitState priorState = _circuitState;
             _circuitState = CircuitState.Closed;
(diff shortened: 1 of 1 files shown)


lakario:
FYI @reisenberger I couldn't find any places where the usage of the LastException needed to be guarded. All tests pass, so I believe this is the only necessary change.

lakario:
Alright, fixed.

lakario:
Weird, it looks like the tests didn't complete?

reisenberger:
@lakario Many thanks for this!  (Your last build simply fell foul of some timing sensitivities in unrelated tests (slower running on AppVeyor); I have hopefully squashed that out in #212 .) 

Code of PR looks good!  Considering when to merge, pending the `v5.0.4-pre` Net4.0Async fixes which will go out just ahead.  

Shall we add the following documentation additions, to clarify?  (Feel free to cross-check / reword if appropriate!)

https://github.com/lakario/Polly/blob/master/src/Polly.Shared/CircuitBreaker/CircuitBreakerPolicy.cs#L34 : Add `<remarks>This will be null when the circuit-breaker is closed.</remarks>`

https://github.com/lakario/Polly/blob/master/src/Polly.Shared/CircuitBreaker/CircuitBreakerPolicy.cs#L85 : `<remarks>This will be null when the circuit-breaker is closed, or when the circuit-breaker is open but circuit-breaking was triggered by a handled result.</remarks>`

https://github.com/lakario/Polly/blob/master/src/Polly.Shared/CircuitBreaker/CircuitBreakerPolicy.cs#L92 : `<remarks>This will be default(TResult) when the circuit-breaker is closed, or when the circuit-breaker is open but circuit-breaking was triggered by a handled exception.</remarks>`

Thanks for your input into the original issue!

reisenberger:
Thanks @lakario !  Merging!

PR #214, opened 2017-01-17
Add NoOp Policy to Polly

lakario:
Created a no op policy which is functionally similar to .Retry(0). @reisenberger Please let me know if more unit tests are needed, this policy doesn't _do_ much. Cheers.


reisenberger:
Thanks @lakario !  Excellent job following the Polly internals!  :+1: 

Had thought: we can remove even the 'cancellationToken.ThrowIfCancellationRequested();'.  Make the 'no op' concept completely pure - does _nothing_ except exec user delegate.

Happy to slip this into the v5.0.5 release if you are good for the amends

Thanks again!

src/Polly.Shared/NoOp/NoOpEngine.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Threading;
+
+namespace Polly.Shared.NoOp
+{
+    internal static partial class NoOpEngine
+    {
+        internal static TResult Implementation<TResult>(Func<CancellationToken, TResult> action,
+            CancellationToken cancellationToken)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
```
reisenberger: We could remove `cancellationToken.ThrowIfCancellationRequested();` to make the "noop" concept completely pure.
(and similar elsewhere where same pattern)

src/Polly.Shared/NoOp/NoOpEngine.cs
```diff
@@ -0,0 +1,21 @@
...
+
+            DelegateResult<TResult> delegateOutcome = 
+                new DelegateResult<TResult>(action(cancellationToken));
```
reisenberger: Could remove the `DelegateResult<TResult>` allocation.  The `delegateOutcome.Result` below only returns the `action(cancellationToken)` that has just been passed in to the ctor.  

(similar where same pattern on async side)

src/Polly.SharedSpecs/NoOp/NoOpAsyncSpecs.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        [Fact]
+        public void Should_not_throw_when_executing()
+        {
+            var policy = Policy.NoOpAsync();
+
+            policy.Awaiting(async p => await p.ExecuteAsync(() => Task.FromResult(0)))
+                .ShouldNotThrow();
+        }
+
```
reisenberger: On the unit-test side, could be nice to have four-line(?) test expressing that `NoOpPolicy` does actually execute the delegate.  `bool executed = false;` flipped to `true` by executing deleg through policy

(similar elsewhere across async / TResult pattern)

src/Polly.Shared/NoOp/NoOpEngine.cs
lakario: Sure, can do. I'll make an update today.

lakario:
@reisenberger Updated per your input. I'm using a "bool executed" for non-TResult usages, and simply asserting the TResult for those with one.

reisenberger:
Thanks @lakario ! AppVeyor reports: `error CS0266: Cannot implicitly convert type 'int?' to 'int'. An explicit conversion exists (are you missing a cast?) `.  Do you want to have a look? 

:+1: clearing `.Shared` out of namespaces.  Spotted yest, forgot to mention.  VS2015 adds that into any classes freshly created, due to the Shared projects approach.

src/Polly.Shared/NoOp/NoOpEngineAsync.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Threading.Tasks;
+
+namespace Polly.NoOp
+{
+    internal static partial class NoOpEngine
+    {
+        internal static async Task<TResult> ImplementationAsync<TResult>(Func<CancellationToken, Task<TResult>> action,
+            CancellationToken cancellationToken)
+        {
+            return await action(cancellationToken);
```
reisenberger: @lakario  Additional tiny spot: we want to add a `.ConfigureAwait(continueOnCapturedContext)` here.  We place these on every `await`, to guarantee we honour the user's choice of `continueOnCapturedContext` all through an async call.

(I've been using [Resharper plugin](https://resharper-plugins.jetbrains.com/packages/ConfigureAwaitChecker.v9) to monitor these ; we are also considering adding the [relevant Roslyn analyzer](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.ApiDesignGuidelines.Analyzers/Microsoft.ApiDesignGuidelines.Analyzers.md#ca2007-do-not-directly-await-a-task) in the CI build)

lakario:
@reisenberger All concerns have been addressed.

reisenberger:
Thanks @lakario for this valuable contribution!  Very useful for those who want a quick way to stub-out Polly for unit-testing.  Also forms a great outline of what is needed to create a new policy.

Merging ... :smiley:

src/Polly.SharedSpecs/NoOp/NoOpTResultAsyncSpecs.cs
```diff
@@ -0,0 +1,47 @@
...
+
+            policy.Awaiting(async p => result = await p.ExecuteAsync(() => Task.FromResult((int?)10)))
+                .ShouldNotThrow();
+
+            result.HasValue.Should().BeTrue();
+            result.Should().Be(10);
+        }
+
+        [Fact]
+        public void Should_execute_if_cancellationtoken_cancelled_before_delegate_reached()
```
lakario: Just noticed I forgot to update the name on this spec. It should _not_ cancel! @reisenberger

reisenberger: @lakario Just adding `NoOpPolicy` doco to the readme and changelog, will look over spec name in the same commit (no probl)

Issue #226, opened 2017-02-07
PROPOSAL: PolicyRegistry


reisenberger:
A `PolicyRegistry` with `Dictionary<string, Policy>`-like semantics, for storing configured policies and retrieving them for use elsewhere.

Useful for those who want to define policies centrally on start-up, then provide them to components from a registry, by eg dependency-injection or ambient context pattern.

_Thanks to @ankitbko for discussion which influenced the proposal!_

### Hide underlying implementation

`PolicyRegistry : Dictionary<string, Policy>` while simple would tie us into an implementation.  Better a shallow wrapper, something [like this](https://github.com/BrighterCommand/Paramore.Brighter/blob/master/paramore.brighter.commandprocessor/PolicyRegistry.cs), allowing changing underlying implementation.

### `IDictionary<TKey, TValue>` semantics, or custom?

Should we have custom-semantics (as eg [this shallow wrapper](https://github.com/BrighterCommand/Paramore.Brighter/blob/master/paramore.brighter.commandprocessor/PolicyRegistry.cs))?  Or should we implement `IDictionary<TKey, TValue>`, for entirely standard semantics?  

_Suggest_: standard semantics?

#### `Dictionary<TKey, TValue>`

If we accept the majority Use Case is likely to be establish policy registry at application startup on a single-thread, read on multiple threads thereafter (ie no concurrent writes), **`Dictionary<TKey, TValue>`** would be adequate; possibly also (fractionally) most read-performant.  

If this, **document** that the default implementation is optimised for this majority Use Case but not safe for concurrent writes.  

We could also provide segregated `IReadOnlyPolicyRegistry` and `IWritablePolicyRegistry` interfaces; recommend pass around only an `IReadOnlyPolicyRegistry` after populating the registry.

#### `ImmutableDictionary<TKey, TValue>`

**`ImmutableDictionary<TKey, TValue>`**: the best fit for write-extensively-at-startup, read-mostly thereafter (but still safe for occasional writes).  Requires tho Polly taking a dependency on `System.Collections.Immutable` (violates no dependencies principle).  Offer this implementation (if demand) as a separate nuget package?

#### `ConcurrentDictionary<TKey, TValue>` 

[...]

PR #231, opened 2017-02-20
Added PolicyRegistry Feature

ankitbko:
Implementation of PolicyRegistry Proposal #226


src/Polly.SharedSpecs/Registry/DefaultPolicyRegistrySpecs.cs
```diff
@@ -0,0 +1,238 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+using FluentAssertions;
+using Polly.Registry;
+using Polly.NoOp;
+
+namespace Polly.SharedSpecs.Registry
```
ankitbko: Completely missed this one out. Fixed.

src/Polly.SharedSpecs/Registry/DefaultPolicyRegistrySpecs.cs
```diff
@@ -0,0 +1,238 @@
...
+    {
+        IPolicyRegistry<string, Policy> _registry;
+
+        public DefaultPolicyRegistrySpecs()
+        {
+            _registry = new DefaultPolicyRegistry();
+        }
+
+        [Fact]
+        public void Should_Allow_Adding_Policy_Using_Add()
```
ankitbko: Done. Didn't knew about this. I also tend to keep things consistent and follow the practice of the project. Will keep this in mind in my future contributions. 😁

src/Polly.SharedSpecs/Registry/DefaultPolicyRegistrySpecs.cs
```diff
@@ -0,0 +1,238 @@
...
+        {
+            Policy policy = Policy.NoOp();
+            string key = Guid.NewGuid().ToString();
+            _registry.Invoking(r => r.Add(key, policy))
+                .ShouldNotThrow();
+
+            _registry.Count.Should().Be(1);
+
+            policy = Policy.NoOp();
+            key = Guid.NewGuid().ToString();
```
ankitbko: Wouldn't make any difference in behavior as I was re-initializing the variables but I see your point. I actually had done the same before but removed it later. 😞 Will make the changes.

src/Polly.SharedSpecs/Registry/DefaultPolicyRegistrySpecs.cs
```diff
@@ -0,0 +1,238 @@
...
+            _registry.Add(key, policy);
+
+            _registry.Invoking(r => outPolicy = r[key])
+                .ShouldNotThrow<KeyNotFoundException>();
+
+            outPolicy.ShouldBeEquivalentTo(policy);
+        }
+
+        [Fact]
+        public void Should_Not_Throw_While_Retrieving_When_Key_Does_Not_Exist_Using_TryGetValue()
```
ankitbko: Added. I also added more test cases to check `ArgumentNullException` is thrown when key is null while adding, removing and retrieving policy.

src/Polly.SharedSpecs/Registry/DefaultPolicyRegistrySpecs.cs
```diff
@@ -0,0 +1,238 @@
...
+            bool result = false;
+
+            _registry.Invoking(r => result = r.TryGetValue(key, out outPolicy))
+                .ShouldNotThrow();
+
+            result.Should().BeFalse();
+        }
+
+        [Fact]
+        public void Should_Overwrite_Existing_Policy_If_Key_Exists_When_Inserting_Using_Idexer()
```
ankitbko: Fixed!

src/Polly.SharedSpecs/Registry/DefaultPolicyRegistrySpecs.cs
```diff
@@ -0,0 +1,238 @@
...
+            key = Guid.NewGuid().ToString();
+
+            _registry.Invoking(r => result = r.ContainsKey(key))
+                .ShouldNotThrow();
+
+            result.Should().BeFalse();
+        }
+
+        [Fact]
+        public void Should_Overwrite_Existing_Policy_When_Using_Indexer()
```
ankitbko: Ah.. Missed to delete it. Done now.

src/Polly.SharedSpecs/Registry/DefaultPolicyRegistrySpecs.cs
```diff
@@ -0,0 +1,238 @@
...
+        public void Should_Check_If_Key_Exists()
+        {
+            Policy policy = Policy.NoOp();
+            string key = Guid.NewGuid().ToString();
+            _registry.Invoking(r => r.Add(key, policy))
+                .ShouldNotThrow();
+
+            bool result = false;
+            _registry.Invoking(r => result = r.ContainsKey(key))
+                .ShouldNotThrow();
```
ankitbko: I have removed `.ShouldNotThrow()` which were redundant unless, as you mentioned, to make a contrast with elsewhere in the test (`Should_not_be_able_to_add_Policy_with_duplicate_key_using_Add`) and where the test explicitly says Should_not_throw (`Should_not_throw_while_retrieving_when_key_does_not_exist_using_TryGetValue`)

PS: I hadn't worked with FluentAssertion much before. Thanks for helping me out with best practices. Really appreciate it.

src/Polly.Shared/Registry/IPolicyRegistry.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// <typeparam name="Policy">The type of Policy to store. Must have <see cref="Polly.Policy"/> as base class. </typeparam>
+    public interface IPolicyRegistry<Key, Policy> where Policy: Polly.Policy
+    {
+        /// <summary>
+        /// Gets or sets policy with specified Key.
+        /// </summary>
+        /// <param name="key">The key of the policy to get or set.</param>
+        /// <returns>The policy with specified Key.</returns>
+        /// <exception cref="System.ArgumentNullException">Key is null.</exception>
+        /// <exception cref="KeyNotFoundException">The property is retrieved and key is not found.</exception>
```
ankitbko: Haha!!! Nice catch. To start with I had copied the comments from `IDictionary` since we had earlier inherited from it. Seems even `IDictionary` has a lot of mistakes. 😁 Fixed.
image

src/Polly.Shared/Registry/IPolicyRegistry.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <exception cref="ArgumentNullException">key is null.</exception>
+        bool Remove(Key key);
+
+        /// <summary>
+        /// Gets the Policy associated with the specified key
+        /// </summary>
+        /// <param name="key">The key whose value to get</param>
+        /// <param name="value">
+        /// When this method returns, the Policy associated with the specified key, if the
+        /// key is found; otherwise, the default value for the type of the value parameter.
```
ankitbko: Done. Default value for reference types would have [returned null](https://msdn.microsoft.com/en-us/library/bb347013(v=vs.110).aspx#Anchor_2). Making it explicit makes it more clear.

src/Polly.SharedSpecs/Registry/DefaultPolicyRegistrySpecs.cs
```diff
@@ -0,0 +1,238 @@
...
+            bool result = false;
+
+            _registry.Add(key, policy);
+
+            _registry.Invoking(r => result = r.TryGetValue(key, out outPolicy))
+                .ShouldNotThrow<ArgumentException>();
+
+            result.Should().BeTrue();
+
+            outPolicy.ShouldBeEquivalentTo(policy);
```
ankitbko: Done.

src/Polly.SharedSpecs/Registry/DefaultPolicyRegistrySpecs.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            //indexer
```
ankitbko: Removed.

src/Polly.SharedSpecs/Registry/DefaultPolicyRegistrySpecs.cs
```diff
@@ -0,0 +1,238 @@
...
+            _registry.Count.Should().Be(1);
+
+            policy = Policy.NoOp();
+            key = Guid.NewGuid().ToString();
+            _registry.Invoking(r => r.Add(key, policy))
+                .ShouldNotThrow();
+
+            _registry.Count.Should().Be(2);
+
+            //Using Indexer
```
ankitbko: Removed.

src/Polly.Shared/Registry/DefaultPolicyRegistry.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// Determines whether the specified Key exists.
+        /// </summary>
+        /// <param name="key">The Key to locate in the registry</param>
+        /// <returns>True if Key exists otherwise false</returns>
+        /// <exception cref="ArgumentNullException"><paramref name="key"/> is null</exception>
+        public bool ContainsKey(string key) =>
+            _registry.ContainsKey(key);
+
+        /// <summary>
+        /// Removes the specified <see cref="Policy"/> from the registry
```
ankitbko: Corrected.

src/Polly.Shared/Registry/IPolicyRegistry.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Collections.Generic;
+
+namespace Polly.Registry
+{
+    /// <summary>
+    /// Represents a collection of <see cref="Polly.Policy"/> keyed by <typeparamref name="Key"/>
+    /// </summary>
+    /// <typeparam name="Key">The type of keys in the dictionary</typeparam>
+    /// <typeparam name="Policy">The type of Policy to store. Must have <see cref="Polly.Policy"/> as base class. </typeparam>
+    public interface IPolicyRegistry<Key, Policy> where Policy: Polly.Policy
```
ankitbko: Done

src/Polly.Shared/Registry/DefaultPolicyRegistry.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// <summary>
+    /// Stores a collection of <see cref="System.String"/> and <see cref="Policy"/> pair.
+    /// </summary>
+    /// <remarks>Uses ConcurrentDictionary to store the collection.</remarks>
+    public class DefaultPolicyRegistry : IPolicyRegistry<string, Policy>
+    {
+        private IDictionary<string, Policy> _registry = new ConcurrentDictionary<string, Policy>();
+
+        /// <summary>
+        /// Gets or sets the value associated with the specified key.
```
ankitbko: Done.

reisenberger:
@ankitbko Apologies for my delay replying (due to volume of paid work!).  Many thanks for the revisions!   This is perfect for `PolicyRegistry` for policies of non-generic type `Policy`!  

When travelling two weeks ago, I realised we hadn't provided a `PolicyRegistry<TResult>` for policies of type `Policy<TResult>`.  

(_Background_: In Polly, `Policy<TResult>`. does not extend `Policy`, because the original non-generic `Policy` class also has a generic _method_ `.Execute<TResult>(...)`. During discussion around v4.3.0/v5.0, people preferred to keep that overload (for its flexibility; and removing would be a big breaking change). But its existence necessarily means `Policy<TResult>` cannot extend `Policy`. `Policy<TResult>` otoh is essential for strongly-typed `.HandleResult<TResult>(...)`, and strongly-typed composition of `PolicyWrap`s.)

**What we need to do**

To complete the PolicyRegistry offering, I think we need to add a `PolicyRegistry<TResult> : IPolicyRegistry<TKey, TResult>`, for policies of type `Policy<TResult>`.  Could I suggest you:

+ Make a PR of [your branch](https://github.com/ankitbko/Polly/tree/ankitbko/feature/policy-registry) onto [new v5.0.7 branch I created](https://github.com/App-vNext/Polly/tree/v5.0.7registry) rather than master
+ Then either you or I can add the `PolicyRegistry<TResult>` concept there, add doco etc.

I think the need for `IPolicyRegistry<TKey, TResult>` means we should probably simplify [`public interface IPolicyRegistry<TKey, TPolicy> where TPolicy: Polly.Policy`](https://github.com/ankitbko/Polly/blob/ankitbko/feature/policy-registry/src/Polly.Shared/Registry/IPolicyRegistry.cs#L11), replace it simply with `public interface IPolicyRegistry<TKey, Policy>`.  Otherwise, the new `public interface IPolicyRegistry<TKey, TResult>` has a signature-clash with  `IPolicyRegistry<TKey, TPolicy>`.  

Views?

ankitbko:
@reisenberger No need to apologies. Its been great working with you. Thanks for such detailed discussion on all topic.

I did not fully understand what you mentioned here of `PolicyRegistry<TResult>` (maybe due to lack of background on `Policy` vs `Policy<TResult>`). From what I understand, `IPolicyRegistry<Key, Policy>` provides a mechanism of storing (any) policies (anywhere). The default implementation stores it in-memory (which is fine).

Are you looking to create another class `PolicyRegistry<TResult>` which can store `Policy<TResult>`? But then I will have two registry, one for `Policy` and another for `Policy<TResult>` (feels weird). 
But then what happens if I have `Policy<int>` and `Policy<string>`? Do I now have `PolicyRegistry<int>` and `PolicyRegistry<string>` to store individual type of policy? (may be I am missing something here)

Just thinking out loud, what if we have another kind of `Policy` tomorrow, say `Policy<TIn, TOut>` (highly unlikely) so we will create another registry?

Ok so the requirement is that to *register* `Policy<TResult>` also, which is currently not possible due `where Policy: Polly.Policy`  generic type constraint . Let's suppose we remove the constraint. Removing it effectively makes it a generic dictionary `IPolicyRegistry<Key, Value>`, allowing user to store anything.

 Then the next challenge is that concrete implementation also depends upon `Policy`.
```csharp
public class DefaultPolicyRegistry : IPolicyRegistry<string, Policy>
```
How about we create an empty interface `IPolicy`, derive both `Policy` and `Policy<TResult>` from it, and then implement `DefaultPolicyRegistry : IPolicyRegistry<string, IPolicy>` and re-introduce the `where` type constraint with `IPolicy`.
This should in general allow me to use same registry to store both `Policy` and `Policy<TResult>`.

For the user, registry just stores policy(can be anywhere) and gives ability to retrieve them based on `Key`. Doesn't it make more sense that we have one registry which can store all kind of policies, and if I have the `Key` I can get whatever I stored.

Apologies if I misunderstood you and you meant something different (I had a long, tiring day).

PS: I did not create a new PR instead I just switched the target branch in this pull request to v5.0.7 (so the discussions are not split in multiple places). I hope it works for you.

reisenberger:
@ankitbko  Perfect to just re-target the PR to the new branch!

The evolution of Polly has led to `Policy<TResult>` does not extend `Policy`.  Past #129 describes further.  Also, non-generic policies can execute `Action`s returning void, while generic `Policy<TResult>`, can only execute `Func`s returning `TResult`.  So, unlike `IEnumerable<T> : IEnumerable`, we do _not_ have `Policy<TResult> : Policy`.  

I was also uncomfortable about multiple registries: one registry would probably be better, as you say.  Creating an empty marker interface, as you suggest, could be a way to do this.  I can think of one other approach.  I will aim to spike out - or describe in more detail - shortly.

reisenberger:
Huge thanks @ankitbko for all the great thinking towards `PolicyRegistry`!  

Empty marker interface definitely a possible way to bridge between `Policy<TResult>` and `Policy` which are (necessarily) not in an inheritance hierarchy.  Two thoughts followed from the empty interface:

(a) if the registry returns instances fulfilling empty `IsPolicy` (say), users would have to cast the received policy every time to the relevant `Policy` or `Policy<TResult>`.  To avoid that, the Registry c/should provide helper getter overloads to reduce the burden, maybe something like:

`public Policy          IPolicyRegistry.Policy(TKey) // or Get(TKey)`
`public Policy<TResult> IPolicyRegistry.PolicyFor<TResult>(TKey) // or Get<TResult>(TKey)`

(b) given an empty interface, users might soon ask why it is empty, and ask for populated interfaces that in fact provide the relevant `.Execute(...)` overloads they can use.  

Which led me to consider whether:

(i) we can/could/should now introduce interfaces, say, `IPolicy`, `IPolicy<TResult>`,  `IPolicyAsync`, `IPolicyAsync<TResult>`, defining the relevant available `.Execute/AndCapture/Async(...)` overloads.  Alongside the possible empty marker interface `IsPolicy`.

(ii) whether either the `Policy<TResult>`/`Policy` split, or the sync/async split, can be reduced/mitigated.  

Re (ii), I think not - intend to post later on this - but now is definitely the time to consider, before we introduce interfaces.  If we go for interfaces now, the helper methods at (a) above also would return those rather than the concrete `Policy<TResult>`/`Policy` .

Thoughts on any aspect of this?

ankitbko:
@reisenberger Very inline with my thoughts. Next steps could very well be populating the relevant interfaces.
We have a method `TryGetValue` in the registry. How about having a `TryGetValue<TResult>` extension method as the helper method? It would typecast and return the `Policy`.

reisenberger:
Hi @ankitbko !  Here [reisenberger/Polly/policy-registry-TResult-nointerfaces][https://github.com/reisenberger/Polly/commits/policy-registry-TResult-nointerfaces) is a prototype branch providing a single policy registry class offering overloads for both `Policy` and `Policy<TResult>`, without using interfaces.  

`this[...]` operators have been replaced by `Get`/`Set` methods.   `Add`, `Set`, `Get`, and `TryGetValue` exist in overloads for both generic and non-generic policies.

Comments welcome!

We could go this way with `PolicyRegistry` _if we don't introduce interfaces_.  (Not necessarily preferring this: just the first prototype which is ready.)  Working separately towards a branch of Polly including a full set of interfaces  (empty) `IsPolicy` and (populated) `IPolicy`, `IPolicyAsync`, `IPolicy<TResult>`, `IPolicyAsync<TResult>`.

reisenberger:
Hi @ankitbko !  Here is a [Registry based on yours, but allowing us to pull policies out of the Registry correctly typed, based on generic methods](https://github.com/reisenberger/Polly/tree/feature/interfacesplusregistry).  The generic methods become particularly useful because of [adding interfaces](https://github.com/reisenberger/Polly/tree/feature/interfaces).  

This version again omits the `this[...]` operators, because the getter on that, `registry[key]`, could never return policies of an immediately usable type: users would always have to cast first.  

Please let me know what you think!  (particularly between the two options of this comment, and the previous one)

ankitbko:
@reisenberger Sorry for delay. This is excellent. Interfaces looks amazing. Couple of thoughts on the Interface implementation of registry - 

- There are two ways to add to registry `Add<TPolicy>(string key, TPolicy policy) where TPolicy : IsPolicy` and `Set<TPolicy>(string key, TPolicy policy) where TPolicy : IsPolicy`. Any differences between two or can we remove `Set` and just keep `Add` (or vice versa)?

- If using Interfaces, is there any need of making `Add` method generic? Alternative would be to just have `Add(string key, IsPolicy policy)`.

- Since now we have interface, does it make sense to have dictionary value of type `object` in `IDictionary<string, object>` or should we make it `IsPolicy` like `IDictionary<string, IsPolicy>`?

- `WithPolicyKey(String policyKey)` method in `ISyncPolicy` has returns type of `Policy` ([Line 22](https://github.com/reisenberger/Polly/blob/feature/interfacesplusregistry/src/Polly.Shared/ISyncPolicy.cs#L22)). Any particular reason of having reference of derived class in interface? (Similar for other sibling interfaces)

I am leaning more towards interfaces side. It leaves registry class clean (rather than having different get set for Policy and Policy<T>). What do you think?

As for `this[key]`, I don't see any benefit of having it (or not having it) as long as it is well documented.
 
But if we change `IDictionary` to store type `IsPolicy`, it *may* be useful if we add methods to `IsPolicy` rather than having it empty. Do you think it is possible to move common methods from `ISyncPolicy` and `IAsyncSyncPolicy` (and their generic counterpart) to `IsPolicy`?  
If we are able to move common methods to `IsPolicy` (say Execute), people can use `this[key]` without having to explicitly typecast it. Does it make sense?

reisenberger:
@ankitbko Huge thanks for your thoughtful feedback on the interface-based approach I overlaid on your PR!

Pushed changes in response to your comments, again to this branch: https://github.com/reisenberger/Polly/tree/feature/interfacesplusregistry

Responses to your comments:

>Since now we have interface, does it make sense to have dictionary value of type object in IDictionary<string, object> or should we make it IsPolicy like IDictionary<string, IsPolicy>?

Agree.  Done!

>WithPolicyKey(String policyKey) method in ISyncPolicy has returns type of Policy (Line 22). Any particular reason of having reference of derived class in interface? (Similar for other sibling interfaces)

Great catch!  Done!

>There are two ways to add to registry Add<TPolicy>(string key, TPolicy policy) where TPolicy : IsPolicy and Set<TPolicy>(string key, TPolicy policy) where TPolicy : IsPolicy. Any differences between two or can we remove Set and just keep Add (or vice versa)?

The intention was that `Set()` was like `this[]`, so `Add()` and `Set()` differed in that `Add()` would throw `ArgumentException` on duplicate key, while `Set()` would not.  The very fact that you asked the q though, shows that [sticking with API precedent](https://twitter.com/jhalt/status/851831507727433728) probably would engender less confusion.  I've therefore switched `Set()` back to `this[]`.

The original reason for not using `this[]` was the thought that users might find it frustrating that `registry[key]` necessarily [returns only an `IsPolicy`](https://github.com/reisenberger/Polly/blob/cf76b3fce333ad5dceddfd58fcd106ad4d179f15/src/Polly.Shared/Registry/PolicyRegistry.cs#L41), which would require the further cast (as you observed), in order to be useful.  I have addressed that now by giving explicit guidance [right in the intellisense](https://github.com/reisenberger/Polly/blob/cf76b3fce333ad5dceddfd58fcd106ad4d179f15/src/Polly.Shared/Registry/PolicyRegistry.cs#L35), highlighting the `Get<TPolicy>` alternative. And can ditto in online doco.

>If using Interfaces, is there any need of making Add method generic? Alternative would be to just have Add(string key, IsPolicy policy).

Agree could reduce it to `IsPolicy`.  Kept it generic just for symmetry with `Get<>()`.  Encourages users into thinking about which policy type they are dealing with, for when they later `Get<>()` it back out of registry.

>I am leaning more towards interfaces side. It leaves registry class clean (rather than having different get set for Policy and Policy). What do you think?

:+1:

>But if we change IDictionary to store type IsPolicy, it may be useful if we add methods to IsPolicy rather than having it empty. Do you think it is possible to move common methods from ISyncPolicy and IAsyncSyncPolicy (and their generic counterpart) to IsPolicy?

Completely get the suggestion.  I'm (starting out) genuinely ambivalent, per following discussion:

(a) The history is that the sync/async split in Polly was, long pre-AppvNext, implemented (unfortunately imo) as a [runtime exception on the Policy class](https://github.com/App-vNext/Polly/blob/9d31709d1d02bbf7342f19574bee69a7afe215e1/src/Polly.Shared/PolicyAsync.cs#L31) (instances of `Policy` are usable only for _either_ sync or async, but enforced only at runtime rather than compile time), instead of defining separate `Policy` and `PolicyAsync` classes at that point.  That historical decision means that users get both sync `.Execute(...)` and async `.ExecuteAsync(...)` overloads available (eg in intellisense) on every concrete `Policy` instance, even though only half the overloads are valid (on any one instance), and the other half will throw.  

My introducing `SyncPolicy` and `AsyncPolicy` as interfaces then is an attempt to move users away from this - to separate out the sync and async functionality currently joint on `Policy`, by that well-known [interface segregation principle](https://en.wikipedia.org/wiki/Interface_segregation_principle).  If users get into the habit of using `ISyncPolicy`/`AsyncPolicy` , they will get only the set of execution overloads that apply :grin: 

As `SyncPolicy` and `AsyncPolicy` are interface segregation, there are no _common_ overloads to move on to `IsPolicy`.  We essentially have two options:

(b-i) Keep `IsPolicy` empty, a marker-interface only (and encourage users towards `SyncPolicy` and `AsyncPolicy`)
(b-ii) Make `IsPolicy` a union of `SyncPolicy` and `AsyncPolicy`
(and similar for generic versions)

Having started out ambivalent, I'm leaning to (b-i).  Doing (b-ii) just places us back to the "false promise" that a single Policy instance can be used for both sync or async, when in fact it can only be used for one or other.  (b-i) at least begins to push the solution architecture in the right direction.  

For further background, [this recent blog post](http://www.thepollyproject.org/2017/06/09/polly-and-synchronous-versus-asynchronous-policies/) sets out why I think we're stuck (without a major change of syntax) with separate Policy instances for sync and async executions, in Polly.

Sorry for that long explanation.  The sync/async split in Polly is an unhappy legacy part of the Polly architecture I've been tussling with, hence the long explanation - and interested in people's views.

:sweat_smile:

reisenberger:
Hi @ankitbko  Per discussion on slack, I have:

+ drafted some wiki documentation ( [PolicyRegistry](https://github.com/App-vNext/Polly/wiki/PolicyRegistry); [Polly and interfaces](https://github.com/App-vNext/Polly/wiki/Polly-and-interfaces) ).
+ also [readme doco](https://github.com/reisenberger/Polly/blob/b043a996af6bdab5b8712552e38a9dab499904ca/README.md#policyregistry-v52) in [my working branch on this feature](https://github.com/reisenberger/Polly/commits/feature/interfacesplusregistry).

Anything you think we should add/change?

I'll aim to sort out the PRs between code branches in next few days

reisenberger:
@ankitbko Merged this to the App-vNext:v5.0.7registry branch.  Decided it would be easier to merge there, then add my interfaces on top.

Thank you for your great contribution to this feature!  :100:

Issue #177, opened 2016-11-11
Calculating sleep duration for WaitAndRetry, based on the execution result; eg 429 headers


jmansar:
Hi, 

I have a small enhancement idea.

Currently the `sleepDurationProvider` function in the `RetryWithSleep` policy doesn't have access to the result of the executed action. The only value passed to the `sleepDurationProvider` is the retry count.

I have a scenario where it would be quite useful to know the result in the wait time calculation function.

#### Scenario
There is an API, which I call in my application. It returns the error response with the `Retry-After` HTTP header set, when client exceeds the request rate limit or it is temporarily unavailable. 
The header contains number of seconds the client should wait before repeating the request.
It should be possible to use that value as a wait time.

Could you extend the sleepDurationProvider function to include the result from the executed action?

reisenberger:
@jmansar Thanks for bringing this really interesting idea across from our [slack channel](http://www.pollytalk.org/)!

Main question in my mind: whether exposing result directly to `sleepDurationProvider` is the way we’d want to support this.  Without knowing the underlying use case, some users might wonder: why is the execution result exposed to _that_ policy-delegate? (and _only_ that one?).  And the pattern might not immediately make sense to extend to other policy-operation-influencing delegates.

So: how else can we do this, that keeps the Polly architecture evolving in a coherent way?

In this case I'm thinking we’d probably want to support this by expanding the role of the `Context` class in executions (something we intend anyway, when we move to [support metrics](https://github.com/App-vNext/Polly/wiki/Polly-Roadmap#emit-events-andor-statistics)).  To fulfil what you need ... we would add the execution result instead to `Context`; and make `sleepDurationProvider` take `Context` as input param instead. 

Making `Context` the central way we channel extra info to all parts of policy execution:

+ Follows a common pattern, with all the `onStateChange` delegates already taking `Context`
+ Reduces the overall number of overloads for different variants (already too proliferated)
+ Allows all the `onStateChange` delegates to also benefit from getting the execution result :smile: 

---

Note there are other similar (_in a very broad sense_) issues like #38 .  Also about exchanging data between execution and  Policy-operation-influencing delegates.

My suggestion-in-the-meantime for your use case is similar to at bottom of #38.  Close over some common variable in the `sleepDurationProvider` and `Execute` delegates, to share that part of the response back to the `sleepDurationProvider` delegate

johnmcase:
+1

[...]

Issue #73, opened 2016-01-07
Add Attempt count to PolicyResult [general: expand info on Policy execution available in PolicyResult]


TrevorPilley:
We are using `ExecuteAndCapture` in our application, the policy is defined in configuration but executed elsewhere. I'd like to log whether a retry has happened (and how many) but all I have is `PolicyResult` to work from.

In one part of the app, we define a function which builds a policy:

```
config.BuildExceptionPolicy(() => Policy.Handle<IOxception>().RetryAsync(2));
```

And execute it in another:

```
var policy = step.BuildExceptionPolicy();
return policy.ExecuteAndCaptureAsync(async () => { ... } );
```

If the outcome is Failure, I log the final exception but along side that I'd like to log the number of attempts that were taken.

reisenberger:
Hi @TrevorPilley.  

Because of the way the Polly execute methods are generalised, capturing the number of attempts into the `PolicyResult` is not that straightforward.  `.ExecuteAndCaptureAsync(...)` is agnostic to (and has no knowledge of) whether what is being run underneath is a retry policy, circuit-breaker policy without retries (or indeed some new kind of action-transforming policy in the future).

An approach in the meantime could be something like :

```
int attempts = 0;
var policy = step.BuildExceptionPolicy();
var result = policy.ExecuteAndCaptureAsync(async () => { 
   attempts++;
   //  ... 
} );
```


@joelhulen : I suggest we leave this issue open for a while so that others can comment if they would also like additional execution metadata in`PolicyResult`.

TrevorPilley:
Could you not just add `public int Attempts {get;set;}` to `PolicyResult`  and do `Result.Attempts++;` to the code which executes the delegate passed to `ExecuteAndCaptureAsync`?

reisenberger:
@TrevorPilley.  I looked at this in more depth in the intervening week, and agree that this would not be as difficult to integrate into Polly as initially feared: the `attempts++` approach could be pushed down to the core Polly codebase.  
 
We have also to consider how this would play in the context of both Retry and CircuitBreaker policies, but I think this would be ok. For CircuitBreaker, the `PolicyResult.Attempts` (or similar) property would just indicate 1 if the call was attempted, and 0 if it was not (due to the circuit being broken).
 
Discussing this one further with the core App vNext team, and hope to come back on this in the coming days.

reisenberger:
Just updating thinking on this issue:
 
[...]

Issue #38, opened 2015-06-23
Passing exception and retry count to execution (general: pass mutable data between onRetry and Execute)


YoniH:
Hi,

In policy execution, I need to access the retry number and the exception that was thrown.

I managed to do this only with class members, as in this console app:

```
class Program
{
    private static Exception mException;
    private static int mRetryCount;

    static void Main(string[] args)
    {
        var policy = Policy.Handle<DivideByZeroException>().Retry((exception, retryCount, context) =>
        {
            mException = exception;
            mRetryCount = retryCount;
        });

        policy.Execute(DoSomething);
    }

    private static void DoSomething()
    {
        Console.WriteLine("Try number {0}", mRetryCount);
        Console.WriteLine("Exception was {0}", mException);
        Console.WriteLine();

        int x = 0;
        var y = 2/x;
    }
}
```

However, is there a way to do this only with Polly, without class members?

YoniH:
After some more experimenting with the library, I want to try to be more accurate. What I want to do is this:
1. On every retry, I want to calculate some numeric value based on the exception and maybe the retry count.
2. This number should be passed as argument to the execution function.

Currently the way I'm doing it is something like this:

    class Program
    {
        static void Main(string[] args)
        {
            var someNumber = new Integer();
            
            var policy = Policy.Handle<DivideByZeroException>().Retry(3, (exception, retryCount, context) =>
            {
                (context["someNumber"] as Integer).Value = 17; //Instead of 17, the value would be calculated based on the exception
            });
            
            policy.Execute(() => DoSomething(someNumber), new Dictionary<string, object>{{"someNumber", someNumber}});
        }

        private static void DoSomething(Integer numberArgument)
        {
            Console.WriteLine("The number is {0}", numberArgument.Value);
            Console.WriteLine();

            int x = 0;
            var y = 2/x;
        }

        class Integer
        {
            public int Value { get; set; }
        }
    }



This seems a bit cumbersome. Is there a better way to do that?

hawkunsh:
Any news on this?

reisenberger:
@YoniH @hawkunsh An AppvNext team has taken over stewardship of Polly, and are now looking at historic issues such as this.  I'll comment in case still useful to you or others.

[...]

PR #244, opened 2017-05-02
v5.1.0: Allow different parts of an execution to exchange state through `Context` flowing with execution

reisenberger:
Polly v5.1.0: Allow different parts of an execution to exchange state through `Context` flowing with execution

+ Makes the `Context` flowing with each execution mutable (safe: `Context` instances are independent per execution, and progress through an individual execution is sequential)
+ Allow `.Execute(...)` (and similar) overloads to take `Context` as an input parameter
+ Allow all Policy-behaviour delegates (eg `onRetry', `onBreak`, `sleepDurationProvider` etc) to take `Context` as an input parameter
+ Rationalise/tidy code layout and tests in various places

Permits #177, also addresses #73 and #38


reisenberger:
( @joelhulen Although there are a large number of commits, suggest merge-without-squashing.  Commits contain code rearrangement [moving/renaming files]; GitHub doco suggests Git can track history when these are separate commits - not if squashed )

Issue #270, opened 2017-07-11
Sync TimeoutPolicy in TimeoutStrategy.Pessimistic unwantedly wraps exceptions in `AggregateException`


reisenberger:
A premise of Polly has been that executing a delegate through a Polly policy shouldn't change the way that exceptions raised by that delegate are expressed.

[`TimeoutPolicy` in `TimeoutStrategy.Pessimistic`](https://github.com/App-vNext/Polly/wiki/Timeout#pessimistic-timeout) is violating that principle (a bug), because it's causing exceptions that would normally be raised directly by an executed delegate, to get wrapped in `AggregateException`.

reisenberger:
Fix for this is easy: replace [`.Wait()`](https://github.com/App-vNext/Polly/blob/75456beee1c4e25b88ab34057434bdc1eec30480/src/Polly.Shared/Timeout/TimeoutEngine.cs#L51) with `.GetAwaiter().GetResult()`

Needs some tests.

reisenberger:
In progress by @hambudi as #273

PR #273, opened 2017-07-17
Fix aggregate exception thrown from pessimistic timeout policy

hambudi:
If pessimistic timeout is selected for timeout policy, the policy will execute the user delegate as a Task on a ThreadPool thread. This causes any exceptions thrown from within the policy to be wrapped in an aggregate exception.

This breaks polly's contract that exceptions thrown from the delegate are not modified by polly.

Fix: Catch aggregate exception and thrown the single inner exception if exists.
Tests: Added tests to ensure that Normal and AggregateExceptions thrown from within the delegate are recived as is.

Potential issues: While working on the fix I saw some articles online saying that exceptions caught and then thrown using the "throw ex" cause the stacktrace to be reset to the new throw statement.


src/Polly.Shared/Timeout/TimeoutEngine.cs
```diff
@@ -47,8 +47,14 @@ internal static partial class TimeoutEngine
...
+                        try
+                        {
+                            actionTask.Wait(timeoutCancellationTokenSource.Token); // cancellation token here cancels the Wait() and causes it to throw, but does not cancel actionTask.  We use only timeoutCancellationTokenSource.Token here, not combinedToken.  If we allowed the user's cancellation token to cancel the Wait(), in this pessimistic scenario where the user delegate may not observe that cancellation, that would create a no-longer-observed task.  That task could in turn later fault before completing, risking an UnobservedTaskException.
+                        }
+                        catch (AggregateException ex) when (ex.InnerExceptions.Count == 1)
```
udlose: And what do you think should happen when there happen to be multiple inner exceptions?  Something like ex.Flatten() should probably be used here.

src/Polly.Shared/Timeout/TimeoutEngine.cs
```diff
@@ -47,8 +47,14 @@ internal static partial class TimeoutEngine
...
+                        {
+                            throw ex.InnerException;
```
udlose: Have you tested this to see if the stack trace is reset?

src/Polly.Shared/Timeout/TimeoutEngine.cs
hambudi: When an exception is thrown inside the task, that exception is wraped in an AggregateException when we use Task.Wait().

In this scenario the AggregateException will always have one inner exception. The inner exception itself can always be an AggregateException with multiple inner exceptions.

This will happen if delegate itself throws an AggregateException with many inner exceptions and I dont want to mess with that in this change. The intention is only to preserve the exception behaviour of the delegate and remove the inconsistency with running it inside a task.

This is also the reason that I am not use .Flatten(), as that would change the exception thrown by the delegate.

src/Polly.Shared/Timeout/TimeoutEngine.cs
hambudi: I have not, based on my reading I understand that the stacktrace will be reset every time we use "throw ex" command and inorder to not reset the trace we have to use throw. 

I dont think using throw is possible here.

I will test it out and see if the stacktrace is being reset and reply back here.

reisenberger: Exactly: `ExceptionDispatchInfo.Capture().Throw()`.

The wrinkle is that .NET4.0 (one of our compile targets) does not support it.  We can consider solutions such as offered [here](https://stackoverflow.com/questions/20171877/is-there-an-analog-of-exceptiondispatchinfo-in-microsoft-bcl-async); whether we exclude .NET4.0 from this, ie do not fix #270 for .NET 4.0.

src/Polly.Shared/Timeout/TimeoutEngine.cs
reisenberger: Thanks @udlose for the observations!  Re:

>And what do you think should happen when there happen to be multiple inner exceptions? 

A general `Task` could, as you say @udlose , throw an `AggregateException` with either single or multiple `InnerExceptions`.  In this case however we know [the specific way this `Task` is invoked](https://github.com/hambudi/Polly/blob/fixAggregateExceptionsInPessimisticsTimeout/src/Polly.Shared/Timeout/TimeoutEngine.cs#L47-L48).  In all tests empirically, this is throwing an `AggregateException` with a single `InnerException`.  Even if the `myDelegate` executed as `TimeoutPolicy.Execute(() => myDelegate)` throws multiple inner exceptions, the result is an: 
`AggregateException -> AggregateException -> (multiple inner exceptions)`, and we can (as the PR does) strip off the outer `AggregateException`, so that the exception the policy rethrows matches the exception that would be thrown if the delegate was not executed through pessimistic `TimeoutPolicy`  - this is the goal.

Given our confidence that all `AggregateException`s thrown here will be an `AggregateException` with single `InnerException`, we could plausibly drop the [`ex.InnerExceptions.Count == 1` check](https://github.com/hambudi/Polly/blob/fixAggregateExceptionsInPessimisticsTimeout/src/Polly.Shared/Timeout/TimeoutEngine.cs#L54).  However, I suggest we retain it, as it prevents Polly ever _invisibly throwing away_ information, should the delegate throw an aggregate-with-mutliple-inners exception.

Re:

>Something like ex.Flatten() should probably be used here.

Couple of reasons why we have chosen not to `ex.Flatten()` within Polly.  First, a principle for Polly policies is that they should not cause any side effects to the way the user's delegate is executed, or expresses exceptions, beyond the stated aim of the policy.  If we choose to `ex.Flatten()` within the library, that causes a side effect as it _changes_ the way exceptions from the user's delegate are expressed.  Second, if we choose to `ex.Flatten()` from within the library, we _take away from the user the choice_ of whether to `ex.Flatten()`.

src/Polly.SharedSpecs/Timeout/TimeoutSpecs.cs
```diff
@@ -139,6 +139,24 @@ public void Should_throw_when_timeout_is_less_than_execution_duration__pessimist
...
+        [Fact]
+        public void Should_rethrow_aggregate_exception_from_inside_delegate__pessimistic()
+        {
+            var policy = Policy.Timeout(TimeSpan.FromMilliseconds(50), TimeoutStrategy.Pessimistic);
+            var msg = "Aggregate Exception thrown from the delegate";
+            // Check to see if nested aggregate exceptions are unwrapped correctly
+            AggregateException exception = new AggregateException(msg, new NotImplementedException());
+            policy.Invoking(p => p.Execute(() => { throw exception; })).ShouldThrow<AggregateException>().WithMessage(msg).WithInnerException<NotImplementedException>();
+        }
+
```
reisenberger: @hambudi Your spec here provides perfect coverage to demonstrate that an aggregate exception thrown within the Policy is correctly unwrapped.  

I built on yours to add couple more specs which demonstrate the cases when the `AggregateException` thrown within the user delegate contains multiple `InnerExceptions`.  Would you be happy to add these (if you agree they are useful) to your PR?

```c#

        [Fact]
        public void Should_rethrow_aggregate_exception_with_multiple_exceptions_from_inside_delegate__pessimistic()
        {
            var policy = Policy.Timeout(TimeSpan.FromMilliseconds(50), TimeoutStrategy.Pessimistic);
            var msg = "Aggregate Exception thrown from the delegate";

            Exception innerException1 = new NotImplementedException();
            Exception innerException2 = new DivideByZeroException();
            AggregateException aggregateException = new AggregateException(msg, innerException1, innerException2);
            Action action = () => { throw aggregateException; };

            // Whether executing the delegate directly, or through the policy, exception behaviour should be the same.
            action.ShouldThrow<AggregateException>()
                .WithMessage(msg)
                .And.InnerExceptions.Should().BeEquivalentTo<Exception>(new[] {innerException1, innerException2});

            policy.Invoking(p => p.Execute(action)).ShouldThrow<AggregateException>()
                .WithMessage(msg)
                .And.InnerExceptions.Should().BeEquivalentTo<Exception>(new[] {innerException1, innerException2});
        }

        [Fact]
        public void Should_rethrow_aggregate_exception_with_example_cause_of_multiple_exceptions_from_inside_delegate__pessimistic()
        {
            var policy = Policy.Timeout(TimeSpan.FromMilliseconds(50), TimeoutStrategy.Pessimistic);

            Exception innerException1 = new NotImplementedException();
            Exception innerException2 = new DivideByZeroException();
            Action action = () =>
            {
                Task task1 = Task.Run(() => { throw innerException1; });
                Task task2 = Task.Run(() => { throw innerException2; });
                Task.WhenAll(task1, task2).Wait();
            };

            // Whether executing the delegate directly, or through the policy, exception behaviour should be the same.
            action.ShouldThrow<AggregateException>()
                .And.InnerExceptions.Should().BeEquivalentTo<Exception>(new[] {innerException1, innerException2});

            policy.Invoking(p => p.Execute(action)).ShouldThrow<AggregateException>()
                .And.InnerExceptions.Should().BeEquivalentTo<Exception>(new[] {innerException1, innerException2});
        }

        [Fact]
        public void Should_rethrow_aggregate_exception_with_another_example_cause_of_multiple_exceptions_from_inside_delegate__pessimistic()
        {
            var policy = Policy.Timeout(TimeSpan.FromMilliseconds(50), TimeoutStrategy.Pessimistic);

            Exception innerException1 = new NotImplementedException();
            Exception innerException2 = new DivideByZeroException();
            Action action = () =>
            {
                Action action1 = () => { throw innerException1; };
                Action action2 = () => { throw innerException2; };
                Parallel.Invoke(action1, action2);
            };

            // Whether executing the delegate directly, or through the policy, exception behaviour should be the same.
            action.ShouldThrow<AggregateException>()
                .And.InnerExceptions.Should().BeEquivalentTo<Exception>(new[] { innerException1, innerException2 });

            policy.Invoking(p => p.Execute(action)).ShouldThrow<AggregateException>()
                .And.InnerExceptions.Should().BeEquivalentTo<Exception>(new[] { innerException1, innerException2 });
        }
```

@udlose These specs demo what I mentioned in previous comment: if the delegate executed through `TimeoutPolicy` in `TimeoutStrategy.Pessimistic` throws an `AggregateException -> (multiple inner exceptions)`, that get's expressed _within_ Polly (due to the way this strategy has to execute the delegate) as `AggregateException -> AggregateException -> (multiple inner exceptions)`, and @hambudi 's PR correctly strips off the outer `AggregateException`.

hambudi: I have added these.

src/Polly.Shared/Timeout/TimeoutEngine.cs
hambudi: I am currently working on adding tests for verifying the stacktrace. I was wondring if you want me to include #280  in this PR or would you like to check that in and then have me rebase?

reisenberger:
@hambudi Thank you for all your work on this PR!  Would you be able to [follow these instructions](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request) to switch the base branch of the PR, to this new branch? : https://github.com/App-vNext/Polly/tree/fixAggregateExceptionsInPessimisticsTimeout

Then I can tackle the timeout issues on the tests (it's an issue with running timing-sensitive tests in AppVeyor), and add similar tests for the generic `TResult` variant of `TimeoutPolicy`, etc.

Thanks for your awesome contribution!

hambudi:
I changed the base branch.

PR #282, opened 2017-07-26
Add an interface providing read-only access to a policy registry.

SeanFarrow:
Splits the PolicyRegistry to reaterfaces.nd-only and writable i


reisenberger:
Closing in favour of #283 at request of original contributor

PR #296, opened 2017-08-18
Clarify async Fallback tests

reisenberger:
Clarify async Fallback tests. Async elision meant the tests were correct, but the new versions provide greater clarity.

### DIFF
--- a/src/Polly.SharedSpecs/Fallback/FallbackAsyncSpecs.cs
+++ b/src/Polly.SharedSpecs/Fallback/FallbackAsyncSpecs.cs
@@ -90,7 +90,7 @@ namespace Polly.Specs.Fallback
         #region Policy operation tests
 
         [Fact]
-        public void Should_not_execute_fallback_when_execute_delegate_does_not_throw()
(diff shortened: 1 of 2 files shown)


reisenberger:
Thanks @udlose for being extra pair of :eyes: .  xunit 2.0+ (which we're using) correctly handles the `async void` tests (by using a custom `SynchronisationContext` as I understand) (Stephen Cleary did an msdn mag article mentioning it?), but yeah, let's switch them to `async Task` just to avoid any ambiguity or potential future false-positives (eg if we switched test-runners in future).

Thanks again for the good spot!

Fixing in #300

udlose:
Np @reisenberger.  Thanks for making the fix!

reisenberger:
>For reference, here's a link to the MSDN article - https://msdn.microsoft.com/en-us/magazine/dn818493.aspx
Thanks @udlose !  

For future ref also, for anyone returning to this thread, here the [xunit 2.0 release notes](https://xunit.github.io/releases/2.0) and [related issue](https://github.com/xunit/resharper-xunit/issues/8) which verify the xunit 2.0 behaviour.

Issue #298, opened 2017-08-18
Make the constructors of the Policy abstract base classes protected, not internal [allow custom policies]


reisenberger:
The constructors of the `Policy` and `Policy<TResult>` abstract base classes are `internal` but should be made ~public~`protected`.  

This would allow users to implement new custom policies, outside of the Polly package, deriving from these base classes.

Needed in files `Policy.cs`, `Policy.TResult.cs`, `PolicyAsync.cs`, `PolicyAsync.TResult.cs`.

ExtRemo75:
Please see pull request #299. Note that I have made more changes than this Issue suggested:
- The constructor access modifiers for Policy have been changed to protected (public is not needed)
- Some properties of Policy have been made public as they are commonly used to instantiate policies
- Almost all properties/methods of Policy have been made virtual, so they can be overridden in a derived Policy
- Some external properties (PolicyBuilder) and delegates have been made public as they are commonly used to instantiate policies

I've added comments for the newly-public constructors/methods/properties.

udlose:
This at least provides the option to consumers of Polly to do so. However,
I think it is an anti-pattern to extend an external API and expose its
implementation throughout your code - even when doing so thru interfaces.
Rather, it should be abstracted away and wrapped so the 3rd party API could
be swapped at some future point.

ExtRemo75:
@udlose Agreed, but this is how it is currently setup. 

Imho, I believe if you are architecting your own solution with the ability to swap the 3rd party API (Polly) out at any point in time, which is a great idea obviously, then you'd probably have to create an abstraction for the concept of 'policies' for your own application, e.g. a policy provider with a specific implementation that uses, and optionally implements additional, Polly policies.

My 2 cents.

udlose:
@ExtRemo75 - yes, you will have to somehow abstract usages of policies somehow.  I was thinking of doing so via aggregation instead of inheritence. Though, I guess it's a matter of preference. Based on my experience, i've had more success using aggregation than inheritance to isolate 3rd party APIs.

reisenberger:
Thanks all for all the great comments.  Hoping to come back to this and #299 before the end of the week.

reisenberger:
@udlose @ExtRemo75  Great discussion re aggregation rather than inheritance for isolating 3rd party APIs.  I guess it depends what you are trying to achieve.  If you want to isolate Polly, then yes, wrapping in your own abstractions is the way to go.  

[...]

PR #299, opened 2017-08-19
Change the access modifiers of Policy constructors, methods and relat…

rjongeneelen:
…ed properties. Make Policy properties/methods virtual so they can be overridden.

See issue https://github.com/App-vNext/Polly/issues/298

### DIFF
--- a/src/Polly.Shared/ExceptionPredicate.cs
+++ b/src/Polly.Shared/ExceptionPredicate.cs
@@ -2,5 +2,10 @@
 
 namespace Polly
 {
-    internal delegate bool ExceptionPredicate(Exception ex); 
+    /// <summary>
+    /// Predicate for validating the exception type and optional exception predicate
+    /// </summary>
+    /// <param name="ex">The exception to validate</param>
+    /// <returns>True if valid</returns>
+    public delegate bool ExceptionPredicate(Exception ex); 
 }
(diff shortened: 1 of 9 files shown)


reisenberger:
@extremo75 , Apologies for the delay responding. _Huge_ thank-you for your work on this PR towards allowing custom policies!

The PR goes further than the original proposal (thanks for  [your note](https://github.com/App-vNext/Polly/issues/298#issuecomment-323504656) to this effect!).  My slight concern is it (at mo) creates a very broad extension surface to Polly, which could cause problems down the line for both Polly and users extending.  It may be better for both if we can achieve the same extensibility but keep extension points more tightly focused?

The concern over a  broad extension surface is the large number of potential points of contact between Polly and custom-policies which it creates - a join with a large number of connection points is potentially more rigid/ inflexible/ brittle, as Polly evolves, than if we can create one with only a small number.  With many extension points, if we change only a single one, it risks a breakage for users with custom policies.  Equally, if we set a precedent that almost every method on `Policy` is `public`/`protected virtual` so that people could override it, users who've created custom policies this way risk having frequently to 'play catch up' (override more methods), as we evolve and extend.  (Items like 299 will cause significant changes.)

A single extension point (or tightly focused group) should otoh flex more easily with Polly as it grows.

I hope this makes sense.  What do you think?

---

So, practically, the main concern was making the `.Execute(...)` overloads override-able.  To be sure I'm not missing anything: did you envisage particular use cases for overriding `.Execute(...)` etc, which could not be solved other ways?  To permit custom policies, I had thought it would be enough just to allow a custom implementation (passed into the base `Policy` constructor at the moment), and let the  `.Execute(...)` overloads do their normal job?  

Thanks again for all your contribution to this!

reisenberger:
Closed in favour of the simpler equivalent, #391

PR #337, opened 2017-10-25
Adding netstandard2.0 support

jelical:
Also fixing SystemClock to be request-safe using AsyncLocal


jelical:
Fixed build failure on warnings

PR #415, opened 2018-03-09
Reference 429 RetryAfter examples in ReadMe

reisenberger:
+ Reference the wiki discussion on handling 429 RetryAfter examples, from the readme.
+ Link XamarinHelp blog referencing Polly

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,8 @@ Policy
   );
 ```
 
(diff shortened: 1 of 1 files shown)


reisenberger:
Merging this as it's only a readme change. The build failure is unrelated, due to cake.core breaking changes against cake plugins - see #416

Issue #416, opened 2018-03-09
Cake build is failing due to Cake.FileHelpers -> Cake.Core version requirement out of sync


reisenberger:
Polly Cake build has started failing due to [Cake.FileHelpers not yet updated to match latest Cake.Core version](https://github.com/cake-contrib/Cake.FileHelpers/issues/19).

To prevent transient cake build failures, we should [pin our version of cake](https://cakebuild.net/docs/tutorials/pinning-cake-version) per [widespread](https://github.com/cake-contrib/Cake.FileHelpers/issues/19#issuecomment-369266695) [recommendations](https://help.appveyor.com/discussions/problems/12599-appveyor-builds-have-stopped-working-due-to-cakecore-version).  Or make a PR on Cake.FileHelpers to update, for them.

SeanFarrow:
@reisenberger , Should we upgrade Cake to v2.0.0? This would fix this issue, now that Cake.FileHelpers has been updated.

I've got to do this for the JSon text serialization caching serializer, so am happy to work through other repos if this would help?

martincostello:
We'd be happy to accept a PR to update to the latest version of Cake.

PR #417, opened 2018-03-09
Pin our version of cake.core

reisenberger:
Pin our version of cake.core to prevent transient build breakages, when cake.core changes cause breaking changes for cake addins.


reisenberger:
Fixes #416

Issue #419, opened 2018-03-20
Conditional Timeout using timeoutProvider (no timeout, via infinite timeout)


Danation:
Right now, using a TimeoutPolicy assumes that there should always be a timeout.  But when specifying a timeoutProvider function, it would be nice to be able to choose not to timeout at all in certain circumstances.

A simple workaround is to specify a very long timeout (many days), but it would be better to have a more clear option.  Perhaps by making it nullable or marking a particular constant as "no timeout" such as TimeSpan.Zero.

What do you think?

Danation:
Hunting through the code, it looks like it uses `CancellationTokenSource.CancelAfter()`, which can take a `-1`.  I wasn't able to dig deep enough into the source to find how it handles -1, but after some testing, it looks like it does what I'd hope (no cancellation). So it looks like returning `TimeSpan.FromMilliseconds(-1)` in a `timeoutProvider` works.  Could anybody confirm that I'm correct on that?

reisenberger:
Hi @danation.  The `Policy.Timeout(...)` configuration overloads will currently throw `ArgumentOutOfRangeException` on constant values `<= TimeSpan.Zero`, but you can use the `Func<TimeSpan> timeoutProvider` to return these values.  My reading of `CancellationTokenSource.CancelAfter()` and `Timer` source code is that it will behave as you desire.  

---

We'd accept a PR to change the [guard clauses in `TimeoutSyntax`](https://github.com/App-vNext/Polly/blob/50ff0d8161a78abe667dd325b11564518c133dd0/src/Polly.Shared/Timeout/TimeoutSyntax.cs#L79) and similarly [`TimeoutSyntaxAsync`](https://github.com/App-vNext/Polly/blob/50ff0d8161a78abe667dd325b11564518c133dd0/src/Polly.Shared/Timeout/TimeoutSyntaxAsync.cs), to formalise allowing [`Timeout.InfiniteTimeSpan`](https://github.com/dotnet/corefx/blob/103639b6ff5aa6ab6097f70732530e411817f09b/src/Common/src/CoreLib/System/Threading/Timeout.cs#L15).  Thanks for raising this!

urig:
I'd like to "grab" this. Would you say fixing the guard clauses and related unit tests would suffice? In other words, how would I go about verifying the correct behavior for the `Timeout` when set to `TimeSpan.Zero`?

reisenberger:
@urig Cool!  Thank you for offering to pick this up! 

[...]

PR #424, opened 2018-03-26
Guard clauses for TimeoutPolicy now allow Timout.InfiniteTimeSpan (Issue #419)

urig:
A `TimeSpan` value of `Timeout.InfiniteTimeSpan` indicates, well, an infinite timeout for the policy.

Please note this PR only addresses the sync version of the Policy. Async version to be address immediately following.

### DIFF
--- a/src/Polly.Shared/Polly.Shared.projitems
+++ b/src/Polly.Shared/Polly.Shared.projitems
@@ -135,7 +135,6 @@
     <Compile Include="$(MSBuildThisFileDirectory)Timeout\ITimeoutPolicy.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Timeout\TimeoutSyntax.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Timeout\TimeoutSyntaxAsync.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)Timeout\TimeoutSyntaxBase.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Timeout\TimeoutTResultSyntax.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Timeout\TimeoutTResultSyntaxAsync.cs" />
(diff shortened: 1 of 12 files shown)


reisenberger:
Thank you @urig , this is looking great!

Hmm. From the build error, it looks like .NET4.0 does not publicly expose the constant `System.Threading.Timeout.InfiniteTimeSpan` (altho my understanding is it behaves the same).  So we may need to define our own constant:

    internal static readonly TimeSpan InfiniteTimeSpan = System.Threading.Timeout.InfiniteTimeSpan;

and then use [#if directive forking](https://github.com/App-vNext/Polly/blob/7bdac691e0f02be9e8cd6c6afe74e2bf030b3616/src/Polly.Shared/Utilities/SystemClock.cs#L25-L29) to just [define it manually for .NET4.0](https://stackoverflow.com/a/30328117/).  Probably pull that into a `TimeoutSyntaxBase` base class (thoughts?) along with the `ValidateTimeoutIsInRange(...)` method, as they'll be used across several syntax classes? (see below).  

Let's also catch the:

    /// <exception cref="System.ArgumentOutOfRangeException">

intellisense for the methods you're updating, update them to the nice new message from  `ValidateTimeoutIsInRange(...)`!

I reckon that would be good to go then across the other syntax classes.  There's a very similar [`TimeoutTResultSyntax.cs`](https://github.com/App-vNext/Polly/blob/master/src/Polly.Shared/Timeout/TimeoutTResultSyntax.cs) class for the generic (`Policy<TResult>`), synchronous versions of TimeoutPolicy.  And then a similar pair (`TimeoutSyntaxAsync.cs`, `TimeoutTResultSyntaxAsync.cs`) for the async variants.

Thank you again for your contribution!

urig:
I have some questions about this. Will try to catch you on Slack!

reisenberger:
Thanks @urig . Merging! Will add credit and v5.9.0 doco for this later.

Issue #464, opened 2018-06-06
UP-FOR-GRABS: Refine debug experience via SourceLink


alexakryvenko:
Hi, 
I would like to request to remove .pdb files generation from Polly and serve them with separate Symbol package for debug purposes.
Having this .pdb files in Release mode of builds in the project made us manually exclude Polly pdbs from code coverage reports.

reisenberger:
That seems like a good suggestion.  We should probably [publish symbol packages](https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages) instead.  _Other community views on this?_

[SourceLink](https://github.com/dotnet/sourcelink) probably isn't an option for us as a library as of this date, as we don't (as a library) want to target a `netcoreapp2.1` tfm.  However, that's based on a quick read of the docs; happy for comment from anyone who has insight on this.

phatcher:
@reisenberger Have a look at https://github.com/Humanizr/Humanizer, they have successfully implemented GitLink and it works nicely

reisenberger:
Also the new `<DebugType>embedded</DebugType>` to consider, as an option for this.  Discussion and further links from here: https://twitter.com/KirillOsenkov/status/1007052524946255872

reisenberger:
It seems like [NewtonSoft.Json is successfully enabling SourceLink](https://twitter.com/JamesNK/status/1012198188520910848) and intends to start including .pdbs in nuget packages, as part of this.  Noting for reference.  We would have either that option of publishing symbol packages as [previously mentioned](https://github.com/App-vNext/Polly/issues/464#issuecomment-395220061).

reisenberger:
Improving debug experience around Polly looks like a great area to improve the Polly experience with minimal other impact.  

Three main options:

1. Create and publish [Symbol packages](https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages) to SymbolSource
2. Use [SourceLink](https://github.com/dotnet/sourcelink)  (more info also in earlier comments)
3. Use the new [portable pdbs](https://github.com/dotnet/core/blob/master/Documentation/diagnostics/portable_pdb.md), probably as `<DebugType>embedded</DebugType>` in the main Polly dll  (more info also in earlier comments) (would want to review size change of DLL; public reports seem v good but low-memory scenarios like Xamarin/embedded to keep in mind)

[...]

PR #497, opened 2018-09-09
reorg project

paulomorgado:
I've consolidated all projects into one Polly project and one Polly.Specs project.

Assembly signing and package signing is still not done.

Do you feel this is something worth doing? I feel it is way simpler to work with multi-targeting projects.

I've also added source linking. But I can add more repository and build information to the assembly and package.

Do you feel this is something worth doing? I've seen great benefits for the user when she/he can just debug into the libraries.


reisenberger:
@paulomorgado This looks very useful: **thank you** for taking on the build metadata improvements!

The PR looks to include a number of different concerns in the same PR (build metadata; project reorganisation; code style changes). Would you be happy to present just the build metadata changes in a first PR?  It would be awesome to get those benefits (cf #464) merged for the Polly community!  

(Otherwise it will have to wait until / depend on we agree all the other changes.)

Thank you again for contributing to Polly!

reisenberger:
Closing in favour of #504 and #502 .

PR #504, opened 2018-09-17
consolidate solution projects

paulomorgado:

Consolidate shared projects into one multi-targeting project.

Added .NET 4.6.2 and .NET 4.7.2 targets to unit tests.

To run .NET 4.6.2 and .NET 4.7.2 tests I had to update the tooling to the latest version.

That caused a few XUnit analyzer warnings that I explicitly ignored in the source files.

Although all tests run in the Visual Studio test runner, with `dotnet test` some are failing for .NET framework and I couldn't figure out why.

### Confirm the following

- [X]  I have merged the latest changes from the dev vX.Y branch
- [ ]  I have targeted the PR against the latest dev vX.Y branch


paulomorgado:
Inlining is always a good thing. For such methods I always add `[MethodImpl(MethodImplOptions.AggressiveInlining)]` to "help" the jitter.

The solution is choose better what to assert.

reisenberger:
That's a useful point to come back to on the production side (thanks for the reminder). On the test side, there can be trade-offs between performance and readability. These tests cover #270. The essence of the test is to assert where the exception expresses as thrown from, and that may be more immediately grokkable as a named method than inlined.

If we inline, the assert might end up something like:

    policy.Invoking(p => p.Execute(() => { throw exception; }))
        .ShouldThrow<AggregateException>()
        .WithMessage(exception.Message)
        .WithInnerException<NotImplementedException>()
        .Where(ex => IsAnonymousClosureWithinThisType(new StackTrace(ex, false).GetFrames()[0].GetMethod()));

where:

    private bool IsAnonymousClosureWithinThisType(MethodBase method)
    {
        return method.DeclaringType.DeclaringType == GetType()
               && method.DeclaringType.GetCustomAttribute<CompilerGeneratedAttribute>() != null;
    }

(but NetStandard1_1 doesn't support all this; and I would prefer not to assert on the string-forms of the way the compiler currently renders lambdas)

If we choose not to inline, the assert might express its purpose like this (renaming the method to `Foo_ThrowsException()` and attributing `[MethodImpl(MethodImplOptions.NoInlining)]`) :

    policy.Invoking(p => p.Execute(() => Foo_ThrowsException(exception)))
        .ShouldThrow<AggregateException>()
        .WithMessage(exception.Message)
        .WithInnerException<NotImplementedException>()
        .Where(ex => new StackTrace(ex, false).GetFrames()[0].GetMethod().Name == nameof(Foo_ThrowsException));

(`new StackTrace(ex, false).GetFrames()[0]` stems from the narrower overloads offered by System.Diagnostics.StackTrace v4.3.0 for NetStandard1_1)

Looking at the above, my feeling is that the clarity of test intent in the latter is more valuable to future devs coming to the codebase than any test performance gain from inlining (the method under consideration for inlining is used by 3 of 1700 tests).

Curious if anyone knows any more elegant way. Let me know any further thoughts; otherwise, we can push something like that onto the PR.

Thanks for pitching in on Polly.

reisenberger:
**TL;DR** For those not interested in the detail: The method under discussion for inlining is on the test side and used by only 3 of 1700 tests, so performance considerations are negligible. The essence of the test is to assert on where the exception is thrown from, and my feeling is that is clearer with a named method than inlined.

paulomorgado:
You might choose to run tests on non-optimized builds. Or there might be a way to instrument the JIT to not optimize.

paulomorgado:
@reisenberger, I misread your message. I thought you were talking about applying `[MethodImpl(MethodImplOptions.AggressiveInlining)]` to Polly code instead of test code.

I applied it to the test and all tests pass now.

reisenberger:
@paulomorgado Thanks again for this PR. I bumped the build version to achieve a green build, and re-targeted against the `App-vNext:onV610dev` base branch.  

Other feature-adding PRs (508, 513, 517) progressed while the method inlining was under discussion, so we cannot merge this PR immediately as is.  I will review and select merge order.

reisenberger:
Update; I am still holding (not merging) this PR while we have other significant PRs in the queue (508) with functionality built against the old project structure.

However, we will make this change to simplify the project/folder structure and to fall into line with recent [MS Open Source Project Guidance](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting), as soon as the functionality PRs are merged.

paulomorgado:
Thanks for the update, @reisenberger. Let me know if you need anything from me.

reisenberger:
This PR was unexpectedly automatically closed by github, when deleting a (now-outdated) base branch which the PR targeted .  

I intend that we consolidate the solution (per this PR) as a next step shortly, before any further work on Polly.  This will also open up the path for sourcelink.  Thanks @paulomorgado for all you originally did on this!

PR #517, opened 2018-10-15
Refactoring timeoutasync optimistic test

toddmeinershagen:

I am looking into why adding a Policy.TimeoutAsync<HttpResponseMessage>() to HttpClientFactory (from the AddPolicy() extension method in Microsoft.Extensions.Http.Polly library) does not allow the HttpClient to do its work and always ends up timing out.  

In the process of confirming the existing tests, I found that the test for TimeoutAsync that does not timeout when optimistically calling an async method was not really doing any work that could show blocking and the test was not properly using FluentAssertions ability to check for an exception for async tasks.  I have refactored the test to support these. 

Happily, I found that the policy works, just not in the context of the HttpClientFactory - so, I will be going to that code base next to validate.

Hope it helps.

### Confirm the following

- [x]  I have merged the latest changes from the dev vX.Y branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -1034,6 +1034,8 @@ When we discover an interesting write-up on Polly, we'll add it to this list. If
 * [Exception handling policies with Polly](http://putridparrot.com/blog/exception-handling-policies-with-polly/) - by [Mark Timmings](http://putridparrot.com/blog/about/)
 * [When you use the Polly circuit-breaker, make sure you share your Policy instances!](https://andrewlock.net/when-you-use-the-polly-circuit-breaker-make-sure-you-share-your-policy-instances-2/) - by [Andrew Lock](https://andrewlock.net/about/)
 * [Polly is Repetitive, and I love it!](http://www.appvnext.com/blog/2015/11/19/polly-is-repetitive-and-i-love-it) - by [Joel Hulen](http://www.thepollyproject.org/author/joel/)
+* [Using the Context to Obtain the Retry Count for Diagnostics](https://www.stevejgordon.co.uk/polly-using-context-to-obtain-retry-count-diagnostics) - by [Steve Gordon](https://twitter.com/stevejgordon)
+* [Passing an ILogger to Polly Policies](https://www.stevejgordon.co.uk/passing-an-ilogger-to-polly-policies) - by [Steve Gordon](https://twitter.com/stevejgordon)
 
 ## Podcasts
 
--- a/src/Polly.SharedSpecs/Timeout/TimeoutAsyncSpecs.cs
+++ b/src/Polly.SharedSpecs/Timeout/TimeoutAsyncSpecs.cs
@@ -213,24 +213,19 @@ namespace Polly.Specs.Timeout
         }
 
         [Fact]
-        public async Task Should_not_throw_when_timeout_is_greater_than_execution_duration__pessimistic()
+        public void Should_not_throw_when_timeout_is_greater_than_execution_duration__pessimistic()
         {
             var policy = Policy.TimeoutAsync(TimeSpan.FromSeconds(1), TimeoutStrategy.Pessimistic);
 
             ResultPrimitive result = ResultPrimitive.Undefined;
-            Exception ex = null;
 
-            try
+            policy.Awaiting(async p =>
             {
-                result = await policy.ExecuteAsync(() => Task.FromResult(ResultPrimitive.Good))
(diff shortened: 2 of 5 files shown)


reisenberger:
@toddmeinershagen Re:

>Happily, I found that Policy.TimeoutAsync() works, just not in the context of the HttpClientFactory

Does our [documentation on using TimeoutPolicy with HttpClientFactory](https://github.com/App-vNext/Polly/wiki/Polly-and-HttpClientFactory#use-case-applying-timeouts) help with the behaviour you are seeing?

toddmeinershagen:
Actually, no.  I have read and applied the documentation and am still
seeing the behavior.  I only have one policy added to the HttpClientFactory
for my type.  If I use it, the method always times out -- as though it is
holding up execution.  If not, it executes.  When I looked over at the code
for Microsoft.Extensions.Http.Polly, I noticed that they are missing
async/await operators.  So, I think that code base is the likely culprit --
not Polly proper.

Hope that helps.

reisenberger:
Thanks @toddmeinershagen for the great spot 👀 and bringing this test into line with others with `.Should/ShouldNotThrow<>()`! Happy also with the idea to express a `Task.Delay` within the test (expresses the test intent more clearly).

If you want to work this into a full PR, we should:

+ Fix other instances of `Should_not_throw_when_timeout_is_greater_than_execution_duration__optimistic()` in the solution (async-`TResult` case; and two sync cases) similarly.
+ Fix 4x `Should_not_throw_when_timeout_is_greater_than_execution_duration__pessimistic()` similarly.
+ But instead of `Task.Delay`, use `SystemClock.Sleep(...)` or `SystemClock.SleepAsync(...)` (as appropriate) in each case.  **Explanation:** The timeout implementations [abstract the notion of time-based cancellation](https://github.com/App-vNext/Polly/blob/2b868ef1b14cd1a8573f1c0bcdefd61701716a95/src/Polly.Shared/Timeout/TimeoutEngineAsync.cs#L33) and the timeout tests [use](https://github.com/App-vNext/Polly/blob/2b868ef1b14cd1a8573f1c0bcdefd61701716a95/src/Polly.SharedSpecs/Timeout/TimeoutAsyncSpecs.cs#L210) and [manipulate the abstraction of](https://github.com/App-vNext/Polly/blob/2b868ef1b14cd1a8573f1c0bcdefd61701716a95/src/Polly.SharedSpecs/Timeout/TimeoutSpecsBase.cs#L24-L84) our [abstracted system clock](https://github.com/App-vNext/Polly/blob/2b868ef1b14cd1a8573f1c0bcdefd61701716a95/src/Polly.Shared/Utilities/SystemClock.cs), with a simulated cancel-when-time-passes mechanism, in order that we can make the timeout tests run quasi-instantaneously rather than incur real time delays.  (For a single test the saving of `500ms` may not seem much, but these principles across the whole test codebase of ~1700 tests drove test execution time down from order-of-10-minutes to ~30 seconds.)

(If you're not available/interested to take that on, I'll tidy it in due course.)

toddmeinershagen:
@reisenberger I think I got all of the changes that you suggested.  Should I issue a pull request from my feature branch to the master branch or the onV610dev?

reisenberger:
@toddmeinershagen I have retargeted this PR against the `onV610dev` branch.

> I think I got all of the changes that you suggested.

There are similar tests `Should_not_throw_when_timeout_is_greater_than_execution_duration__optimistic()` and `Should_not_throw_when_timeout_is_greater_than_execution_duration__pessimistic()` which it would be good to make consistent with the async changes (ie add some `SystemClock.Sleep(...)` shorter than the configured timeout).  Are you happy to add those?

Thank you for contributing to Polly!

reisenberger:
@toddmeinershagen You will also need to sign the [CLA agreement](https://github.com/App-vNext/Polly/pull/517#issuecomment-430000115) in order for us to be able to use your changes (this is policy for .NET Foundation projects).  Many thanks.

toddmeinershagen:
@reisenberger I signed the CLA.  Sorry it took a few days.  I was waiting to hear back from the Legal department at my company.  Let me know if you need anything else.  Hope it helps!  (Thank you for your help as well.)

reisenberger:
@toddmeinershagen Thank you. No problem; we understand corporate legal departments need to see this.

reisenberger:
@toddmeinershagen Thank you for your contribution to Polly!  Merged to the dev branch!

Issue #526, opened 2018-11-05
Feature suggestion: In-box DecorrelatedJitter


grant-d:
If DecorrelatedJitter is a common-enough requirement, could we turn it into a first class Polly citizen.
I am happy to submit a PR. Otherwise, proposed solution is below.
This is a little different to what's shown in the wiki; it's an encapsulated and testable class.

**[Edited to fix bug]**
```csharp
    // Background: https://www.awsarchitectureblog.com/2015/03/backoff.html
    public sealed class DecorrelatedJitter
    {
        [ThreadStatic]
        private static readonly Random s_random = new Random(); // Default ctor uses a random seed
        
        private readonly Random _random;
        public int RetryCount { get; }
        public TimeSpan MinDelay { get; }
        public TimeSpan MaxDelay { get; }

        public DecorrelatedJitter(int retryCount, TimeSpan minDelay, TimeSpan maxDelay, Random random = null)
        {
            if (retryCount < 0) throw new ArgumentOutOfRangeException(nameof(retryCount));
            if (minDelay < TimeSpan.Zero) throw new ArgumentOutOfRangeException(nameof(minDelay));
            if (maxDelay < minDelay) throw new ArgumentOutOfRangeException(nameof(maxDelay));
            
            _random = random ?? s_random; // Custom Random enables unit-testing
            RetryCount = retryCount;
            MinDelay = minDelay;
            MaxDelay = maxDelay;
        }

        public IEnumerable<TimeSpan> Generate()
        {
            var delays = new TimeSpan[RetryCount];

            double range = MaxDelay.TotalMilliseconds - MinDelay.TotalMilliseconds;  // Range

            for (var i = 0; i < delays.Length; i++)
            {
                var ms = range * _random.NextDouble(); // Ceiling
                ms += MinDelay.TotalMilliseconds; // Floor

                delays[i] = TimeSpan.FromMilliseconds(ms);
            }

            return delays;
        }
    }
```

Here's an example of how to call it. Note that the class is instantiated as a singleton (directly or via DI) and each execution uses the `.Generate()`.
```csharp
private static readonly DecorrelatedJitter s_jitterer = new DecorrelatedJitter(RetryCount, MinDelay, MaxDelay);

return Policy<IRestResponse<T>>
    .Handle<HttpRequestException>()
    .OrResult(resultPredicate)
    .WaitAndRetryAsync(s_jitterer.Generate()) // <-- note usage
[...]

Issue #530, opened 2018-11-09
Decorrelated jitter does not jitter well


george-polevoy:
**Summary: Jitter the retry delays to achieve low concurrency**

<br/>
<br/>

**I'm trying to use the suggested "decorrelated jitter" strategy**  

![jittering](https://user-images.githubusercontent.com/2792885/48259864-6fc8f380-e42a-11e8-9b9d-2311ce3b3e5d.PNG)

In my experiment, the [suggested decorrelated jitter](https://github.com/App-vNext/Polly/wiki/Retry-with-jitter) does not achieve what it supposed to.

First, `new Random()` is not quite random, when it comes to reducing concurrency.
You can get multiple instances with the same seed, if many retries are issued simultaneously.

Following code achieves much lower concurrency and better distribution.
In the attached screenshot you can see that the 'advanced' graph has much lower maximum and smoother distribution, which is highly desired to avoid concurrency spikes.

```
IEnumerable<TimeSpan> DecorrelatedExponent()
{
    var r = new Random(Guid.NewGuid().GetHashCode());
    for (var softAttemptCount = 0.0; softAttemptCount < 4; ) {
        softAttemptCount += r.NextDouble() * 2;
        yield return TimeSpan.FromSeconds(
            Math.Pow(2, softAttemptCount) * random.NextDouble());
    }
}

```

reisenberger:
Thank you @george-polevoy for diving deeper on this than I had time to and for sharing that with the community, this is super interesting.  The [decorrelated Jitter strategy quoted](https://github.com/App-vNext/Polly/wiki/Retry-with-jitter#more-complex-jitter) came from that [AWS team research](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/), definitely open to investigating different options.

First, could you say what the axes on the graphs represent? <sup>(A guess ... the number of instances (vertical) of a specific delay of some-or-other timespan (horizontal) when averaged over 100000 iterations of the algorithm ... but you can state precisely? )</sup>

Second, would you be happy to share the test code which generated the results? (maybe in a ZIP or github gist). It would be great for you/ I/ others to be able to validate, and to build further on your results and explore how tweaking the algorithms affects outcomes.   

>Following code achieves much better distribution

I am definitely seeing in your graphs that the green line is much more smoothly distributed than the yellow line as generated in the tests! :+1: 

>Following code achieves much lower latency

[...]

Issue #143, opened 2016-07-25
Using Configuration to Setup Polly


RehanSaeed:
I've written the following extension method to help me configure Polly from my ASP.NET Core JSON configuration file. It may help others if something more complete was included in Polly itself.

```
Policy policy = Policy.Handle<Exception>().FromOptions(options);

public class PolicyOptions
{
    public TimeSpan? DurationOfBreak { get; set; }

    public int? ExceptionsAllowedBeforeBreaking { get; set; }

    public int? RetryCount { get; set; }

    public bool RetryForever { get; set; }
}

public static Policy FromOptions(this PolicyBuilder policyBuilder, PolicyOptions options)
{
    if (policyBuilder == null)
    {
        throw new ArgumentNullException(nameof(policyBuilder));
    }

    if (options == null)
    {
        throw new ArgumentNullException(nameof(options));
    }

    Policy policy;
    if (options.RetryCount.HasValue)
    {
        policy = policyBuilder.RetryAsync(options.RetryCount.Value);
    }
    else if (options.RetryForever)
    {
        policy = policyBuilder.RetryForeverAsync();
    }
    else if (options.DurationOfBreak.HasValue && options.ExceptionsAllowedBeforeBreaking.HasValue)
    {
        policy = policyBuilder.CircuitBreakerAsync(options.ExceptionsAllowedBeforeBreaking.Value, options.DurationOfBreak.Value);
    }
    else
    {
        throw new ArgumentException(
            $"Invalid Policy Options.",
            nameof(options));
    }

    return policy;
}
```

reisenberger:
@RehanSaeed Thanks for sharing this.  There are some similar ideas in issue #50, with discussion around the different aspects of a policy that could be configured (numeric parameters; exceptions handled?); different configuration sources and methods of application; and configuration-only-at-startup versus reconfiguration-during-execution.   Please feel free to have a read of #50, and add any further ideas/comments!

RehanSaeed:
While the two issues talk about config, I feel they are talking about two different kinds of config. That issue is talking about using `System.Configuration` which should probably exist as a separate NuGet package. What I'm talking about is being able to create a policy based on a POCO with some very simple properties. This POCO can then be used to bind to a JSON config file. This POCO should probably live in the Polly core NuGet package.

reisenberger:
@RehanSaeed Agreed: if a strategy for configuring policies depends on approaches not part of .NET Core, that would certainly be part of a separate package.

[...]

PR #536, opened 2018-11-14
DecorrelatedJitter is a first-class citizen

grant-d:
### `ISleepDurationSeriesStrategy` providers

Per my design proposal in  #526, this PR includes changes that make `DecorrelatedJitterBackoff`, `ConstantBackoff`, `LinearBackoff` and `ExponentialBackoff` first-class citizens in Polly.

### Questions

Details in the linked issue.
* Any other backoff strategies we need **[Edit: No]**
* Is the interface sufficient **[Edit: Trimmed]**
* I have added **one** overload to `PolicyBuilder` to demonstrate the pattern should we decide to integrate it directly. See additional notes there. I am happy to revert such integration if need be. **[Edit: Scoped-out]**
* What about the `*Forever` overloads, I have not integrated that at all. **[Edit: Scoped-out]**

### Notes
* The shared `Random` instance is `lock` controlled (per separate discussion)
* I have made lots of comments on the CR to explain what's going on
* Please forgive a small amount of reformatting - my VS setup is configured to do it automatically. I will revert if necessary.

### Usage
Details are in the associated design document. But for quick reference, here's how you would call a strategy. Of course, we may decide to provide direct overloads too:
Here's an example of how to call it. Note that the class is instantiated as a singleton (directly or via DI) and each execution uses the `.Generate()`.
```csharp
private static readonly ISleepDurationStrategy s_backoff = new DecorrelatedJitterBackoff(MinDelay, MaxDelay);

return Policy<IRestResponse<T>>
    .Handle<HttpRequestException>()
    .OrResult(resultPredicate)
    .WaitAndRetryAsync(s_backoff.Generate(retryCount: 5)) // <-- Note usage
    .ExecuteAsync(() => client.ExecuteTaskAsync<T>(request, cancellationToken));
```

## Risks
* **Mitigated**: the use of `yield` instead of an eager `TimeSpan[]` array has mitigated the following concern: Inadequate guidance could lead to consumers using the exact same instance of a strategy across multiple calls. It is **essential** that `.Generate` is called just-in-time, not pre-generated. (Docs should include info in https://github.com/App-vNext/Polly/pull/536#issuecomment-441363716)

### Confirm the following
~_(I am not sure which of the 2 dev branches to merge from and target)_~ **EDIT by @reisenberger** : I fixed this and targeted the branch.

- [ ]  I have merged the latest changes from the dev vX.Y branch


src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,118 @@
...
+            }
+
+            return delays;
+        }
+
+        /// <summary>
+        /// Generate a continuous sequence of <see cref="TimeSpan"/> values to use as sleep-durations.
+        /// </summary>
+        /// <param name="retryCount">The maximum number of retries to use, in addition to the original call.</param>
+        public IEnumerable<TimeSpan> Continuous(int retryCount)
```
grant-d: This method is meant to be used in the `Forever()` methods. But not sure if it's adequate or suitable.

src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,118 @@
...
+            if (FastFirst)
+            {
+                i++;
+                yield return TimeSpan.Zero;
+            }
+
+            double range = MaxDelay.TotalMilliseconds - MinDelay.Milliseconds;
+            double max = MinDelay.TotalMilliseconds;
+
+            for (; i < retryCount; i++)
```
grant-d: It returns the expected values for the first N retries, then (see next comment)

src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,118 @@
...
+                double ms = range * _random.NextDouble(); // Range
+                ms += MinDelay.TotalMilliseconds; // Floor
+                ms = Math.Min(ms, MaxDelay.TotalMilliseconds); // Ceiling
+
+                max = Math.Max(ms, max); // Extra
+
+                yield return TimeSpan.FromMilliseconds(ms);
+            }
+
+            while (true)
```
grant-d: ...then returns the `max` value thereafter

src/Polly.Shared/Duration/DurationUtils.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Polly.Duration
+{
+    /// <summary>
+    /// Utilities useful for retry duration management.
+    /// </summary>
+    internal static class DurationUtils
+    {
+        [ThreadStatic]
+        private static readonly Random s_random = new Random(); // Default ctor uses a time-based seed
```
grant-d: I will noodle on exactly how we do this - `[ThreadStatic]` or `lock`

src/Polly.Shared/Retry/RetryTResultSyntax.cs
```diff
@@ -208,9 +209,7 @@ public static RetryPolicy<TResult> RetryForever<TResult>(this PolicyBuilder<TRes
         /// <returns>The policy instance.</returns>
         public static RetryPolicy<TResult> WaitAndRetry<TResult>(this PolicyBuilder<TResult> policyBuilder, int retryCount, Func<int, TimeSpan> sleepDurationProvider)
         {
-            Action<DelegateResult<TResult>, TimeSpan, int, Context> doNothing = (_, __, ___, ____) => { };
-
-            return policyBuilder.WaitAndRetry(retryCount, sleepDurationProvider, doNothing);
+            return policyBuilder.WaitAndRetry(retryCount, sleepDurationProvider, DoNothing);
```
grant-d: Move the lambda into a `private static` function - it's better for perf

src/Polly.Shared/Retry/RetryTResultSyntax.cs
```diff
@@ -95,12 +96,12 @@ public static RetryPolicy<TResult> Retry<TResult>(this PolicyBuilder<TResult> po
             return new RetryPolicy<TResult>(
                 (action, context, cancellationToken) => RetryEngine.Implementation(
                     action,
-                    context, 
+                    context,
```
grant-d: No changes here; this is just my VS setup cleaning up the formatting

src/Polly.Shared/Retry/RetryTResultSyntax.cs
```diff
@@ -241,6 +240,37 @@ public static RetryPolicy<TResult> WaitAndRetry<TResult>(this PolicyBuilder<TRes
...
+        /// <param name="sleepDurationStrategy">The class that provides the duration to wait for for a particular retry attempt.</param>
+        /// <param name="onRetry">The action to call on each retry.</param>
+        /// <returns>The policy instance.</returns>
+        /// <exception cref="System.ArgumentOutOfRangeException">retryCount;Value must be greater than or equal to zero.</exception>
+        /// <exception cref="System.ArgumentNullException">
+        /// sleepDurationProvider
+        /// or
+        /// onRetry
+        /// </exception>
+        public static RetryPolicy<TResult> WaitAndRetry<TResult>(this PolicyBuilder<TResult> policyBuilder, int retryCount, ISleepDurationStrategy sleepDurationStrategy, Action<DelegateResult<TResult>, TimeSpan> onRetry)
```
grant-d: This overload uses `ISleepDurationStrategy` instead of `IEnumerable<TimeSpan>`

src/Polly.Shared/Retry/RetryTResultSyntax.cs
```diff
@@ -241,6 +240,37 @@ public static RetryPolicy<TResult> WaitAndRetry<TResult>(this PolicyBuilder<TRes
...
+        {
+            if (onRetry == null) throw new ArgumentNullException(nameof(onRetry));
+            if (sleepDurationStrategy == null) throw new ArgumentNullException(nameof(sleepDurationStrategy));
+
+            IReadOnlyList<TimeSpan> delays = sleepDurationStrategy.Discrete(retryCount);
```
grant-d: Here's where it pre-generates the `TimeSpan` series

src/Polly.Shared/Retry/RetryTResultSyntax.cs
```diff
@@ -241,6 +240,37 @@ public static RetryPolicy<TResult> WaitAndRetry<TResult>(this PolicyBuilder<TRes
...
+
+            return policyBuilder.WaitAndRetry(
+                retryCount,
+                (i, _, __) => delays[i],
```
grant-d: And then calls the existing overload that handles the `TimeSpan[]`

src/Polly.Shared/Retry/RetryTResultSyntax.cs
```diff
@@ -319,9 +349,7 @@ public static RetryPolicy<TResult> WaitAndRetry<TResult>(this PolicyBuilder<TRes
         /// <returns>The policy instance.</returns>
         public static RetryPolicy<TResult> WaitAndRetry<TResult>(this PolicyBuilder<TResult> policyBuilder, int retryCount, Func<int, Context, TimeSpan> sleepDurationProvider)
         {
-            Action<DelegateResult<TResult>, TimeSpan, int, Context> doNothing = (_, __, ___, ____) => { };
-
-            return policyBuilder.WaitAndRetry(retryCount, sleepDurationProvider, doNothing);
+            return policyBuilder.WaitAndRetry(retryCount, sleepDurationProvider, DoNothing);
```
grant-d: Another example of moving lambda into shared local static function

src/Polly.SharedSpecs/Retry/WaitAndRetryTResultSpecs.cs
```diff
@@ -25,17 +27,17 @@ public void Should_be_able_to_calculate_retry_timespans_based_on_the_handled_fau
                 {ResultPrimitive.FaultAgain, 4.Seconds()},
             };
 
-            var actualRetryWaits = new List<TimeSpan>();
+            List<TimeSpan> actualRetryWaits = new List<TimeSpan>();
```
grant-d: I will try revert these - VS is overly ambitious with its formatting

src/Polly.Shared/Duration/ConstantBackoff.cs
```diff
@@ -0,0 +1,83 @@
+using System;
+using System.Collections.Generic;
+
+namespace Polly.Duration
```
grant-d: New `namespace`. Happy to change if necessary

src/Polly.Shared/Duration/ISleepDurationStrategy.cs
```diff
@@ -0,0 +1,30 @@
+using System;
+using System.Collections.Generic;
+
+namespace Polly.Duration
+{
+    /// <summary>
+    /// An interface defining strategies for managing the sleep-duration of retries. 
+    /// </summary>
+    public interface ISleepDurationStrategy
```
grant-d: Happy to discuss a better name if necessary

src/Polly.Shared/Duration/ConstantBackoff.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            Delay = delay;
+            FastFirst = fastFirst;
+        }
+
+        /// <summary>
+        /// Generate the sequence of <see cref="TimeSpan"/> values to use as sleep-durations.
+        /// </summary>
+        /// <param name="retryCount">The maximum number of retries to use, in addition to the original call.</param>
+        public IReadOnlyList<TimeSpan> Discrete(int retryCount)
```
grant-d: Note how `retryCount` is not a `property` (ie passed in the `ctor`) but a `param`. 
I used to have it as a property but I thought that making it a param would let you reuse the strategy and call it with different counts.

src/Polly.Shared/Duration/ConstantBackoff.cs
```diff
@@ -0,0 +1,83 @@
...
+            }
+
+            return delays;
+        }
+
+        /// <summary>
+        /// Generate a continuous sequence of <see cref="TimeSpan"/> values to use as sleep-durations.
+        /// </summary>
+        /// <param name="retryCount">The maximum number of retries to use, in addition to the original call.</param>
+        public IEnumerable<TimeSpan> Continuous(int retryCount)
```
grant-d: This method is meant to be used in the `Forever()` methods. But not sure if it's adequate or suitable.

src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,118 @@
...
+            {
+                yield return TimeSpan.FromMilliseconds(max);
```
grant-d: This shows the `max' thing better (see previous comment in this regard)

src/Polly.Shared/Retry/RetryTResultSyntax.cs
```diff
@@ -732,5 +834,12 @@ public static RetryPolicy<TResult> WaitAndRetryForever<TResult>(this PolicyBuild
                 policyBuilder.ResultPredicates
             );
         }
+
+        #region Helpers
+
+        private static void DoNothing<TResult>(DelegateResult<TResult> outcome, TimeSpan span, int i, Context ctx)
```
grant-d: Here's that lambda replacement

src/Polly.Shared/Retry/RetryTResultSyntax.cs
grant-d: Note that we don't have to implement any of these functions yet. We can provide the strategies and let consumers use them directly to generate `TimeSpan[]` series with which they can call the `IEnumerable<TimeSpan>` overloads.

src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        #region Nested
+
+        /// <summary>
+        /// A random number generator with a Uniform distribution that is thread-safe (via locking).
+        /// Can be instantiated with a custom <see cref="Random"/> instance, for example to make
+        /// it act in a deterministic manner.
+        /// </summary>
+        private sealed class ConcurrentRandom
```
grant-d: Per separate discussion, here's the `lock` controlled instance of `Random`

src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,167 @@
...
+            /// <summary>
+            /// Returns a random floating-point number that is greater than or equal to 0.0,
+            /// and less than 1.0.
+            /// This method uses locks in order to avoid issues with concurrent access.
+            /// </summary>
+            public double NextDouble()
+            {
+                // It is safe to lock on _random since it's not directly exposed 
+                // to outside use, so it cannot be locked externally.
+                lock (_random)
```
grant-d: Note `lock` around the only method we expose

reisenberger:
@grant-d Thank you. A courtesy alert that private commitments and paid work mean that I likely first able to look at this at weekend or Friday evening.

src/Polly.Shared/Retry/RetryTResultSyntax.cs
```diff
@@ -241,6 +242,37 @@ public static RetryPolicy<TResult> WaitAndRetry<TResult>(this PolicyBuilder<TRes
...
+        /// sleepDurationProvider
+        /// or
+        /// onRetry
+        /// </exception>
+        public static RetryPolicy<TResult> WaitAndRetry<TResult>(this PolicyBuilder<TResult> policyBuilder, int retryCount, ISleepDurationStrategy sleepDurationStrategy, Action<DelegateResult<TResult>, TimeSpan> onRetry)
+        {
+            if (onRetry == null) throw new ArgumentNullException(nameof(onRetry));
+            if (sleepDurationStrategy == null) throw new ArgumentNullException(nameof(sleepDurationStrategy));
+
+            TimeSpan[] delays = sleepDurationStrategy.Generate(retryCount).ToArray();
```
grant-d: Here's where it _pre-generates_ the TimeSpan series

src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,150 @@
...
+            MaxDelay = maxDelay;
+            FastFirst = fastFirst;
+        }
+
+        /// <summary>
+        /// Generate the sequence of <see cref="TimeSpan"/> values to use as sleep-durations.
+        /// For example: 1s, 3s, 4s, 2s...
+        /// </summary>
+        /// <param name="retryCount">The maximum number of retries to use, in addition to the original call.</param>
+        public IEnumerable<TimeSpan> Generate(int retryCount)
```
grant-d: Note how `retryCount` is not a `property` (ie passed in the `ctor`) but a `param`. 
I used to have it as a property but I thought that making it a param would let you reuse the strategy and call it with different counts. See the example in the PR notes.

src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        /// <summary>
+        /// Creates a new instance of the class.
+        /// </summary>
+        /// <param name="minDelay">The minimum duration value to use for each retry.</param>
+        /// <param name="maxDelay">The maximum duration value to use for each retry.</param>
+        /// <param name="fastFirst">Whether the first retry will be immediate or not.</param>
+        /// <param name="random">An optional <see cref="Random"/> instance to use.
+        /// If not specified, will use a shared (singleton) instance with a random seed.</param>
+        public DecorrelatedJitterBackoff(TimeSpan minDelay, TimeSpan maxDelay, bool fastFirst = false, Random random = null)
```
grant-d: The last param `random` is there for determinism and/or unit testing. You may want a specific seed or you may want a repeatable sequence (as you often do in units)

src/Polly.Shared/Duration/DurationExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <summary>
+        /// Generate a continuous sequence of <see cref="TimeSpan"/> values to use as sleep-durations.
+        /// The first <paramref name="retryCount"/> durations are generated in the same manner as
+        /// the corresponding <see cref="ISleepDurationStrategy"/>, and thereafter the maximum value 
+        /// from that sequence is returned.
+        /// For example: 2s, 4s, 6s, 8s; 8s, 8s, 8s...
+        /// </summary>
+        /// <param name="strategy">The duration strategy to use.</param>
+        /// <param name="retryCount">The maximum number of retries to use, in addition to the original call.</param>
+        public static IEnumerable<TimeSpan> Continuous(this ISleepDurationStrategy strategy, int retryCount)
```
grant-d: This method is meant to be used in the `Forever()` methods. But not sure if it's adequate.
It produces an expected series of `TimeSpan` values for the first `retryCount` iterations, and thereafter returns the `max` value it found in that sequence. 
Some of the other strategies demonstrate this better.

src/Polly.Shared/Retry/RetryTResultSyntax.cs
```diff
@@ -241,6 +242,37 @@ public static RetryPolicy<TResult> WaitAndRetry<TResult>(this PolicyBuilder<TRes
...
+
```
grant-d: IEnumerable will be better here

src/Polly.SharedSpecs/Duration/LinearBackoffSpecs.cs
```diff
@@ -0,0 +1,232 @@
...
+                discrete1.First().Should().Be(discrete2.First());
+            }
+
+            discrete1.Skip(2).First().Should().NotBe(discrete2.Skip(2).First());
+        }
+
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public static void Should_have_an_adequate_variance_when_range_small(bool fastFirst)
```
reisenberger: Just `Should_have_expected_variance(...)` ?

src/Polly.SharedSpecs/Duration/LinearBackoffSpecs.cs
```diff
@@ -0,0 +1,232 @@
...
+                .GroupBy(n => n)
+                .Count();
+
+            groupCount.Should().Be(expectedCount);
+        }
+
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public static void Should_have_an_adequate_variance_when_range_large(bool fastFirst)
```
reisenberger: Duplicate of above test?

src/Polly.SharedSpecs/Duration/LinearBackoffSpecs.cs
```diff
@@ -0,0 +1,232 @@
...
+                .GroupBy(n => n)
+                .Count();
+
+            groupCount.Should().Be(expectedCount);
+        }
+
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public static void Should_have_no_variance_when_range_zero(bool fastFirst)
```
reisenberger: Could have similar test: `Should_have_no_variance_when_factor_zero(...)` ?

(where `minDelay > TimeSpan.Zero`)

src/Polly.Shared/Duration/LinearBackoff.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        /// <summary>
+        /// Whether the first retry will be immediate or not.
+        /// </summary>
+        public bool FastFirst { get; }
+
+        /// <summary>
+        /// The duration value for the first retry.
+        /// </summary>
+        public TimeSpan Delay { get; }
```
reisenberger: Could call this `InitialDelay` (and the corresponding ctor param, `initialDelay`). I realise this changes the pattern (such that not all similar are called Delay), but it is more directly what it is?

src/Polly.Shared/Retry/RetryTResultSyntax.cs
grant-d: Per decision in previous comment, we won't be integrating new overloads for now

src/Polly.Shared/Retry/RetryTResultSyntax.cs
grant-d: Per decision in previous comment, we won't be integrating new overloads for now

src/Polly.Shared/Duration/ConstantBackoff.cs
grant-d: Done

src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
grant-d: Good idea, will leave as-is for now

src/Polly.Shared/Duration/DurationExtensions.cs
grant-d: Agree - I like cutting scope too; it can be painful but then you don't end up with a spaghetti api after

src/Polly.SharedSpecs/Duration/ConstantBackoffSpecs.cs
```diff
@@ -0,0 +1,124 @@
...
+                .GroupBy(n => n)
+                .Count();
+
+            groupCount.Should().Be(1);
+        }
+
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public static void Should_have_an_adequate_variance_when_range_large(bool fastFirst)
```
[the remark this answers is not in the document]
grant-d: Nice catch, that is indeed redundant, removed

src/Polly.SharedSpecs/Duration/LinearBackoffSpecs.cs
grant-d: Fixed

src/Polly.SharedSpecs/Duration/LinearBackoffSpecs.cs
grant-d: Removed

src/Polly.SharedSpecs/Duration/LinearBackoffSpecs.cs
grant-d: Done

src/Polly.Shared/Duration/LinearBackoff.cs
grant-d: Agree. Should we make the same change on `ExponentialBackoff`?
It's a little more ambiguous there, since it is indeed the initial-delay, but it's also the factor in the ongoing calculation

grant-d:
> which params should sit in the ctors, and which in the [methods]

Agree. I went back & forth on that many times. I tried to balance which would be 'constant' in a policy library, and which you would want to change on each instantiation. I ended up with the latter having `count` being the minimum requirement, but we can move nearly all the other params there too. It doesn't have a clear answer; both are suitable in different circumstances

src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,141 @@
...
+                yield return TimeSpan.Zero;
+            }
+
+            double ms = MinDelay.TotalMilliseconds;
+            for (; i < retryCount; i++)
+            {
+                // https://github.com/aws-samples/aws-arch-backoff-simulator/blob/master/src/backoff_simulator.py#L45
+                // self.sleep = min(self.cap, random.uniform(self.base, self.sleep * 3))
+                ms = _random.Uniform(MinDelay.TotalMilliseconds, ms * 3);
+                ms = Math.Min(MaxDelay.TotalMilliseconds, ms);
```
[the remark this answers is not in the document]
grant-d: Let's see what @george-polevoy comes back with on his modeling experiments. If we follow the AWS paper then the current algo is correct, but I also noticed the high occurrence of `max` in the output. Let's discuss some more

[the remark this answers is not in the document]
grant-d: Actually, I take that back. I like your variant better than mine, and it does not stray from the paper. Done.

src/Polly.Shared/Duration/ISleepDurationStrategy.cs
grant-d: I am now debating that, in the absence of the integration overloads, whether we need to make this interface `public` anymore. If we made it `internal` it would be less api noise. But in that case consumers would not be able to write their own extensions over any concrete class. For example, a consumer would not be able to write a `Continuous` extension like I had previously.
So maybe it's good to keep the interface public?

reisenberger:
Re the following note added to the head of the issue:

>## Risks
>Inadequate guidance could lead to consumers using the exact same instance of a strategy across multiple calls. It is **essential** that `.Generate` is called just-in-time, not pre-generated.

This may not be quite as risky as that note may suggest, because of the way `IEnumerable<TimeSpan>` represents a _recipe for getting_ a sequence of timespans, not the timespans themselves. And because Polly  reifies that enumerable at execution time, not at configuration time.  It _is_ essential (for jitter) that the sequence is not pre-generated, but it is _not_ essential for users to create/configure a policy with `DecorrelatedJitterBackoff.Generate(...)` only just before use.  It will be perfectly fine for users to configure a policy like this once at startup, even to hold that policy in a `PolicyRegistry` as a singleton:

    var jitterPolicy = Policy
        .Handle<HttpRequestException>()
        .OrResult<HttpResponseMessage>(resultPredicate)
        .WaitAndRetryAsync(s_backoff.Generate(retryCount: 5));

and then perfectly safe for that policy instance to be re-used multiple times/places all round the app - it will still generate different jitter sequences.

`IEnumerable<TimeSpan>` represents a recipe for getting a sequence of timespans, not the timespans themselves. The `.Generate(...)` method is only actually executed (any enumerable is only actually reified) when code calls `GetEnumerator(...)` on the `IEnumerable`. This happens in Polly [for example here](https://github.com/App-vNext/Polly/blob/f9e9c88e8ca80a746dce27dccc78543fb82658ab/src/Polly.Shared/Retry/RetryStateWaitAndRetryAsync.cs#L18) on a _per-execution basis_ when the retry state for an execution is instantiated.

What users must avoid (if they want the policy to be re-usable and generate random jitter each use) is reifying the sequence at policy configuration time, ie avoid `.WaitAndRetryAsync(s_backoff.Generate(5).ToList());`

grant-d:
> This is not as risky as that note may suggest

Thanks @reisenberger, that's a fair point. In a previous incarnation of the code, I was eagerly generating the durations into a `TimeSpan[]` array, so it was reified once and only once. I wasn't thinking straight when I made that note, since I had subsequently backed down from that and used `yield` instead.

src/Polly.Shared/Duration/ConstantBackoff.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </summary>
+    public sealed class ConstantBackoff : ISleepDurationStrategy
+    {
+        /// <summary>
+        /// Whether the first retry will be immediate or not.
+        /// </summary>
+        public bool FastFirst { get; }
+
+        /// <summary>
+        /// The duration value for the first retry.
```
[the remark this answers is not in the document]
grant-d: Fixed

src/Polly.Shared/Duration/ConstantBackoff.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// </summary>
+        public TimeSpan Delay { get; }
+
+        /// <summary>
+        /// Creates a new instance of the class.
+        /// </summary>
+        /// <param name="delay">The duration value for the first retry.</param>
```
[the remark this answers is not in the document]
grant-d: Fixed

src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,141 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Polly.Duration
+{
+    /// <summary>
+    /// Generates sleep durations in an jittered manner, making sure to mitigate any correlations.
+    /// For example: 1s, 3s, 4s, 2s...
```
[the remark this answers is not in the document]
grant-d: Ah, good point. Fixed

src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,141 @@
...
+    {
+        private readonly ConcurrentRandom _random;
+
+        /// <summary>
+        /// Whether the first retry will be immediate or not.
+        /// </summary>
+        public bool FastFirst { get; }
+
+        /// <summary>
+        /// The minimum duration value to use for each retry.
```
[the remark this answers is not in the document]
grant-d: Agree. Fixed

src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,141 @@
...
+        public TimeSpan MaxDelay { get; }
+
+        /// <summary>
+        /// Creates a new instance of the class.
+        /// </summary>
+        /// <param name="minDelay">The minimum duration value to use for each retry.</param>
+        /// <param name="maxDelay">The maximum duration value to use for each retry.</param>
+        /// <param name="fastFirst">Whether the first retry will be immediate or not.</param>
+        /// <param name="seed">An optional <see cref="Random"/> seed to use.
+        /// If not specified, will use a shared instance with a random seed.</param>
```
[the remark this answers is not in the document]
grant-d: Agree. Fixed

src/Polly.Shared/Duration/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,141 @@
...
+        #region Nested
+
+        /// <summary>
+        /// A random number generator with a Uniform distribution that is thread-safe (via locking).
+        /// Can be instantiated with a custom <see cref="Random"/> instance, for example to make
+        /// it act in a deterministic manner.
+        /// </summary>
+        private sealed class ConcurrentRandom
+        {
+            private static readonly Random s_random = new Random();
```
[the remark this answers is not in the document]
grant-d: Done

src/Polly.Shared/Duration/ISleepDurationStrategy.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// <summary>
+        /// Whether the first retry will be immediate or not.
+        /// </summary>
+        bool FastFirst { get; }
+
+        /// <summary>
+        /// Generate the sequence of <see cref="TimeSpan"/> values to use as sleep-durations.
+        /// </summary>
+        /// <param name="retryCount">The maximum number of retries to use, in addition to the original call.</param>
+        IEnumerable<TimeSpan> Generate(int retryCount);
```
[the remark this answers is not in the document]
grant-d: I like that. I had previously gone through several name changes, `Generate`, `Create`, `Build`, etc. Your suggestion is clean.

src/Polly.SharedSpecs/Duration/DecorrelatedJitterBackoffSpecs.cs
```diff
@@ -0,0 +1,231 @@
...
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public static void Should_have_no_variance_when_range_zero(bool fastFirst)
+        {
+            const int count = 1000;
+            TimeSpan minDelay = TimeSpan.FromMilliseconds(0);
+            TimeSpan maxDelay = TimeSpan.FromMilliseconds(0);
+
+            DecorrelatedJitterBackoff backoff = new DecorrelatedJitterBackoff(minDelay, maxDelay, fastFirst, Seed);
```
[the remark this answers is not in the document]
grant-d: Agree

grant-d:
@reisenberger, sorry I somehow missed some of the changes you requested last week. I have attended to all of them this morning. I think there may be a couple outstanding but I am having trouble navigating to find them, let me know.

src/Polly.Shared/Retry/ExponentialBackoff.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        /// <summary>
+        /// The duration value for the wait before the first retry.
+        /// </summary>
+        public TimeSpan InitialDelay { get; }
+
+        /// <summary>
+        /// The exponent to multiply each subsequent duration by.
+        /// </summary>
+        public double Factor { get; }
```
grant-d: Added an explicit factor, so that you are not limited to `x 2`

src/Polly.Shared/Retry/ExponentialBackoff.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// Creates a new instance of the class.
+        /// </summary>
+        /// <param name="initialDelay">The duration value for the wait before the first retry.</param>
+        /// <param name="factor">The exponent to multiply each subsequent duration by.</param>
+        /// <param name="fastFirst">Whether the first retry will be immediate or not.</param>
+        public ExponentialBackoff(TimeSpan initialDelay, double factor = 2.0, bool fastFirst = false)
+        {
+            if (initialDelay < TimeSpan.Zero) throw new ArgumentOutOfRangeException(nameof(initialDelay), initialDelay, "should be >= 0ms");
+            if (factor < 1.0) throw new ArgumentOutOfRangeException(nameof(factor), factor, "should be >= 1.0");
+            if (factor > 8.0) throw new ArgumentOutOfRangeException(nameof(factor), factor, "should be <= 8.0");
```
grant-d: Not sure about these two limits, what they should be. `8.0` is kind of arbitrary

src/Polly.Shared/Retry/ConcurrentRandom.cs
```diff
@@ -0,0 +1,63 @@
+using System;
+using System.Diagnostics;
+
+namespace Polly.Retry
+{
+    /// <summary>
+    /// A random number generator with a Uniform distribution that is thread-safe (via locking).
+    /// Can be instantiated with a custom <see cref="int"/> seed to make it act deterministically.
+    /// </summary>
+    internal sealed class ConcurrentRandom
```
grant-d: Move this into its own file.

reisenberger:
Thanks @grant-d ! ( Yes, just the params issue to come back to. )

grant-d:
> Yes, just the params issue to come back to

Here's an idea. What if I push a version of the code that has an 'all params' approach. We won't keep both versions, but then it's easier for us to compare & contrast.

src/Polly.Shared/Retry/Alternative.cs
```diff
@@ -0,0 +1,147 @@
+using System;
+using System.Collections.Generic;
+
+namespace Polly.Retry
+{
+    /// <summary>
+    /// </summary>
+    public static class Alternative // Experiment
```
grant-d: This is the alternative API shape. 
* We'd need a static class that holds the methods, called say 'SleepDurationSeriesStrategy`. 
* We would delete 'ISleepDurationSeriesStrategy` and all its concrete classes, since the `interface` would have no purpose (it would be a marker interface only)
* We would add the different sleep duration methods shown below.

Pros: Less cruft
Cons: Less encapsulation

(Please focus on api shape for now, not on comments, etc)

src/Polly.SharedSpecs/Retry/DecorrelatedJitterBackoffSpecs.cs
```diff
@@ -0,0 +1,229 @@
...
+            TimeSpan minDelay = TimeSpan.FromMilliseconds(10);
+            TimeSpan maxDelay = TimeSpan.FromMilliseconds(1500);
+
+            const int seed = 123456789;
+            DecorrelatedJitterBackoff backoff1 = new DecorrelatedJitterBackoff(minDelay, maxDelay, fastFirst, seed);
+            DecorrelatedJitterBackoff backoff2 = new DecorrelatedJitterBackoff(minDelay, maxDelay, fastFirst, seed);
+            IEnumerable<TimeSpan> discrete1 = backoff1.GetSleepDurations(count).ToList();
+            IEnumerable<TimeSpan> discrete2 = backoff2.GetSleepDurations(count).ToList();
+
+            // Experiment
```
grant-d: Here's the single unit updated to align with **experimental api shape**

src/Polly.Shared/Retry/Alternative.cs
```diff
@@ -0,0 +1,147 @@
...
+        /// </summary>
+        /// <param name="initialDelay">The duration value for the wait before the first retry.</param>
+        /// <param name="factor">The exponent to multiply each subsequent duration by.</param>
+        /// <param name="retryCount">The maximum number of retries to use, in addition to the original call.</param>
+        /// <param name="fastFirst">Whether the first retry will be immediate or not.</param>
+        public static IEnumerable<TimeSpan> ExponentialBackoff(TimeSpan initialDelay, int retryCount, double factor = 2.0, bool fastFirst = false)
+        {
+            if (initialDelay < TimeSpan.Zero) throw new ArgumentOutOfRangeException(nameof(initialDelay), initialDelay, "should be >= 0ms");
+            if (retryCount < 0) throw new ArgumentOutOfRangeException(nameof(retryCount), retryCount, "should be >= 0");
+            if (factor < 1.0) throw new ArgumentOutOfRangeException(nameof(factor), factor, "should be >= 1.0");
```
grant-d: Not sure about this limits, what it should be. Anything less than `1.0` makes the durations _shrink_ not grow. Perhaps that is a use-case?

grant-d:
> can be enumerated multiple times, giving different randomised results …

But I think this would only work in serial contexts. What would be our guidance when users try to use the 'same' strategy in a concurrent scenario.

reisenberger:
>>can be enumerated multiple times, giving different randomised results …

>I think the above would only work in serializable contexts

IIRC the details of `yield return` implementation, this will be fine whether different enumerations of the enumerable are executing serially or in parallel/concurrently. When the compiler sees a class returning `IEnumerable` which is implemented with `yield return`, it (IIRC) compiler-generates a hidden inner class to represent that method returning `IEnumerable`. Each _enumeration_ instantiates a separate instance of that compiler-generated inner-class. (This actually gives the thing you rightly identify is needed: "a distinct `enumerator` per [execution]".)  That's a quick explanation from memory; blogs should confirm.

So it all works thread-safely - provided, of course, that within our `IEnumerable<T>`-returning, `yield return`-implemented method, we don't use anything non-thread-safely.  Everything mutable within `DecorrelatedJitter.GetSleepDurations(...)` has method-execution-scope, and we take care that we use the only shared/external item, `_random`, thread-safely, so it looks good to me.

grant-d:
> this will be fine whether different enumerations of the enumerable are executing serially or in parallel/concurrently ...

You are right, I couldn't recall whether the hidden state machine was per-execution or not. I convinced myself that it works as follows:

```csharp
            const int count = 10;
            DecorrelatedJitterBackoff durationStrategy = new DecorrelatedJitterBackoff(TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(3));

            // Take an instance of IEnumerable<TimeSpan> as directly returned by DecorrelatedJitterBackoff.Generate(...), exactly as users should when configuring a policy.
            IEnumerable<TimeSpan> generate = durationStrategy.GetSleepDurations(count);

            var tasks = new Task<TimeSpan[]>[100];
            var start = new ManualResetEvent(false);

            for (var i = 0; i < tasks.Length; i++)
            {
                var tcs = new TaskCompletionSource<TimeSpan[]>();
                tasks[i] = tcs.Task;

                var icopy = i;
                var thread = new Thread(() =>
                {
                    start.WaitOne();
                    tcs.SetResult(generate.ToArray()); // Every thread enumerates the same shared singleton
                });
                thread.Start();
            }

            start.Set(); // Unleash the fury
            Task.WaitAll(tasks);

            tasks.Should().NotContain(n => n.IsFaulted); // All passed
```

src/Polly.Shared/Retry/Alternative.cs
reisenberger: @grant-d Great to see 👍 .  I was thinking that all the configuration parameters to generate the `IEnumerable<TimeSpan>` should be configured together (not some in the ctor, some in a method), which essentially gives a choice between forms like:

(1) `new DecorellatedJitter(param1, param2, retryCount).GetSleepDurations()` // pre-configured instance with a no-params get method
(2) `DecorellatedJitter.GetSleepDurations(param1, param2, retryCount)` // static factory method on static class-per-strategy  .  <sub>Same as your `SleepDurationSeriesStrategy` factory class, except not overloading all variants on a single factory class</sub>

Community views on advs / disadvs either way welcome.  

Polly configuration has tended to take a static-factory-method declarative approach like (2), so my initial instinct lies here.  If a given set of parameters isn't re-used, (2) also reads more concisely than (1) with its extra `new`.  But an instance approach (1) could bind better with the .NET Core `IOptions<>` approach to config binding.  

A disadvantage of the instance approach (1) could be if it tempts / inadvertently guides users to a singleton `DecorrelatedJitter` instance (or single `ISleepDurationSeriesStrategy` fulfilment instance) to use all around their app. Singleton configuration can be a good starting point but isn't always a good fit for real world infrastructure - it's common to have a mix of more reliable/nearby (in network terms) resources like eg other Azure resources, versus more fragile eg external legacy systems, which might need different strategies.  And the absence of named-registrations in classic Net Core DI makes multiple variant registrations of the same class/interface hard.  Then again, this could be covered by documentation.  And a _recommended_ / more typical Polly pattern is to configure all policies at startup, so it would be less usual (tho possible) to be DI-injecting `DecorellatedJitter` instances into point-of-use classes, to configure a policy there.  

Community views?

reisenberger:
@grant-d @george-polevoy . [Similar to for the current fault-injection work](https://github.com/App-vNext/Polly/pull/508#issuecomment-443497304), I am going to propose that we promote this contribution to its own package, in this case (something like) `Polly.Contrib.WaitAndRetryStrategies`.  Advantages:

+ More freedom to rev `Polly.Contrib.WaitAndRetryStrategies` independently from (not blocked by) the Polly rev cycle - grow/refine more freely as enhancements to the algorithms occur to us/anyone
+ We can include more of @george-polevoy 's algorithm variants (if George is happy to)
+ We can give you greater control over the `Polly.Contrib` package (if you wish); and headlining as contributors (either way)
+ More space to document the math in depth 

Please let me know any thoughts.

Mechanics:

+ I would need time to set up the separate package.
+ We would likely first merge, then-demerge, this PR on Polly in any case, to ensure you get the due [contributor credit](https://github.com/App-vNext/Polly/graphs/contributors)
+ We would advertise the package directly from the Polly retry wiki page (and everywhere else relevant)

reisenberger:
@grant-d The [separate package proposal](https://github.com/App-vNext/Polly/pull/536#issuecomment-446862770) (if agreed) helps guide some of our decisions around [interfaces-or-not; and instances-of-a-class configuration versus static factory methods](https://github.com/App-vNext/Polly/pull/536#discussion_r238102552).  Propose we may go for:

(a) static factory methods on a separate-class-per-strategy, ie a syntax `DecorrelatedJitter.GetSleepDurations(param1, param2, retryCount)` , `LinearBackoff.GetSleepDurations(....)`, etc.  

+ static class-per-strategy grows more easily than one class holding all variants

(b) static methods, no instances, no interface

+ `IEnumerable<TimeSpan>` seems a clear and well enough defined interface between `Polly.Contrib.WaitAndRetryStrategies` and `Polly`

+ Less cruft, [as you say](https://github.com/App-vNext/Polly/pull/536#discussion_r236733960).  Similarly, I have tended to avoid introducing configuration instance classes to Polly, [despite requests](https://github.com/App-vNext/Polly/issues/143), to avoid it becoming extra surface to maintain, constraining the API (adding options breaks the classes), and it is [easy enough for Polly users to do themselves in the particular way they want anyway](https://github.com/App-vNext/Polly/wiki/Polly-Roadmap#consider-introducing-configuration-provider-interfaces-andor-pocos).

Again, let me know if this path forward seems ok.

grant-d:
Sorry, been snowed under at work, will try get to this this week

grant-d:
Regarding `core` vs `contrib`. I don't mind either way except for the risk that other `contribs` I have used in the past have become ill-maintained and/or runaway trains, filled with everyone's favorite idea. As long as we curate both repos strictly, I think we should be fine.

src/Polly.Shared/Retry/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// For example: 117ms, 236ms, 141ms, 424ms, ...
+        /// For background, see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/.
+        /// </summary>
+        /// <param name="minDelay">The minimum duration value to use for the wait before each retry.</param>
+        /// <param name="maxDelay">The maximum duration value to use for the wait before each retry.</param>
+        /// <param name="retryCount">The maximum number of retries to use, in addition to the original call.</param>
+        /// <param name="fastFirst">Whether the first retry will be immediate or not.</param>
+        /// <param name="seed">An optional <see cref="Random"/> seed to use.
+        /// If not specified, will use a shared instance with a random seed, per Microsoft recommendation for maximum randomness.</param>
+        public static IEnumerable<TimeSpan> Create(TimeSpan minDelay, TimeSpan maxDelay, int retryCount, bool fastFirst = false, int? seed = null)
```
grant-d: Here is the new factory method.

src/Polly.SharedSpecs/Retry/DecorrelatedJitterBackoffSpecs.cs
```diff
@@ -0,0 +1,270 @@
...
+            DecorrelatedJitterBackoff backoff2 = new DecorrelatedJitterBackoff(minDelay, maxDelay, fastFirst, seed);
+            IEnumerable<TimeSpan> discrete1 = backoff1.GetSleepDurations(count).ToList();
+            IEnumerable<TimeSpan> discrete2 = backoff2.GetSleepDurations(count).ToList();
+
+            discrete1.Should().HaveCount(count);
+            discrete2.Should().HaveCount(count);
+            discrete1.Should().ContainInOrder(discrete2);
+
+            // Factory-based instantiation
+            IEnumerable<TimeSpan> discrete3 = DecorrelatedJitterBackoff.Create(minDelay, maxDelay, count, fastFirst, seed).ToList();
```
grant-d: Sample callsite for factory method

src/Polly.Shared/Retry/DecorrelatedJitterBackoff_static.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Collections.Generic;
+
+namespace Polly.Retry
+{
+    /// <summary>
+    /// Generates sleep durations in an jittered manner, making sure to mitigate any correlations.
+    /// For example: 117ms, 236ms, 141ms, 424ms, ...
+    /// For background, see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/.
+    /// </summary>
+    public static class DecorrelatedJitterBackoff_static // Temporary name
```
grant-d: Here's the option where we have static methods returning `IEnumerable` only

src/Polly.SharedSpecs/Retry/DecorrelatedJitterBackoffSpecs.cs
```diff
@@ -0,0 +1,274 @@
...
+            discrete3.Should().HaveCount(count);
+
+            // Static-based instantiation
+            IEnumerable<TimeSpan> discrete4 = DecorrelatedJitterBackoff_static.Create(minDelay, maxDelay, count, fastFirst, seed).ToList();
```
grant-d: Sample callsite for static method

src/Polly.Shared/Retry/DecorrelatedJitterBackoff.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            var backoff = new DecorrelatedJitterBackoff(minDelay, maxDelay, fastFirst, seed);
+            return backoff.GetSleepDurations(retryCount);
+        }
```
grant-d: Here is the new factory method.

grant-d: Here is the new factory method

src/Polly.SharedSpecs/Retry/DecorrelatedJitterBackoffSpecs.cs
```diff
@@ -0,0 +1,274 @@
...
+            discrete4.Should().HaveCount(count);
```
grant-d: Sample callsite for static instantiation

src/Polly.SharedSpecs/Retry/DecorrelatedJitterBackoffSpecs.cs
```diff
@@ -0,0 +1,274 @@
...
+            IEnumerable<TimeSpan> discrete1 = backoff1.GetSleepDurations(count).ToList();
+            IEnumerable<TimeSpan> discrete2 = backoff2.GetSleepDurations(count).ToList();
+
+            discrete1.Should().HaveCount(count);
+            discrete2.Should().HaveCount(count);
+            discrete1.Should().ContainInOrder(discrete2);
+
+            // Factory-based instantiation
+            IEnumerable<TimeSpan> discrete3 = DecorrelatedJitterBackoff.Create(minDelay, maxDelay, count, fastFirst, seed).ToList();
+            discrete3.Should().HaveCount(count);
```
grant-d: Sample callsite for factory instantiation

src/Polly.Shared/Retry/DecorrelatedJitterBackoff.cs
grant-d: Here is the new factory method

reisenberger:
Thanks @grant-d .  I [disposed of a question around the new jitter strategy](https://github.com/App-vNext/Polly/issues/530#issuecomment-450553899) also over the festive season, probably going to be a bit more kanban over next couple of weeks (_courtesy update_) to focus purely on getting Polly v7.0 and Simmy out, then let's bring this to fruition also. Thanks again!

reisenberger:
This PR was unexpectedly automatically closed by github, when deleting a (now-outdated) base branch which the PR targeted .  

@grant-d Thank you for everything you have done on this!  Now that Polly v7.0.0 is released, I hope to revert shortly to how we may take this forward.

grant-d:
@reisenberger any news on the plans here

reisenberger:
@grant-d , I hope to come back to this at the weekend.

grant-d:
Cool. My personal feeling is that class called `Alternative` is the best approach. I have used such a pattern with Polly in my own code recently, and it's clean.

Issue #551, opened 2018-12-29
Extensibility: Enable custom policies - support Polly.Contrib, chaos engineering, fakes ... and anything ...


reisenberger:
A long-term goal for Polly has been to provide an easy extensibility mechanism for custom policies.  

#### (1) Support the Polly team and associates to provide new policy functionality in packages separate from Polly

_Example:_  Chaos Engineering and fault-faking, #508.  This will [have clearer identity as a separate package](https://github.com/App-vNext/Polly/pull/508#issuecomment-443497304), yet needs to integrate tightly with Polly.

#### (2) Open the possibility for new policies to form contributions via a Polly.Contrib

With Polly in a project, [`PolicyWrap`](https://github.com/App-vNext/Polly/wiki/PolicyWrap) can morph towards an execution-despatch middleware for _any_ key operations in your app (rather as [`DelegatingHandlers` are for `HttpClient`](https://www.stevejgordon.co.uk/httpclientfactory-aspnetcore-outgoing-request-middleware-pipeline-delegatinghandlers), but for any execution).

It can be useful then to add extra functionality around the executions - extra logging, capture timing information, attach correlation ids, etc.  Some of these concerns are off the core focus of Polly (resilience), but nevertheless can be valuable to integrate into execution via PolicyWrap.

_Example_: There are periodically requests for a policy to [log exceptions and rethrow](https://stackoverflow.com/questions/42952057/polly-policy-to-log-exception-and-rethrow). 
_Example_: It would be trivial to write a policy capturing execution duration.

#### (3) Allow users to easily develop and integrate custom policies for any purpose

---

### Requirements

Custom policies should integrate easily into Polly's existing configuration syntax and execution despatch:

(a) integrate into [`PolicyWrap`](https://github.com/App-vNext/Polly/wiki/PolicyWrap)
(b) support other execution-scoped information such as [execution keys](https://github.com/App-vNext/Polly/wiki/Keys-and-Context-Data), the [exchange of information via execution-scoped `Context`](http://www.thepollyproject.org/2017/05/04/putting-the-context-into-polly/), and (potentially later) metrics
(c) integrate with the in-built co-operative cancellation in PolicyWrap
(d) be able to take advantage of Polly's existing [`.Handle<>()` syntax](https://github.com/App-vNext/Polly#step-1--specify-the--exceptionsfaults-you-want-the-policy-to-handle) for reactive policies
[...]

Issue #550, opened 2018-12-29
Restore Exception binary serialization for .Net Standard 2.0 offering


reisenberger:
.NET Core 2.0 now [supports binary serialization for `Exception` again](https://docs.microsoft.com/en-us/dotnet/standard/serialization/binary-serialization#binary-serialization-in-net-core), although this is more for general backward compatibility of APIs from .NET Framework than for [remoting and AppDomains, which are deprecated technologies](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/porting.md#unsupported-technologies).

This issue proposes that (for completeness) we reintroduce the `[SerializableAttribute]` on exceptions thrown by Polly, for the .Net Standard 2.0 target only.

Anyone with good insight to reasons to do / _not_ to do this - please do comment.

Issue #281, opened 2017-07-25
Cleanly formalise the separation of sync and async policies (was: option to unify them)


reisenberger:
**TL;DR At the expense of some syntax changes** (which would eventually need to be pushed through as breaking changes), **we could unify sync and async policies in Polly.  What do users think?**   

A new syntax such as below would be a breaking change, but would allow sync and async policies to be unified:

```c#
.WaitAndRetry(int retryCount, Func<int, Context, TimeSpan> sleepDurationProvider)
    .OnRetry(Action<Exception, TimeSpan, int, Context> onRetry) // optional postfix configuration step
    .OnRetryAsync(Func<Exception, TimeSpan, int, Context, Task> onRetryAsync) // optional postfix configuration step
``` 

---

## Background: Current sync/async split 

With Polly we currently must use separate Polly policies for synchronous and asynchronous executions:  

+ When you define a policy, you define it as either (eg) `Retry()` or `RetryAsync()`
+ You can only `Execute()` (etc) on synchronous policies; `ExecuteAsync()` on async policies; but [not use one policy instance for both sync and async executions](https://github.com/App-vNext/Polly/blob/cfef09907d45cb82846e03e6a11b6a4f6d51abfe/src/Polly.Shared/PolicyAsync.cs#L1126-L1127).

That architecture evolved due to historic decisions (before current maintainers), and [this blog post](http://www.thepollyproject.org/2017/06/09/polly-and-synchronous-versus-asynchronous-policies/) explored some of the reasons and blocks to changing this - to unifying sync and async policies.  

Decisive is that users expect async policies to have async state-change delegates, while sync policies must have sync ones.  So a policy that would act for both sync and async executions, must at least _allow_ the definition of all state-change delegates in both sync and async forms.

## The current syntax - a block to change? 

A block to defining policies with both/either sync and async state-change delegates is Polly's current syntax.  Take one of the fuller `WaitAndRetry()` configuration overloads:

[`.WaitAndRetry(int retryCount, Func<int, Context, TimeSpan> sleepDurationProvider, Action<Exception, TimeSpan, int, Context> onRetry)`](
https://github.com/App-vNext/Polly/blob/cfef09907d45cb82846e03e6a11b6a4f6d51abfe/src/Polly.Shared/Retry/RetrySyntax.cs#L326).  

Add the async state-change delegate, and you have the slightly clumsy: 

[...]

PR #552, opened 2018-12-29
Set baseline for v7.0.0: custom policies; confirm and clean sync-async split

### LINKED ISSUE #546: (very edge case) Context passed to onBreak or onReset when manually break or reset the circuit, should not be shared/mutable
**Summary**: 

_An exceptionally edge-case issue unlikely ever to have been hit, but can be tided away._

Circuit-breaker state-transition delegates `onBreak` and `onReset` take a `Context` parameter, passing to the delegate the context of the execution when the circuit broke or reset.

When the circuit is broken _manually_ with the `CircuitBreaker.Isolate(...)` method, or reset manually with `CircuitBreaker.Reset(...)`, those are manual changes outside of an individual policy execution, and so the `onBreak` and `onReset` delegates invoked were passed an empty `Context`.

**Expected behavior**: 

When  `CircuitBreaker.Isolate(...)` method or `CircuitBreaker.Reset(...)` is invoked, the `Context` passed to `onBreak` and `onReset` should always be empty.

**Actual behaviour**:

The `Context` instance passed to the methods in those cases was in fact long-lived and shared between invocations of those methods. The `IDictionary` methods on `Context` could potentially have been used to add data to that long-lived, intended-to-be-empty instance of `Context`; which was never the intention.

reisenberger:

+ Implements #551 Extensibility for custom policies and external policies such as fakes and chaos policies, and wider Polly.Contrib
+ Implements #281 Confirm and formalise the sync-async split in policies

Also:

+ Implements #550 Restore binary serialization for exceptions, for .Net Standard 2.0
+ Fixes #546 edge case around `Context` mutability

### Confirm the following

- [x]  I have merged the latest changes from the dev vX.Y branch


mebjas:
@reisenberger wow this seem like a lot of work. Kudus to you for this :)

PR #554, opened 2018-12-31
Make the IReadOnlyPolicyRegistry interface inherit IEnumerable

### LINKED ISSUE #520: Inherit IReadOnlyPolicyRegistry from IEnumerable<TKey, IsPolicy>
It would be really nice if IReadOnlyPolicyRegistry could implement IEnumerable<TKey, IsPolicy>. This would allow one to iterate through the configured policies and also allow the use of collection initializers when constructing a writable Policy registry.
I'm happy to do this if it would be warranted, but my only question would be what tests would be warranted?

SeanFarrow:
Fixes #520 to enable the use of IEnumerable<KeyValuePair<TKey, IsPolicy>> in the IReadOnlyPolicyRegistry interface.

This also allows the .Net collection initializer syntax to be used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 ## 7.0.0
 - Clarify separation of sync and async policies
 - Enable extensibility by custom policies hosted external to Polly
-     
+- Enable collection initialization syntax for PolicyRegistry
+
 ## 6.1.2
 - Bug Fix: Async continuation control for async executions (issue 540, affected only v6.1.1)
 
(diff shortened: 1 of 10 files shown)


reisenberger:
Hi @SeanFarrow . Looks great to me!  Going to make a tiny test change, then just push some some documentation (and credit to you) onto the PR, then I think it should be good to merge.  Probably leave it a day to merge in case you want to second-read the changes I pushed.  Thanks, as ever, for your contributions.

Issue #674, opened 2019-08-08
AsyncRetryPolicy does not use ContinueOnCapturedContext configuration


simluk:
In our MVC web application there is breaking issue with HTTP context being lost when AsyncRetryPolicy is used. Polly is using ConfigureAwait(false) on retry delegate where we need to have ConfigureAwait(true) on each retry call to avoid losing the context. 
<br/><br/>

In AsyncRetryPolicy.ImplementationAsync you need to forward continueOnCapturedContext to AsyncRetryEngine.ImplementationAsync.  

All other Policies seem to behave correctly in this matter.

Please let me know if this was done intentionally or you just missed it out.

reisenberger:
Thank you for identifying this. It looks like the correct behavior (passing continueOnCapturedContext) may have been lost in a refactor. Please feel free to make a PR against the vXXX branch, to fix.

simluk:
Thanks for merging this in ! :)

reisenberger:
@simluk Thanks for both raising the bug and contributing the fix. Now released to nuget in [Polly 7.1.1](https://www.nuget.org/packages/Polly/7.1.1).

PR #675, opened 2019-08-12
BUG-674: AsyncRetryPolicy does not use ContinueOnCapturedContext conf…

simluk:
Trivial bug-fix to keep context with ConfigureAwait between retries in AsyncRetryEngine.ImplementationAsync:
https://github.com/App-vNext/Polly/issues/674

### DIFF
--- a/src/Polly/Retry/AsyncRetryPolicy.cs
+++ b/src/Polly/Retry/AsyncRetryPolicy.cs
@@ -47,7 +47,8 @@ namespace Polly.Retry
                 _sleepDurationsEnumerable,
                 _sleepDurationProvider != null 
                     ? (retryCount, outcome, ctx) =>  _sleepDurationProvider(retryCount, outcome.Exception, ctx) 
-                    : (Func<int, DelegateResult<TResult>, Context, TimeSpan>)null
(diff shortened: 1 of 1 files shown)


reisenberger:
Thank you for this. We hope to prepare and publish a v7.1.1 incorporating this on or around 19 August.

reisenberger:
@simluk Thank you for your patience . I was on vacation but made sure to reply rather than leave you with no replies. Now returned from vacation - aiming to cut a Polly release soon, but it could be a few days (depending on other priorities). Thanks.

simluk:
> 
> 
> @simluk Thank you for your patience . I was on vacation but made sure to reply rather than leave you with no replies. Now returned from vacation - aiming to cut a Polly release soon, but it could be a few days (depending on other priorities). Thanks.

No worries, thanks for looking into it :)

reisenberger:
I would hope that we can make a release (including this) by Monday 02 September.

reisenberger:
This build (with minimal code changes) is good locally - not clear why AppVeyor has started raising errors on a process it didn't before, but will investigate that separately.  Merging to v711-or-v720 (preceding later merge to master).

PR #681, opened 2019-08-22
Include symbols in package

kzu:
Given that the new portable PDBs generated for NS projects are really small for the added benefit of 
useful stacktraces on errors, it would complement the sourcelink that's already in place. 

In particular, for VS extensions that use this package, including symbols in the package makes it easier 
for crash dumps to contain useful stacktraces. i.e. see https://github.com/jeffkl/ManagedDism/pull/76

This would fix bug http://work.azdo.io/971173 too :)

### Confirm the following

- [x]  I started this PR by branching from the head of the latest dev vX.Y branch, or I have rebased on the latest dev vX.Y branch, or I have merged the latest changes from the dev vX.Y branch

### DIFF
--- a/src/Polly/Polly.csproj
+++ b/src/Polly/Polly.csproj
@@ -18,6 +18,7 @@
     <Authors>Michael Wolfenden, App vNext</Authors>
     <AssemblyOriginatorKeyFile>..\Polly.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
   </PropertyGroup>
   <PropertyGroup Label="SourceLink">
     <PublishRepositoryUrl>true</PublishRepositoryUrl>


kzu:
CI seems broken for completely unrelated reasons :/

reisenberger:
Thanks for the PR! _Community views?_  

Feels like there could be mixture of benefits/drawbacks for different use cases.  Improved stack traces on one hand; increased package size otoh for those where that could be a concern (eg Xamarin and serverless consumers).  Planning to consult colleagues / others in OSS community before taking definitive view.

Also, looks like our last `.snupkg` never got published at https://www.nuget.org/packages/Polly/ even tho our build generates it 👇 . Would like to publish that and see if it helps - wld be awesome if we could fix your issue w/o adding weight to the nupkg.

image

Aware of the relevant here [(sourcelink)](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink); [(symbol packages)](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/nuget#symbol-packages).  

Further community views?

---

Re build issue, the PR should target the latest develop branch (I changed it), already fixes the build issue you encountered against `master`.

Thanks for all the awesome you've done around Moq 🙂

reisenberger:
@kzu We published a [version of Polly now with the `.snupkg` available via nuget](https://www.nuget.org/packages/polly). Does this make any difference for your use case? Not full of hope on that - but would be good to know.

reisenberger:
Reviewed this again. Electing to merge this and re-add the `pdb`s to the nupkg. 

We did in fact include the `pdb`s up to v7.0.3 without major issues being raised by the community. Including them gives Polly consumers more flexibility than excluding them. Users who want to exclude them from final builds can do so by configuring their release builds appropriately.

PR #690, opened 2019-10-02
Migrate Fluent Assertions from v4 to v5

jnyrup:

I stumbled upon a comment in the tests about `Func<T>` not being supported in Fluent Assertions... but we added that in version [5.5.1](https://github.com/fluentassertions/fluentassertions/releases/tag/5.5.1) about a year ago :).

I've split the PR into several commits, which should make it easier to review, but as the commits aren't atomic, they should probably be squashed before merging.

Some behavioral changes in Fluent Assertions: 
* In `BeEquivalentTo` the expectation now drives the comparison. So for all the tests where a subject is compared against an anonymous class, only the members of the anonymous class are used in the comparison.
* Fluent Assertions returns the proper exception from async methods, so when `ThrowIfCancellationRequested` throws an `OperationCanceledException`, the assertions can assert on that type instead of `TaskCanceledException`.
* `WithInnerException` returns the inner exception instead of the outer exception.

Other changes:
* When asserting on the returned policies from `GetPolices` using `BeEquivalentTo`, I've changed the comparison of elements to use reference equality.
* `Guid` has native support for comparison, so they can be compared using `Be` instead of `BeEquivalentTo`.
*

An overview about changes in version 5 are described [here](https://www.continuousimprover.com/2018/02/fluent-assertions-50-best-unit-test.html).
### Confirm the following

- [x]  I started this PR by branching from the head of the latest dev vX.Y branch, or I have rebased on the latest dev vX.Y branch, or I have merged the latest changes from the dev vX.Y branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -1058,6 +1058,7 @@ Both templates contain a full project structure referencing Polly, Polly's defau
 * [@aprooks](https://github.com/aprooks) - Build script tweaks for Mac and mono.
 * [@kesmy](https://github.com/Kesmy) - Add Soucelink support, clean up cake build.
 * [@simluk](https://github.com/simluk) - Fix continueOnCaptureContext not being honored in async retry implementation (bug in v7.1.0 only).
+* [@jnyrup](https://github.com/jnyrup) - Upgrade tests to Fluent Assertions v5.9.0
 
 # Sample Projects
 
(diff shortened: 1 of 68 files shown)


reisenberger:
@jnyrup Thank you for taking the time to make this detailed set of changes. Although I was aware of some of the updates in FluentAssertions v5 from work in other projects, I had not had time to update this in Polly. 

Thank you also for the clear, separate commits and detailed cover note, which made reviewing the volume of changes much easier. :+1:

I'll be merging this, after pushing a credit to you for the contribution! 

Thanks for everything the FluentAssertions teams does 💯  ; we're a great fan of the library.

PR #699, opened 2019-10-09
remove some redundant generic type specs

SimonCropp:
this is my preferences. please close if u prefer it to be explicit


SimonCropp:
conflicts resolved

reisenberger:
Thanks for this! Having reviewed, I'm happy to _make_ all the changes in the files ticked in the screen-grab below, plus in `PolicyWrapContextAndKeySpecs.cs` 

Issue699HappyToChange

All the others I prefer to leave as are, for human readability. _Explanation:_ Polly has lots of similar method overloads, and types, in both non-generic and generic forms, eg `.Execute()` and `.Execute<TResult>()`.  Although the compiler can often infer the generic type parameter so we could remove it, that creates possible confusion for _human_ readers given the existence of close-lookalike generic and non-generic forms.

---

Feel free to clean up the PR along these lines, if you want to.

PR #700, opened 2019-10-09
dont ignore AsyncBulkheadPolicy _maxParallelization

SimonCropp:
this is a guess. not sure if it is a bug?

### DIFF
--- a/src/Polly/Bulkhead/AsyncBulkheadPolicy.cs
+++ b/src/Polly/Bulkhead/AsyncBulkheadPolicy.cs
@@ -33,7 +33,7 @@ namespace Polly.Bulkhead
         /// <summary>
         /// Gets the number of slots currently available for executing actions through the bulkhead.
         /// </summary>
-        public int BulkheadAvailableCount => _maxParallelizationSemaphore.CurrentCount;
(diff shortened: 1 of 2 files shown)


reisenberger:
Thanks for raising this along with all the other contributions.  

I don't think this change is needed. The `CurrentCount` in these semaphores cannot exceed `_maxParallelization`, because the semaphores (internal to Polly) are created with maximum capacity `_maxParallelization` ([example](https://github.com/App-vNext/Polly/blob/4f1e0d2c583c54ab549da2375567a6924948d46f/src/Polly/Bulkhead/BulkheadSyntax.cs#L66)).  A `SemaphoreFullException` would be thrown before `CurrentCount` could exceed `_maxParallelization`.

Let me know if you think I'm missing something; otherwise I'll close this shortly.

SimonCropp:
in that case dont pass in maxParallelization?

reisenberger:
Hmm. A more interesting question (looking at the code) is why the `SemaphoreSlim` are being passed in from outside rather than created in the policy constructor. I believe that's a remnant from an earlier, different structure to the code. 

Feel free to amend the PR (if you agree) such that the `SemaphoreSlim`s are created within the policies. (That will also make use of the `maxParallelization` variable.)

Thanks for raising this!

reisenberger:
Replaced by #716

PR #706, opened 2019-10-09
remove some redundant casts

SimonCropp:


SimonCropp:
conflicts resolved

src/Polly.Specs/PolicyAsyncSpecs.cs
```diff
@@ -194,7 +194,7 @@ public void Executing_the_policy_action_should_throw_when_context_is_null()
                 .Handle<DivideByZeroException>()
                 .RetryAsync((_, __, ___) => { });
 
-            policy.Awaiting(async p => await p.ExecuteAsync(ctx => TaskHelper.EmptyTask, (Context)null))
+            policy.Awaiting(async p => await p.ExecuteAsync(ctx => TaskHelper.EmptyTask, null))
```
reisenberger: I prefer to keep this explicit cast `(Context)null` (and other similar ones).

Without it, to understand at a glance why the compiler automatically chooses that overload without ambiguity, the reader would have to have in mind [point 6 of the overload resolution rules](https://docs.microsoft.com/en-us/dotnet/visual-basic/reference/language-specification/overload-resolution), as well as knowing that `Context : IDictionary<string, object>` (which is what happens to make one Execute overload more specific than the other). 

That's not going to be nearly as grokkable for a new reader to the code, than having the explicit cast in place. IE I'm opting to prioritise human readability over compiler readability in this case.

src/Polly.Specs/Wrap/PolicyWrapSpecsAsync.cs
```diff
@@ -28,7 +28,7 @@ public void Nongeneric_wraps_generic_instance_syntax_wrapping_null_should_throw(
         {
             var retry = Policy.Handle<Exception>().RetryAsync(1);
 
-            Action config = () => retry.WrapAsync<int>((AsyncPolicy<int>)null);
+            Action config = () => retry.WrapAsync<int>(null);
```
reisenberger: These `.WrapAsync()` methods exist in multiple forms: allowing generic-policies to wrap either non-generic or generic policies; and likewise non-generic policies to wrap either non-generic or generic policies. This is part of the flexibility of Polly `PolicyWrap`, so 'stuff just works'.

Again, tooling will suggest (absolutely correctly, from a technical perspective) that generic parameters or casts can be removed, but there is a lot of room for _human reader_ ambiguity (or having to puzzle out the meaning) once those are removed, given the number of similar overloads Polly offers.

So again: I'm preferring to retain the explicit casts to prioritise human readability.

SimonCropp:
no worries at all

Issue #818, opened 2021-01-13
When using HandleInner() not all AggregateExceptions are being checked


sideproject:
When handling exceptions that that are wrapped in an AggregateException, it looks like only the first exception is recursively searched for a matching predicate.  If there are two (or more) exceptions in the InnerExceptions collection, they are not searched.

```csharp
//Example Policy
var policy = Policy
    .HandleInner<DivideByZeroException>()
    .Retry(3, (exception, _) => passedToOnRetry = exception);
```
```csharp
//The policy will find and retry the nested DivideByZeroException
Exception willRetryException = new AggregateException(
    new Exception("First: With Inner Exception"),
        new DivideByZeroException()));
    new Exception("Second: Without Inner Exception",
```
```csharp
//The policy will NOT find and retry the nested DivideByZeroException
Exception willNotRetryException = new AggregateException(
    new Exception("First: Without Inner Exception"),
    new Exception("Second: With Inner Exception",
        new DivideByZeroException()));
```

I was able to make this work by modifying the HandleInner function in  src/Polly/PolicyBuilder.OrSyntax.cs:55.

```csharp
internal static ExceptionPredicate HandleInner(Func<Exception, bool> predicate)
{
    return exception =>
    {
        if (exception is AggregateException aggregateException)
        {
            foreach (var innerException in aggregateException.Flatten().InnerExceptions)
            {
                var matchedInAggregate = HandleInnerNested(predicate, innerException);
                if (matchedInAggregate != null)
                    return matchedInAggregate;
            }
        }

        return HandleInnerNested(predicate, exception);
    };
}
```

I'm happy to submit a pull request with tests for this if you're willing to entertain it.

sideproject:
Just checking in - is this change something that you'd entertain?  I'm happy to submit a pull request for discussion.

martincostello:
This _seems_ reasonable to me, but I'm not sure whether or not it could be considered a breaking behavioural change or not.

I guess if it's a minor enough code change to make you could open a PR for it and solicit further comment/opinion.

sideproject:
Understood.  I've submitted the PR for your review.

sideproject:
Understood, I've created PR #822 for your review.

The CI appears to have failed but doesn't seem to be related to my change - `Could not load file or assembly 'YamlDotNet,`

(Apologies for the second PR, I had some trouble with the first one so I started over)

martincostello:
Fixed in v7.2.2.

PR #822, opened 2021-01-23
Recursively search all AggregateException InnerExceptions for predicate matches when using HandleInner()

sideproject:
Recursively search all AggregateException InnerExceptions for predicate matches when using HandleInner()

Related Issue: #818

(Apologies for the second PR, I had some trouble with the first one so I started over)

### DIFF
--- a/src/Polly.Specs/Retry/RetrySpecs.cs
+++ b/src/Polly.Specs/Retry/RetrySpecs.cs
@@ -294,6 +294,52 @@ namespace Polly.Specs.Retry
         }
 
         [Fact]
+        public void Should_call_onretry_with_handled_exception_nested_in_aggregate_inside_another_aggregate_as_first_exception()
+        {
+            Exception passedToOnRetry = null;
+
+            var policy = Policy
+                .HandleInner<DivideByZeroException>()
+                .Retry(3, (exception, _) => passedToOnRetry = exception);
+
+            Exception toRaiseAsInner = new DivideByZeroException();
+
+            Exception aggregateException = new AggregateException(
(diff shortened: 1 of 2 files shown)


martincostello (approved):
Looks good to me. I wonder if it's worth also adding a further test case for where an `AggregateException` contains an `AggregateException` which _then_ contains the exception being searched for?

sideproject:
Good idea.  I've added two more tests to verify that nested AggregateExceptions are also searched.

martincostello:
I'm not sure why the CI keeps failing, but it's missing some different changes I made to fix the CI in the `v722-or-v730` branch this weekend. Could you rebase your branch with the latest changes to rule that out please?

Otherwise, this looks good to merge.

sideproject:
I just rebased from upstream and pushed - did it help?

martincostello:
Apparently not 😕

My only other guess is that the CI auto-versioning stuff doesn't like the name of your branch being the same name as the target branch as a PR I opened yesterday branched from the same point is fine: #823.

sideproject:
Would you like me to submit a new PR with a different branch name?

martincostello:
We could try to rule it out. If it does fix it then it'll be something to do with an old issue logged against the CI being flakey. If it doesn't then I'm a bit stumped.

Leave this one open for now and try opening another for a different branch name - if it works, then we can swap over to that PR, otherwise you can close it and we'll stick with this one.

martincostello:
Yep, looks like the branch name was what is breaking it for some reason.

Will close this PR in favour of #824.

PR #865, opened 2021-06-07
update to c# 9

SimonCropp:


src/Directory.Build.props
```diff
@@ -3,5 +3,6 @@
   <PropertyGroup>
     <AssemblyOriginatorKeyFile>..\Polly.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
+    <LangVersion>9.0</LangVersion>
```
martincostello, line 6: Could just set to `latest`? Then we don't have to update it every release and it will just use whatever is stable in the SDK version in use.

SimonCropp: done

martincostello (approved):

PR #872, opened 2021-07-22
Modify CircuitStateController to public, allow user use it.

pengweiqhca:


martincostello:
These types are internal implementation details. Exposing them as-is would make it harder to refactor/iterate the implementation of Polly without making binary and/or source breaking changes to the public API surface.

What problem exactly are you having that having full public access to these types would solve?

pengweiqhca:
I want use ICircuitStateController to control my own CircuitBreaker logic rather than CircuitBreakerPolicy.

I can use reflection, but AOT not support reflection and I also think use reflection is very ugly.

PR #923, opened 2022-02-14
Fixed possible NRE.

FoxTes:
Corrected possible NRE.

Now, when using var `sleepDurationsEnumerator `, it is first checked that it is not null.


martincostello:
Thanks!

PR #925, opened 2022-02-14
Exception classes should be publicly accessible.

FoxTes:
Fixed exception class inaccessible to other builds. If such an exception is thrown, the external code will have to catch instances of the nearest accessible parent class such as the base class of all exceptions, 'Exception'. This hinders exception handling since the code of other builds cannot identify the problem precisely.

### DIFF
--- a/src/Polly/Utilities/TimedLock.cs
+++ b/src/Polly/Utilities/TimedLock.cs
@@ -86,9 +86,12 @@ namespace Polly.Utilities
 
     }
 
-    internal class LockTimeoutException : Exception
+    /// <summary>
+    /// Exception caused by if a lock cannot be obtained within the Lock Timeout.
+    /// </summary>
+    public class LockTimeoutException : Exception
     {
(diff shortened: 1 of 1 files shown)


martincostello:
While making exceptions public is useful for being able to catch it ([CA1064: Exceptions should be public](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1064)), if we were to make this specific exception public, we'd want it to follow the recommended guidance for implementing them, such as [CA1032: Implement standard exception constructors](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1032) as well before doing so. I would also make it `sealed`.

FoxTes:
Yes, your remark is correct.

SimonCropp:
exceptions should only be public if there is a legitimate reason to catch them

FoxTes:
So what decision will we make?

PR #974, opened 2022-12-14
update to Moq 4.18.3

SimonCropp:


martincostello:
Are we dropping the guidance to merge to `v724-or-v730` instead of the default branch?

martincostello:
I’m happy with that if @joelhulen’s cool with it - it’s less effort 😆

As long as we’re all doing the same thing so we don’t end up with two different branches and a merge headache later.

joelhulen:
@SimonCropp and @martincostello, I'm cool with that change. It's less confusing and less effort this way.

martincostello (approved):

PR #1006, opened 2023-01-08
Move to implicit usings

SimonCropp:


src/Polly.Specs/GlobalUsings.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Global using directives
```
martincostello, line 1: IMHO using the `<Using Include="..." />` item syntax in the project file is nicer that having to create an extra `.cs` file.

You can also put them in the `Directory.Build.props` file and cascade them to all projects and use conditions to activate them (`IsTestProject == true`) etc.

src/Polly/Polly.csproj
```diff
@@ -33,6 +33,7 @@
     <PackageReference Include="System.ValueTuple" Version="4.5.0" />
   </ItemGroup>
   <ItemGroup>
+    <Using Remove="System.Net.Http" />
```
martincostello, line 36: Did you mean to add this one?

SimonCropp: yeah thats a known issue with implicit usings. one of the default includes when u do `ImplicitUsings>enable` but it doesnt exist in older frameworks. u can see how common it is https://sourcegraph.com/search?q=context:global+%3CUsing+Remove%3D%22System.Net.Http%22+/%3E&patternType=standard&sm=1

src/Polly.Specs/GlobalUsings.cs
SimonCropp: `<Using Include="..." />`. ok cool. i have gone that approach in some projects. I will move this PR over

SimonCropp:
```
You can also put them in the Directory.Build.props file and cascade them to all projects and use conditions to activate them (IsTestProject == true) etc.
```

will do

SimonCropp:
@martincostello feedback addressed. note it will need a squash if u decide to merge it

martincostello (approved):

PR #1007, opened 2023-01-08
Enable nullables part 1

SimonCropp:
this is a first pass at enabling nullables. 

I think this is a big enough chunk to do a code review

### DIFF
--- a/src/Polly/AsyncPolicy.NonGenericImplementation.cs
+++ b/src/Polly/AsyncPolicy.NonGenericImplementation.cs
@@ -40,5 +40,6 @@ public abstract partial class AsyncPolicy
         Context context,
         CancellationToken cancellationToken,
         bool continueOnCapturedContext
-    );
+    )
+        where TResult : notnull;
 }
--- a/src/Polly/AsyncPolicy.NonGenericImplementation.cs
+++ b/src/Polly/AsyncPolicy.NonGenericImplementation.cs
@@ -1,9 +1,4 @@
-﻿using System;
-using System.Threading;
-using System.Threading.Tasks;
-using Polly.Utilities;
-
-namespace Polly;
+﻿namespace Polly;
 
 public abstract partial class AsyncPolicy
 {
--- a/src/Polly/AsyncPolicy.NonGenericImplementation.cs
+++ b/src/Polly/AsyncPolicy.NonGenericImplementation.cs
@@ -41,5 +41,4 @@ public abstract partial class AsyncPolicy
         CancellationToken cancellationToken,
         bool continueOnCapturedContext
     );
-
 }
--- a/src/Polly/AsyncPolicy.TResult.cs
+++ b/src/Polly/AsyncPolicy.TResult.cs
@@ -1,4 +1,5 @@
-﻿namespace Polly;
+﻿#nullable enable
+namespace Polly;
 
 /// <summary>
 /// Transient exception handling policies that can be applied to asynchronous delegates
@@ -12,8 +13,8 @@ public abstract partial class AsyncPolicy<TResult> : PolicyBase<TResult>
     /// <param name="exceptionPredicates">Predicates indicating which exceptions the policy should handle. </param>
     /// <param name="resultPredicates">Predicates indicating which results the policy should handle. </param>
     internal AsyncPolicy(
-        ExceptionPredicates exceptionPredicates,
-        ResultPredicates<TResult> resultPredicates)
+        ExceptionPredicates? exceptionPredicates,
+        ResultPredicates<TResult>? resultPredicates)
         : base(exceptionPredicates, resultPredicates)
     {
     }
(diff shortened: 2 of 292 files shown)


SimonCropp:
note i have already enable nullables for the tests

src/Polly/AsyncPolicy.TResult.cs
```diff
@@ -1,4 +1,5 @@
-﻿namespace Polly;
+﻿#nullable enable
```
martincostello, line 1: Is the plan to remove these and just set it in the project file once it's "done"?

src/Polly/AsyncPolicy.NonGenericImplementation.cs
```diff
@@ -40,5 +40,6 @@ public abstract partial class AsyncPolicy
         Context context,
         CancellationToken cancellationToken,
         bool continueOnCapturedContext
-    );
+    )
+        where TResult : notnull;
```
martincostello, line 44: Does this preclude someone from a return type like `Task<Foo?>`?

src/Polly/Caching/AsyncCacheEngine.cs
```diff
@@ -43,7 +45,7 @@ internal static class AsyncCacheEngine
         if (cacheHit)
```
martincostello, line 45: Is there a way to annotate it so the compiler knows that if `cacheHit` is `true` then `valueFromCache` is non-null?

martincostello: Or, are there use cases where `null` is a valid value in a cache (e.g. _"I've looked this up and it doesn't exist, so cache that fact"_), and we should therefore allow null?

src/Polly/Caching/AsyncGenericCacheProvider.cs
```diff
@@ -9,16 +10,17 @@ namespace Polly.Caching;
...
 internal class AsyncGenericCacheProvider<TCacheFormat> : IAsyncCacheProvider<TCacheFormat>
+    where TCacheFormat : notnull
 {
     private readonly IAsyncCacheProvider _wrappedCacheProvider;
 
     internal AsyncGenericCacheProvider(IAsyncCacheProvider nonGenericCacheProvider)
         => _wrappedCacheProvider = nonGenericCacheProvider ?? throw new ArgumentNullException(nameof(nonGenericCacheProvider));
 
-    async Task<(bool, TCacheFormat)> IAsyncCacheProvider<TCacheFormat>.TryGetAsync(string key, CancellationToken cancellationToken, bool continueOnCapturedContext)
+    async Task<(bool, TCacheFormat?)> IAsyncCacheProvider<TCacheFormat>.TryGetAsync(string key, CancellationToken cancellationToken, bool continueOnCapturedContext)
```
martincostello, line 20: A named tuple could be good here as this is internal.

src/Polly/Caching/AsyncSerializingCacheProvider.cs
```diff
@@ -36,10 +38,10 @@ public AsyncSerializingCacheProvider(IAsyncCacheProvider<TSerialized> wrappedCac
     /// A <see cref="Task{TResult}" /> promising as Result a tuple whose first element is a value indicating whether
     /// the key was found in the cache, and whose second element is the value from the cache (null if not found).
     /// </returns>
-    public async Task<(bool, object)> TryGetAsync(string key, CancellationToken cancellationToken, bool continueOnCapturedContext)
+    public async Task<(bool, object?)> TryGetAsync(string key, CancellationToken cancellationToken, bool continueOnCapturedContext)
```
martincostello, line 41: +1 to named tuple.

martincostello: Nevermind, I see that it's implementing a public interface.

src/Polly/CircuitBreaker/RollingHealthMetrics.cs
```diff
@@ -24,14 +25,14 @@ public void IncrementSuccess_NeedsLock()
     {
         ActualiseCurrentMetric_NeedsLock();
 
-        _currentWindow.Successes++;
+        _currentWindow!.Successes++;
```
martincostello, line 28: Is there an annotation we can add to hint the compiler that calls to a certain method will result in this being non-null?

src/Polly/CircuitBreaker/SingleHealthMetrics.cs
```diff
@@ -7,22 +8,22 @@ internal class SingleHealthMetrics : IHealthMetrics
...
+    private HealthCount? _current;
 
     public SingleHealthMetrics(TimeSpan samplingDuration) => _samplingDuration = samplingDuration.Ticks;
 
     public void IncrementSuccess_NeedsLock()
     {
         ActualiseCurrentMetric_NeedsLock();
 
-        _current.Successes++;
+        _current!.Successes++;
```
martincostello, line 19: Same question.

martincostello (commented):
Maybe I've just gotten confused between _nullable_ and _nullability_, but I'm not sure if the constraints for `TResult: nonnull` are correct.

src/Polly/AsyncPolicy.TResult.cs
SimonCropp: yep. `#nullable enable` allows us to gradually transition

src/Polly/AsyncPolicy.NonGenericImplementation.cs
SimonCropp: yes. is that correct?

martincostello: That doesn't sound right to me - if a user wants to return `null` from their policy they should be able to.

SimonCropp:
@martincostello i think in retrospect this one is too large to reason about. i will submit a few smaller ones

PR #1009, opened 2023-01-08
also benchmark on net7

SimonCropp:
IMO benchmarks  should be run against the current stable .net


martincostello (approved):
Maybe have the option for both TFMs as 6 is LTS, but otherwise I don't think benchmarking with 7 is an issue.

SimonCropp:
@martincostello done

> Maybe have the option for both TFMs as 6 is LTS

PR #1010, opened 2023-01-08
also run tests on net7

SimonCropp:
I wasnt sure if we should still run tests on net6?


martincostello (approved):
As 6 is LTS and 7 isn't, IMHO we should test both if we test with 7.

SimonCropp:
> As 6 is LTS and 7 isn't, IMHO we should test both if we test with 7.

@martincostello done

PR #1011, opened 2023-01-08
move tests from net472 to 481

SimonCropp:
given it is trivial for people to move from net472 to 481, i think 481 is what we should test against


martincostello (approved):
I don't think it's a problem for us to test on the latest version, but 4.8.1 is only supported on Windows 10+, so older still-supported versions of Windows Server (like 2019) can only go as high as 4.8 so not all users would/could be running it for their production workloads.

PR #1015, opened 2023-01-11
try get nullable

SimonCropp:
@martincostello 

i put together a minimal example to show how i came to use the NotNull for `TResult` in  `IAsyncCacheProvider<TResult>`

this PR enables nullable for `AsyncGenericCacheProvider.cs` and `IAsyncCacheProvider`

you will notice we get the warning

```
  AsyncGenericCacheProvider.cs(22, 48): [CS8604] Possible null reference argument for parameter 'value' in 'Task IAsyncCacheProvider.PutAsync(string key, object value, Ttl ttl, CancellationToken cancellationToken, bool continueOnCapturedContext)'.
```

so in 
```
internal class AsyncGenericCacheProvider<TCacheFormat> : IAsyncCacheProvider<TCacheFormat>
{
    private readonly IAsyncCacheProvider _wrappedCacheProvider;

    internal AsyncGenericCacheProvider(IAsyncCacheProvider nonGenericCacheProvider)
        => _wrappedCacheProvider = nonGenericCacheProvider ?? throw new ArgumentNullException(nameof(nonGenericCacheProvider));

    async Task<(bool, TCacheFormat?)> IAsyncCacheProvider<TCacheFormat>.TryGetAsync(string key, CancellationToken cancellationToken, bool continueOnCapturedContext)
    {
        (bool cacheHit, object? result) = await _wrappedCacheProvider.TryGetAsync(key, cancellationToken, continueOnCapturedContext).ConfigureAwait(continueOnCapturedContext);
        return (cacheHit, (TCacheFormat?)(result ?? default(TCacheFormat)));
    }

    Task IAsyncCacheProvider<TCacheFormat>.PutAsync(string key, TCacheFormat value, Ttl ttl, CancellationToken cancellationToken, bool continueOnCapturedContext)
        => _wrappedCacheProvider.PutAsync(key, value, ttl, cancellationToken, continueOnCapturedContext);
}
```

since someone could define `AsyncGenericCacheProvider<MyType?>` it means `value` could be null in 

```
_wrappedCacheProvider.PutAsync(key, value, ttl, cancellationToken, continueOnCapturedContext);
```

so we could allow null for value through the stack. so `IAsyncCacheProvider` would need to accept null for `PutAsync`. Do we want that?

if we `notnull` on `TCacheFormat` it means people cant pass a null `value` to `PutAsync`.

TBH i think the fix here is to add a `Clear` method that takes a key??

thoughts?

### DIFF
--- a/src/Polly/Caching/AsyncCacheEngine.cs
+++ b/src/Polly/Caching/AsyncCacheEngine.cs
@@ -1,4 +1,5 @@
-﻿namespace Polly.Caching;
+﻿#nullable enable
+namespace Polly.Caching;
 
 internal static class AsyncCacheEngine
 {
@@ -25,7 +26,7 @@ internal static class AsyncCacheEngine
         }
 
         bool cacheHit;
-        TResult valueFromCache;
+        TResult? valueFromCache;
         try
         {
             (cacheHit, valueFromCache) = await cacheProvider.TryGetAsync(cacheKey, cancellationToken, continueOnCapturedContext).ConfigureAwait(continueOnCapturedContext);
@@ -39,7 +40,7 @@ internal static class AsyncCacheEngine
         if (cacheHit)
         {
             onCacheGet(context, cacheKey);
-            return valueFromCache;
+            return valueFromCache!;
         }
         else
         {
--- a/src/Polly/Caching/CacheEngine.cs
+++ b/src/Polly/Caching/CacheEngine.cs
@@ -1,4 +1,5 @@
-﻿namespace Polly.Caching;
+﻿#nullable enable
+namespace Polly.Caching;
 
 internal static class CacheEngine
 {
@@ -24,7 +25,7 @@ internal static class CacheEngine
         }
 
         bool cacheHit;
-        TResult valueFromCache;
+        TResult? valueFromCache;
         try
         {
             (cacheHit, valueFromCache) = cacheProvider.TryGet(cacheKey);
@@ -38,7 +39,7 @@ internal static class CacheEngine
         if (cacheHit)
         {
             onCacheGet(context, cacheKey);
-            return valueFromCache;
+            return valueFromCache!;
         }
         else
         {
--- a/src/Polly.Benchmarks/Cache.cs
+++ b/src/Polly.Benchmarks/Cache.cs
@@ -67,7 +67,7 @@ public class Cache
             return (cacheHit, value);
         }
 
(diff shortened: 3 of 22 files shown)


martincostello:
I think it depends on what the original intention of the cache policy is - it currently allows null values, so I would have thought the nullable annotations would therefore continue to allow that, otherwise that's a change in behaviour.

Looking at `IMemoryCache` in dotnet/runtime for some prior art, it allows null for cached values, just not keys:

* [`object? ICacheEntry.Value { get; set; }`](https://github.com/dotnet/runtime/blob/21af7c40b1141030921ec1d825f31c88a0fa6cf5/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/ICacheEntry.cs#L21-L24)
* [`public static object? Get(this IMemoryCache cache, object key)`](https://github.com/dotnet/runtime/blob/21af7c40b1141030921ec1d825f31c88a0fa6cf5/src/libraries/Microsoft.Extensions.Caching.Abstractions/src/MemoryCacheExtensions.cs#L12-L16)

SimonCropp:
@martincostello i made some changes based on your feedback. can u take another look.

src/Polly/Caching/AsyncCacheEngine.cs
```diff
@@ -39,7 +40,7 @@ internal static class AsyncCacheEngine
         if (cacheHit)
         {
             onCacheGet(context, cacheKey);
-            return valueFromCache;
+            return valueFromCache!;
```
martincostello, line 43: Should this method return `TResult?` so that this isn't needed?

src/Polly/Caching/CacheEngine.cs
```diff
@@ -38,7 +39,7 @@ internal static class CacheEngine
         if (cacheHit)
         {
             onCacheGet(context, cacheKey);
-            return valueFromCache;
+            return valueFromCache!;
```
martincostello, line 42: +1

src/Polly/Caching/AsyncCacheEngine.cs
SimonCropp: i dont think so. if we have a cacheHit then we know valueFromCache is not null. so we can be sure ImplementationAsync will return a not null

martincostello: I thought we changed things based on my earlier comment to allow you to set `null` into the cache, so in that case you'd get a cache hit on a null.

PR #1019, opened 2023-01-12
Polly V8 proof of concept demo

martintmk:

The V8 proof of concept (POC) demo. The docs are [here](https://github.com/App-vNext/Polly/blob/mtomka/v8-poc/docs/Polly-V8.md).

Not applicable.

### Confirm the following

Not applicable.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/App-vNext/Polly?pullRequest=1019) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/App-vNext/Polly?pullRequest=1019) before we can accept your contribution.<br/><br/>:x: martintmk [sign now](https://cla.dotnetfoundation.org/App-vNext/Polly?pullRequest=1019)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/App-vNext/Polly?pullRequest=1019) it.</sub>

SimonCropp:
what is the value of having an two extra cspsoj?

martintmk:
> what is the value of having two extra csproj?

Smaller projects are easier to manage and orient around. They can also improve the layering and architecture. Also, it does not mean we have to have the same number of packages. We can have just one package that has multiple assemblies.

For me, in this draft I tried to organize the projects by the logical layers. Maybe some stuff can be merged together, it's still a prototype.

SimonCropp:
> Smaller projects are easier to manage and orient around. They can also improve the layering and architecture

that is artificial. you can achieve the same with directories and namespaces in a single project

martintmk:
> that is artificial. you can achieve the same with directories and namespaces in a single project

I tried to put some more details into the packages here:
https://github.com/App-vNext/Polly/blob/mtomka/v8-poc/docs/Polly-V8.md#packages

Maybe the Core and Strategies can be merged together, however, I see the value of having `Polly.Abstractions` with the absolute minimum API surface and dependencies. The abstractions packages follow the same pattern as many libraries in `Microsoft.Extensions*` and allows for 3rd party libraries to take dependency on Polly and enable extensibility without caring about all the different strategies and infrastructure around that.

src/Polly.Abstractions/VoidResult.cs
```diff
@@ -0,0 +1,10 @@
+namespace Polly;
+
+internal class VoidResult
```
martincostello, line 3: `sealed`?

src/Polly.Core/Polly.Core.csproj
```diff
@@ -0,0 +1,18 @@
...
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net7.0</TargetFrameworks>
+    <UsePollyTarget>true</UsePollyTarget>
+    <AssemblyTitle>Polly.Core</AssemblyTitle>
+  </PropertyGroup>
+
+  
+  <ItemGroup Label="PackageReferences">
+    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="7.0.0" />
+  </ItemGroup>
```
martincostello, line 12: Same question here regarding TFM and multi-targeting.

src/Polly.Core/ResilienceStrategyBuilder.cs
```diff
@@ -0,0 +1,59 @@
...
+    public IResilienceStrategyBuilder AddStrategy(Func<ResilienceStrategyBuilderContext, IResilienceStrategy> factory, ResilienceStrategyOptions? options = null)
+    {
+        _entries.Add(new Entry(factory, options ?? new ResilienceStrategyOptions()));
+
+        return this;
+    }
+
+    public IResilienceStrategy Create()
+    {
+        var strategies = new List<DelegatingResilienceStrategy>();
```
martincostello, line 22:
```suggestion
        var strategies = new List<DelegatingResilienceStrategy>(_entries.Count);
```

src/Polly.Core/ResilienceStrategyBuilder.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        for (var i = 0; i < strategies.Count - 1; i++)
+        {
+            strategies[i].Next = strategies[i + 1];
+        }
+
+        return strategies[0];
+    }
+
+    private class Entry
```
martincostello, line 47:
```suggestion
    private sealed class Entry
```

src/Polly.DependencyInjection/Polly.DependencyInjection.csproj
```diff
@@ -0,0 +1,19 @@
...
+
+
+  <ItemGroup>
+    <ProjectReference Include="..\Polly.Core\Polly.Core.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Options" Version="7.0.0" />
+  </ItemGroup>
```
martincostello, line 17: +1 TFM and multi-targeting.

src/Polly.Strategies.Bench/Internals/EmptyResilienceStrategy.cs
```diff
@@ -0,0 +1,8 @@
+// See https://aka.ms/new-console-template for more information
+
```
martincostello, line 2:
```suggestion
```

src/Polly.Strategies.Bench/Internals/Helper.SimplePipeline.cs
```diff
@@ -0,0 +1,55 @@
...
+                    StrategyName = "retries"
+                });
+
+                builder.AddTimeout(new TimeoutStrategyOptions
+                {
+                    TimeoutInterval = innerTimeout,
+                    OnTimeout = new Events<TimeoutTaskArguments>().Add(args => new ValueTask()),
+                    StrategyName = "inner-timeout"
+                });
+            }),
```
martincostello, line 51: One observation here is that comparing v7 to v8, v8 is a lot more verbose.

src/Polly.Strategies.Bench/Program.cs
```diff
@@ -0,0 +1,16 @@
+// © Microsoft Corporation. All rights reserved.
+
```
martincostello, line 2:
```suggestion
```

src/Polly.Strategies.Bench/Program.cs
```diff
@@ -0,0 +1,16 @@
...
+using BenchmarkDotNet.Running;
+using BenchmarkDotNet.Toolchains.InProcess.Emit;
+
+var config = ManualConfig
+    .Create(DefaultConfig.Instance)
+    .AddJob(Job.MediumRun.WithToolchain(InProcessEmitToolchain.Instance))
+    .AddDiagnoser(MemoryDiagnoser.Default);
+
+// BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args, config);
+
```
martincostello, line 15:
```suggestion
```

src/Polly.Strategies.Bench/README.md
```diff
@@ -0,0 +1,40 @@
+﻿# Benchmark results
```
martincostello, line 1: Would be good to maybe refactor the benchmarks slightly to use different methods for v7 and v8 so we can use baseline and get the comparisons computed for us to make the tables, and this the improvements, easier to grok.

src/Polly.Strategies/Events.cs
```diff
@@ -0,0 +1,49 @@
+﻿namespace Polly;
+
+public delegate ValueTask EventsDelegate<T, TContext>(Outcome<T> outcome, TContext context);
+
+public class Events<TContext>
```
martincostello, line 5: What's the purpose of this type in the big picture?

src/Polly.Strategies/Retry/RetryStrategyOptions.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Polly.Retry;
+public class RetryStrategyOptions : ResilienceStrategyOptions
```
martincostello, line 2:
```suggestion

public class RetryStrategyOptions : ResilienceStrategyOptions
```

src/Polly.Strategies/Retry/RetryStrategyOptions.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+    public Predicates<ShouldRetryArguments> ShouldRetry { get; set; } = new();
+
+    public Events<RetryActionArguments> OnRetry { get; set; } = new();
+
+    public int RetryCount { get; set; } = 3;
+
+    public Func<int, TimeSpan> RetryDelayGenerator { get; set; } = _ => TimeSpan.FromSeconds(3);
```
martincostello, line 10: I'm guessing these are our current defaults in v7?

src/Polly/Polly.csproj
```diff
@@ -1,50 +1,12 @@
...
-    <Description>Polly is a library that allows developers to express resilience and transient fault handling policies such as Retry, Circuit Breaker, Timeout, Bulkhead Isolation, and Fallback in a fluent and thread-safe manner.</Description>
-    <DefaultLanguage>en-US</DefaultLanguage>
-    <GenerateAssemblyInfo>true</GenerateAssemblyInfo>
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <Authors>Michael Wolfenden, App vNext</Authors>
-    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
-    <EmbedUntrackedSources>true</EmbedUntrackedSources> <!-- EmbedUntrackedSources for deterministic build -->
-    <ContinuousIntegrationBuild Condition=" '$(CI)' != '' ">true</ContinuousIntegrationBuild>
-    <Deterministic>true</Deterministic>
+    <TargetFrameworks>netstandard2.0;net461;net472</TargetFrameworks>
```
martincostello, line 4: Did we discuss dropping `net461` in our last meeting? I can't remember.

martincostello (commented):
I've reviewed about 75% of the diff so far. I'll get to the rest by the end of Monday.

On the whole I think this is shaping up nicely.

src/Polly.Abstractions/ResilienceStrategyExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+public static partial class ResilienceStrategyExtensions
+{
+    public static async ValueTask ExecuteAsync<TState>(this IResilienceStrategy strategy, Func<TState, CancellationToken, ValueTask> execute, TState state, CancellationToken cancellationToken = default)
+    {
+        var context = ResilienceContext.Get(cancellationToken);
+        context.IsVoid = true;
+
+        try
+        {
+            await strategy.ExecuteAsync(
```
geeknoid, line 14: Need to call ConfigureAwait(false) for all uses of await in the code.

src/Polly.Strategies/Timeout/TimeoutResilienceStrategy.cs
```diff
@@ -0,0 +1,54 @@
+using Microsoft.Extensions.ObjectPool;
+using Polly.Internals;
+
+namespace Polly.Timeout;
+
+internal class TimeoutResilienceStrategy : DelegatingResilienceStrategy
```
geeknoid, line 6: All internal class should be sesled for improved perf.

src/Polly.Strategies/Predicates.cs
```diff
@@ -0,0 +1,57 @@
+﻿namespace Polly;
+
+public delegate ValueTask<bool> PredicatesDelegate<T, TContext>(Outcome<T> outcome, TContext context);
+
+public class Predicates<TContext>
+{
+    internal Dictionary<Type, List<object>> Callbacks { get; } = new();
+
+    public Predicates<TContext> AddException<T>()
```
geeknoid, line 9: Can all these single-line functions be coded with expression syntax for brevity?

src/Polly.Strategies/Outcome.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Polly;
+
+public struct Outcome<T>
```
geeknoid, line 3: Should this be readonly?

src/Polly.Strategies/Outcome.cs
```diff
@@ -0,0 +1,19 @@
...
+        Result = result;
+    }
+
+    public Outcome(Exception? exception) : this()
+    {
+        Exception = exception;
+        Result = default!;
+    }
+
+    public T Result;
```
geeknoid, line 16: Should this be a readonly property, instead of a read/write field?

src/Polly.Strategies/Predicates.cs
```diff
@@ -0,0 +1,57 @@
+﻿namespace Polly;
+
+public delegate ValueTask<bool> PredicatesDelegate<T, TContext>(Outcome<T> outcome, TContext context);
```
geeknoid, line 3: Best not to use the word "Delegate" in the name. Just like we don't call classes FooClass. Don't focus on the programming construct, focus on what the delegete is semantically for.

src/Polly.Strategies/Outcome.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Polly;
+
```
geeknoid, line 2: Would it make sense to put the Outcome, Events, Predicates type in the .Abstractions package? Specifically, if I were to implement new strategies, I should be able to do so by importing only the .Abstractions package and not the standard Strategies package.

src/Polly.Strategies/Timeout/TimeoutTaskArguments.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Polly.Timeout;
+
+public readonly struct TimeoutTaskArguments
```
geeknoid, line 3: Why is Task in the name?

geeknoid: Maybe TimeoutStrategyArguments? Then you could use the same pattern for RetryStrategyArguments, etc.

src/Polly.Strategies/Retry/RetryStrategy.cs
```diff
@@ -0,0 +1,96 @@
...
+    internal static Task SleepAsync(ResilienceContext context, TimeSpan delay)
+    {
+        if (delay == TimeSpan.Zero)
+        {
+            return Task.CompletedTask;
+        }
+
+        if (context.IsSynchronous)
+        {
+            Thread.Sleep(delay);
```
geeknoid, line 88: We must virtualize all timer usage to enable robust testing.

docs/Polly-V8.md
```diff
@@ -0,0 +1,234 @@
...
+
+- `ISyncPolicy`
+- `IAsyncPolicy`
+- `ISyncPolicy<T>`
+- `IAsyncPolicy<T>`
+
+``` csharp
+public interface IResilienceStrategy
+{
+    ValueTask<T> ExecuteAsync<T, TState>(Func<ResilienceContext, TState, ValueTask<T>> execution, ResilienceContext context, TState state);
```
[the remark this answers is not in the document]
martintmk: I agree `TResult` is more descriptive so I made the changes. As for the order I did it this way to be consistent in cases where `TState` is not used.  User then expects `TResult` in the first place. If he adds `TState` then the order is consistent, for example `TResult, then `TState`.

Similar patterns as in [here](https://source.dot.net/#Microsoft.Extensions.Options/OptionsBuilder.cs,198)

docs/Polly-V8.md
```diff
@@ -0,0 +1,234 @@
...
+}
+```
+
+The `ResilienceContext` is defined as:
+
+``` csharp
+public class ResilienceContext
```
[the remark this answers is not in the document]
martintmk: I would vote for sealed. In our experience is better to expose the absolute minimum API and expand it later when there is a request. If we seal it will be fully under our control and also slightly faster. 

We can still unseal it later if someone needs it. 

Also, in our scenarios we didn't find any cases to derive from `Context` and we used pretty advanced stuff, at most we are adding custom data to the context using the dictionary. For simplicity I have not added this support to `ResilienceContext` in the POC.

docs/Polly-V8.md
```diff
@@ -0,0 +1,234 @@
...
+``` csharp
+public interface IResilienceStrategyBuilder
+{
+    ResilienceStrategyBuilderOptions Options { get; set; }
+
+    IResilienceStrategyBuilder AddStrategy(IResilienceStrategy strategy, ResilienceStrategyOptions? options = null);
+
+    IResilienceStrategyBuilder AddStrategy(Func<ResilienceStrategyBuilderContext, IResilienceStrategy> factory, ResilienceStrategyOptions? options = null);
+
+    IResilienceStrategy Create();
```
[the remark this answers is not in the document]
martintmk: True, `Build` seems better.

docs/Polly-V8.md
```diff
@@ -0,0 +1,234 @@
...
+
+Pipeline of strategies:
+
+``` csharp
+var resilienceStrategy = new ResilienceStrategyBuilder()
+    .AddRetry()
+    .AddCircuitBreaker()
+    .AddTimeout(new TimeoutStrategyOptions() { ... })
+    .Create();
+```
```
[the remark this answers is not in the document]
martintmk: I think we can focus first on the core primitives ;) The registry and strategy retrieval will be layer above that. 

For this reason, I have also dropped extra DI package to keep the POC simple.

src/Polly.Strategies.Bench/README.md
martintmk: Can you elaborate here? Not sure what to do and how to cahnge them.

docs/Polly-V8.md
```diff
@@ -0,0 +1,234 @@
...
+- `DelegatingResilienceStrategy`
+
+#### Polly.Core
+
+Contains the builder for resilience strategies and core primitives that can be reused across various strategies.
+
+**Dependencies:**
+
+- `Poly.Abstractions`
+- `Microsoft.Extensions.Logging.Abstractions`
```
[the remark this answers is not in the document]
martintmk: We have added a lot of boilerplate code on top of Polly just to add telemetry. For us telemetry is a must as we need to track any failures and react on that. It would be great if all strategies have telemetry by default. 

If not enabled, it adds basically zero overheat. For this reason, I have included the `Microsoft.Extensions.Logging.Abstractions` in here. It's a core package that's very stable, because a LOT of modern libraries depend on it. We can also make our logger just to keep this dependency out, but I would discourage that.

docs/Polly-V8.md
```diff
@@ -0,0 +1,234 @@
...
+**Dependencies:**
+
+- `Poly.Core`
+
+**Types**:
+
+- `RetryStrategyOptions`
+- `TimeoutStrategyOptions`
+- `RetryExtensions`
+- `TimeoutExtensions`
```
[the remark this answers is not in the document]
martintmk: Yup, just omitted those. Update the docs slightly.

src/Polly.Strategies/Events.cs
martintmk: I have updated the docs and added [more details](https://github.com/App-vNext/Polly/blob/mtomka/v8-poc/docs/Polly-V8.md#handling-of-different-result-types) about these.

src/Polly.Abstractions/ResilienceStrategyExtensions.cs
```diff
@@ -0,0 +1,89 @@
+﻿using Polly;
+
+namespace Resilience;
+
+public static partial class ResilienceStrategyExtensions
+{
+    public static async ValueTask ExecuteAsync<TState>(this IResilienceStrategy strategy, Func<TState, CancellationToken, ValueTask> execute, TState state, CancellationToken cancellationToken = default)
+    {
+        var context = ResilienceContext.Get(cancellationToken);
```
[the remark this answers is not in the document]
martintmk: I have completely omitted any argument checking in the POC for simplicity. Once we start splitting it into smaller chunks and creating real PRs, we can make it proper.

src/Polly.Strategies/Outcome.cs
martintmk: Implementation of new strategy also assumes integration with `IResilienceStrategyBuilder` which is in the Core. 

However, the `IResilienceStrategyBuilder` also pulls in more packages (`Logging.Abstractions`) that's why I thought we can keep it in Core.

src/Polly.Strategies/Retry/RetryStrategy.cs
martintmk: Yeah definitely, have not included it here for simplicity. I think we will have some property like `ResilienceStrategyBuilderOptions.SystemClock`. Similar to what's in the Polly now, but it won't be static anymore.

src/Polly.Strategies/Timeout/TimeoutTaskArguments.cs
martintmk: I think the arguments class should follow the name of the event where it is used. The property is called `OnTimeout` so my vote would be `OnTimeoutArguments`. 

The task stuff is just a copy-pate error :)

src/Polly.Abstractions/ResilienceContext.cs
```diff
@@ -0,0 +1,44 @@
...
+
+public sealed class ResilienceContext
+{
+    private static readonly ObjectPool<ResilienceContext> s_pool = ObjectPool.Create<ResilienceContext>();
+
+    public ResilienceContext()
+    {
+    }
+
+    public CancellationToken CancellationToken { get; set; }
```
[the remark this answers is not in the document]
martintmk: I was thinking we should use the same pattern as in `HttpContext` where modifications such as these are allowed. For example, timeout strategy replaces this token with a new one and passes it down the chain.

src/Polly.Strategies/Outcome.cs
martintmk: Yeah, my laziness. Fixed.

src/Polly.Strategies/Outcome.cs
martintmk: Yup.

src/Polly/Polly.csproj
martintmk: The `net461` is out of support, I think we can support the following frameworks:

- `net462`: still is supported
- `nestandard2.0`
- net6.0
- net7.0

The core frameworks bellow `net6.0` are out of support and are covered by  `nestandard2.0`

src/Polly.Abstractions/ResilienceStrategyExtensions.Synchronous.cs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private static T GetResult<T>(this ValueTask<T> task)
+    {
+        Debug.Assert(
+            task.IsCompleted,
+            "The value task should be already completed at this point. If not, its and indication that the strategy does not respect the ResilienceContext.IsSynchronous value.");
+
+#pragma warning disable S5034 // "ValueTask" should be consumed correctly
+#pragma warning disable VSTHRD002 // Avoid problematic synchronous waits
```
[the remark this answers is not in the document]
martintmk: We are using the following public analyzers:

    <PackageVersion Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="7.0.0-preview1.22464.1" />
    <PackageVersion Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.3.48" />
    <PackageVersion Include="SonarAnalyzer.CSharp" Version="8.47.0.55603" />
    <PackageVersion Include="StyleCop.Analyzers.Unstable" Version="1.2.0.376" />

The `VSTHRD002` is from internal one, so I dropped the suppression.

src/Polly.Core/Polly.Core.csproj
martintmk: The `Microsoft.Extensions.Logging.Abstractions` is not implicit in any framework and needs to be explicitly added.

src/Polly.DependencyInjection/Polly.DependencyInjection.csproj
martintmk: Dropped this project. Not needed for POC.

src/Polly.Strategies.Bench/Internals/Helper.SimplePipeline.cs
martintmk: We can also include additional convenience extensions on top of the options-based ones. In POC I kept it simple.

src/Polly.Strategies/Retry/RetryStrategyOptions.cs
martintmk: Not really, just some random values I picked. we can fine-tune these in real PRs.

src/Polly.Strategies/Predicates.cs
martintmk: Renamed to same term as used in docs `PredicatesCallback`

src/Polly.Abstractions/Polly.Abstractions.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;net7.0</TargetFrameworks>
```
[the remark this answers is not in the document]
martintmk: Done.

src/Polly.Abstractions/Polly.Abstractions.csproj
```diff
@@ -0,0 +1,14 @@
...
+    <UsePollyTarget>true</UsePollyTarget>
+    <AssemblyTitle>Polly.Abstractions</AssemblyTitle>
+  </PropertyGroup>
+
+  <ItemGroup Label="PackageReferences">
+    <PackageReference Include="Microsoft.Extensions.ObjectPool" Version="7.0.1" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
```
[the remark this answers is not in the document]
martintmk: Yup correct, I suggest to start using https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management to simplify multi-targeting. 

In POC is just a single version for simplicity.

src/Polly.Strategies/Timeout/TimeoutResilienceStrategy.cs
```diff
@@ -0,0 +1,54 @@
+using Microsoft.Extensions.ObjectPool;
+using Polly.Internals;
+
+namespace Polly.Timeout;
+
+internal sealed class TimeoutResilienceStrategy : DelegatingResilienceStrategy
```
geeknoid, line 6: Rename to TimeoutStrategy to be consistent in naming style with RetryStrategy.

docs/Polly-V8.md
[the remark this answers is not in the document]
martincostello: Making it sealed and then opening later if an unforeseen use case arises makes sense to me 👍

docs/Polly-V8.md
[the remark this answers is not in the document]
martincostello: Yeah I can get the reasoning - I just know that the wider community can get a bit _"flaming pitchfork"_ over additional dependencies and which logging framework to use etc.

In a vacuum, I wouldn't mind adding it, just raising the point for discussion.

src/Polly.Abstractions/Polly.Abstractions.csproj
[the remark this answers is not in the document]
martincostello: I use CPMV in my own personal projects, so I wouldn't have a problem with enabling it in Polly if people are in agreement.

src/Polly.Abstractions/ResilienceStrategyExtensions.Synchronous.cs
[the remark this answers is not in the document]
martincostello: I'll take a look at enabling all/more analysers in `main`, which should then make it easy to surface/identity stuff like `ConfigureAwait(false)`, null-checking etc, that's already been commented on elsewhere in the PR.

src/Polly.Strategies/Retry/RetryStrategy.cs
martincostello: That would be good - that would also fit with things like the [`ISystemClock`](https://github.com/dotnet/aspnetcore/blob/main/src/Middleware/OutputCaching/src/ISystemClock.cs) pattern used in various places inside ASP.NET Core.

src/Polly.Strategies.Bench/README.md
martincostello: I meant something like this:

```csharp
public class IndexOfAnyBenchmarks
{
    private const string AUrlWithAPathAndQueryString = "http://www.example.com/path/to/file.html?query=string";
    private static readonly char[] QueryStringAndFragmentTokens = new[] { '?', '#' };

    [Benchmark(Baseline = true)]
    public int IndexOfAny_String()
    {
        return AUrlWithAPathAndQueryString.IndexOfAny(QueryStringAndFragmentTokens);
    }

    [Benchmark]
    public int IndexOfAny_Span_Array()
    {
        return AUrlWithAPathAndQueryString.AsSpan().IndexOfAny(QueryStringAndFragmentTokens);
    }
}
```

Then you get the extra columns in the results (`Ratio`, `RatioSD`, Alloc Ratio`) that show you how much faster `IndexOfAny_String` is compared to `IndexOfAny_Span_Array` in this example without having to work it out yourself (here, 29% faster).

``` ini

BenchmarkDotNet=v0.13.4, OS=Windows 11 (10.0.22000.1335/21H2)
Intel Core i9-9980HK CPU 2.40GHz, 1 CPU, 16 logical and 8 physical cores
.NET SDK=7.0.102
  [Host]     : .NET 6.0.13 (6.0.1322.58009), X64 RyuJIT AVX2
  DefaultJob : .NET 6.0.13 (6.0.1322.58009), X64 RyuJIT AVX2


```
|                    Method |     Mean |     Error |    StdDev |   Median | Ratio | RatioSD | Allocated | Alloc Ratio |
|-------------------------- |---------:|----------:|----------:|---------:|------:|--------:|----------:|------------:|
|         IndexOfAny_String | 9.167 ns | 0.2493 ns | 0.7193 ns | 8.938 ns |  1.00 |    0.00 |         - |          NA |
|     IndexOfAny_Span_Array | 6.529 ns | 0.1603 ns | 0.1574 ns | 6.525 ns |  0.71 |    0.03 |         - |          NA |

src/Polly.Strategies/Retry/RetryStrategy.cs
martintmk: So, are you OK if I keep this out of POC or should I add the abstraction?

src/Polly.Abstractions/Polly.Abstractions.csproj
```diff
@@ -0,0 +1,22 @@
...
+  <ItemGroup Label="PackageReferences">
+    <!-- This reference is just in demo, Won't be included in real code. -->
+    <PackageReference Include="Microsoft.Extensions.ObjectPool" Version="7.0.1" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net462' or '$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net462'">
```
martincostello, line 18: Assuming it isn't available in any .NET Framework version:

```suggestion
  <ItemGroup Condition="$([MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)')) == '.NETFramework'">
```

src/Polly.Abstractions/ResilienceStrategyExtensions.cs
```diff
@@ -0,0 +1,86 @@
...
+        }
+        finally
+        {
+            ResilienceContext.Return(context);
+        }
+    }
+
+    public static ValueTask<TResult> ExecuteAsync<TResult>(this IResilienceStrategy strategy, Func<ResilienceContext, ValueTask<TResult>> execute, ResilienceContext context) => strategy.ExecuteAsync(static (context, state) => state(context), context, execute);
+
+    public static async ValueTask<TResult> ExecuteAsync<TResult, TState>(this IResilienceStrategy strategy, Func<TState, CancellationToken, ValueTask<TResult>> execute, TState state, CancellationToken cancellationToken)
```
martincostello, line 73: Nit: these lines are getting so long that GitHub is starting to wrap them, so would be nice to shorten to one parameter per line style where appropriate (IMHO).  E.g.:

```
    public static ValueTask<TResult> ExecuteAsync<TResult>(
        this IResilienceStrategy strategy,
        Func<ResilienceContext, ValueTask<TResult>> execute,
        ResilienceContext context)

    public static async ValueTask<TResult> ExecuteAsync<TResult, TState>(
        this IResilienceStrategy strategy,
        Func<TState, CancellationToken, ValueTask<TResult>> execute,
        TState state,
        CancellationToken cancellationToken)
```

src/Polly.Strategies/Internals/EventsHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                return;
+            }
+
+            foreach (var callback in callbacks)
+            {
+                var outcomeCallback = (EventsCallback<T, TContext>)callback;
+
+                await outcomeCallback(outcome, context).ConfigureAwait(resilienceContext.ContinueOnCapturedContext);
+            }
```
martincostello, line 47: Rather than return early, just invert the condition and put the `foreach` inside it and then there's only one point of return in the whole method?

src/Polly.Strategies/Internals/PredicatesHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+            if (!_callbacks.TryGetValue(typeof(T), out var predicates))
+            {
+                return false;
+            }
+
+            foreach (var predicate in predicates)
+            {
+                var outcomePredicate = (PredicatesCallback<T, TContext>)predicate;
+
+                if (await outcomePredicate(outcome, context).ConfigureAwait(resilienceContext.ContinueOnCapturedContext))
```
martincostello, line 29: Should we defensively cast with an expression (so we could ignore/skip), or should it never not cast so blowing up is fine?

src/Polly.Strategies/Retry/RetryStrategy.cs
```diff
@@ -0,0 +1,96 @@
...
+                    return result;
+                }
+
+                canRetry = tryCount < _maxRetries;
+                if (!canRetry)
+                {
+                    return result;
+                }
+
+                return result;
```
martincostello: Is this code right? Seems to return `result` no matter what `canRetry`'s value is.

src/Polly.Strategies/Timeout/TimeoutResilienceStrategy.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    private static readonly ObjectPool<CancellationTokenSource> _cancellations = ObjectPool.Create<CancellationTokenSource>();
```
martincostello: In another class you prefixed this `s_`. Personally I'd prefer without the `s`, but whichever choice it would be better to be consistent.

src/Polly.Strategies/Timeout/TimeoutStrategyOptions.cs
```diff
@@ -0,0 +1,10 @@
+namespace Polly.Timeout;
+
+public class TimeoutStrategyOptions : ResilienceStrategyOptions
+{
+    public TimeoutStrategyOptions() => StrategyType = "Timeout";
+
+    public TimeSpan TimeoutInterval { get; set; } = TimeSpan.FromSeconds(30);
```
martincostello, line 7: Default to be reviewed at a later point.

src/Polly.Strategies/Retry/Internals/RetryStrategy.cs
```diff
@@ -0,0 +1,96 @@
...
+                    return result;
+                }
+
+                canRetry = tryCount < _maxRetries;
+                if (!canRetry)
+                {
+                    return result;
+                }
+
+                return result;
```
martincostello, line 48: Is this code right? Seems to always `return result` regardless of the value of `canRetry`.

src/Polly.Strategies/Timeout/Internals/TimeoutResilienceStrategy.cs
```diff
@@ -0,0 +1,54 @@
+using Microsoft.Extensions.ObjectPool;
+using Polly.Internals;
+
+namespace Polly.Timeout.Internals;
+
+internal sealed class TimeoutStrategy : DelegatingResilienceStrategy
+{
+    private static readonly ObjectPool<CancellationTokenSource> _cancellations = ObjectPool.Create<CancellationTokenSource>();
```
martincostello, line 8: In another class this is prefixed with `s_`. I'd prefer without, but whichever is chosen it would be good to be consistent.

src/Polly.Strategies/Retry/RetryStrategy.cs
martincostello: I think it's fine to leave out for now.

src/Polly.Strategies.Bench/README.md
martintmk: Got it, updated the benchmarks slightly. Now baseline is also included.

src/Polly.Strategies/Internals/EventsHandler.cs
martintmk: My thought for callbacks was that we want to execute every callback registered. So even for one result type there can be multiple callbacks registered.

src/Polly.Strategies/Internals/PredicatesHandler.cs
martintmk: It should always case fine, because we have the setup fully in our control. The caller can only put supported types there (and is always stored as `PredicatesCallback<T, TContext>`).

src/Polly.Strategies/Timeout/TimeoutResilienceStrategy.cs
martintmk: I prefer `_` to, so renamed the other one.

src/Polly/ResilienceStrategyExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+
+    public static IAsyncPolicy<T> AsAsyncPolicy<T>(this IResilienceStrategy strategy) => new StrategyAsyncPolicy<T>(strategy);
+
+    public static ISyncPolicy AsSyncPolicy(this IResilienceStrategy strategy) => new StrategySyncPolicy(strategy);
+
+    public static ISyncPolicy<T> AsSyncPolicy<T>(this IResilienceStrategy strategy) => new StrategySyncPolicy<T>(strategy);
+
+    public static ResilienceContext Update(this ResilienceContext context, Context pollyContext) =>
+        // we will probably just use the Context directly
+        context;
```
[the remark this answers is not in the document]
martintmk: Yup, added.

src/Polly.Abstractions/ResilienceStrategyExtensions.Synchronous.cs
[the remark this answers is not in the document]
martincostello: #1027

PR #1023, opened 2023-01-14
propagate null for error delegates

SimonCropp:
reduces alloc and deletes some code


martincostello:
I guess this was more _allow_ null than _propagate_ null.

PR #1024, opened 2023-01-15
remove Serializable

SimonCropp:
given binary serializer is no longer supported.

and it was never implementd properly in Polly. take `RateLimitRejectedException` for example. the `RetryAfter` property is not added in GetObjectData and not read back in the deserialize constructor

### DIFF
--- a/src/Polly/Bulkhead/BulkheadRejectedException.cs
+++ b/src/Polly/Bulkhead/BulkheadRejectedException.cs
@@ -1,16 +1,10 @@
 ﻿#nullable enable
-#if NETSTANDARD2_0
-using System.Runtime.Serialization;
-#endif
 
 namespace Polly.Bulkhead;
 
 /// <summary>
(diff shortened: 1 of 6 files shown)


SimonCropp:
this came about when converting BrokenCircuitException to nullable. since the nullablility of TResult when deserialized must be null

martincostello (approved):
Nothing against the change in principle, though once merged this is a breaking change so means any future patch releases we may need to do before v8 will need to be developed from a branch.

martintmk:
Hey @martincostello @SimonCropp do you think we are able to merge this into V8? I encountered this when exploring the codebase and investigating moving some exception types to `Polly.Core` (`ExecutionRejectedException`).

Would simplify some of the next steps.

martincostello:
Fine with me as we've already merged various v8 changes into `main`.

martincostello:
Closing due to merge conflicts and age.

PR #1025, opened 2023-01-16
Add a Guard and leverage CallerArgumentExpression

SimonCropp:


src/Polly.Specs/Helpers/Bulkhead/AssertionFailure.cs
```diff
@@ -4,7 +4,7 @@ public class AssertionFailure
 {
     public AssertionFailure(int expected, int actual, string measure)
     {
-        if (string.IsNullOrWhiteSpace(measure)) throw new ArgumentNullException(nameof(measure));
+        Guard.NotNull(measure);
```
martincostello, line 7: Did you mean to change this to allow white space? I think it's the only occurrence in the diff, and it's in a test, but thought I'd ask to check.

src/Polly/Guard.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Runtime.CompilerServices;
+
+static class Guard
```
martincostello, line 3: I prefer to be explicit that have ambiguity of maybe someone forgot to add it.

```suggestion
internal static class Guard
```

martincostello:
Closing due to merge conflicts and age.

PR #1026, opened 2023-01-16
support omitting error delegate

SimonCropp:


src/Polly/Caching/AsyncCacheSyntax.cs
```diff
@@ -138,8 +138,8 @@ public static AsyncCachePolicy CacheAsync(IAsyncCacheProvider cacheProvider, ITt
         Action<Context, string> onCacheGet,
         Action<Context, string> onCacheMiss,
         Action<Context, string> onCachePut,
-        Action<Context, string, Exception>? onCacheGetError,
-        Action<Context, string, Exception>? onCachePutError) =>
+        Action<Context, string, Exception>? onCacheGetError = null,
+        Action<Context, string, Exception>? onCachePutError = null) =>
```
martincostello, line 142: Do we want to definitely make this (and others like it in this file) breaking change to the public API?

I know breaking changes can be done for v8, but until that's further along I think it might be better to leave out such changes until we've for a more concrete plan on how breaking v8 will be.

src/Polly/Caching/CacheSyntax.cs
```diff
@@ -140,8 +140,8 @@ public static CachePolicy Cache(ISyncCacheProvider cacheProvider, ITtlStrategy t
         Action<Context, string> onCacheGet,
         Action<Context, string> onCacheMiss,
         Action<Context, string> onCachePut,
-        Action<Context, string, Exception>? onCacheGetError,
-        Action<Context, string, Exception>? onCachePutError) =>
+        Action<Context, string, Exception>? onCacheGetError = null,
+        Action<Context, string, Exception>? onCachePutError = null) =>
```
martincostello, line 144: +1 to breaking changes.

src/Polly/Caching/CacheTResultSyntax.cs
```diff
@@ -146,8 +146,8 @@ public static CachePolicy<TResult> Cache<TResult>(ISyncCacheProvider cacheProvid
         Action<Context, string> onCacheGet,
         Action<Context, string> onCacheMiss,
         Action<Context, string> onCachePut,
-        Action<Context, string, Exception>? onCacheGetError,
-        Action<Context, string, Exception>? onCachePutError)
+        Action<Context, string, Exception>? onCacheGetError = null,
+        Action<Context, string, Exception>? onCachePutError = null)
```
martincostello, line 150: +1 to breaking changes.

src/Polly/Caching/AsyncCacheSyntax.cs
SimonCropp: forgive my ignorance. but how is adding a parameter default considered a breaking change?

martincostello: I thought it came under the same umbrella as with adding new parameters that are optional, which definitely is breaking, but I can't find guidance either way [in this document](https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules#members) that says whether making a parameter optional where it wasn't before is allowed or not.

Without pouring over more documentation, I now don't know whether it is a breaking change or not, which still puts me on the side of caution and not making the change until knowing definitively whether it is or not.

Typically changing almost anything existing on the public surface of a library makes cautious. Maybe this scenario is actually fine and I've just not encountered it before, but without knowing for sure it's safe, I'd err against it.

SimonCropp: > adding new parameters that are optional

this is changing the public api

changing an existing param to be optional is no changing the  public api. the only edge case is if it causes an existing overload to become ambiguous

martincostello: I borrowed Immo's time for a few minutes on the Polly catch-up just now to ask about this, and I was just being over-cautious - it's basically fine for our purposes in the case where a required parameter becomes optional.

martincostello (approved):

PR #1027, opened 2023-01-17
Add static analyzers

martincostello:

Add static analysis packages to identify code issues before peer review.

More diagnostics need to be fixed/suppressed yet, and the default rules/preferences need to be configured for rules we're not concerned by or don't agree with the defaults of.

Some of the diagnostics seem to suggest breaking changes, so they'll need suppressing where relevant too.

See https://github.com/App-vNext/Polly/pull/1019#discussion_r1071379092 for context.

### DIFF
--- a/src/Polly/Polly.csproj
+++ b/src/Polly/Polly.csproj
@@ -18,6 +18,16 @@
     <ContinuousIntegrationBuild Condition=" '$(CI)' != '' ">true</ContinuousIntegrationBuild>
     <Deterministic>true</Deterministic>
   </PropertyGroup>
+  <PropertyGroup Label="Analyzers">
+    <NoWarn>$(NoWarn);S4457</NoWarn>
+  </PropertyGroup>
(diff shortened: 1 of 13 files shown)


src/Polly/Polly.csproj
```diff
@@ -18,6 +18,16 @@
     <ContinuousIntegrationBuild Condition=" '$(CI)' != '' ">true</ContinuousIntegrationBuild>
     <Deterministic>true</Deterministic>
   </PropertyGroup>
+  <PropertyGroup Label="Analyzers">
```
martintmk, line 21: Do you think it's possible to extract some of these pieces that will be reused across multiple projects into `Directory.Build.props` so we don't have to refactor this later?

martincostello: Yep, but I'm just starting here because the test projects created a lot of noise.

martincostello:
This has mostly been implemented through other PRs now.

PR #1031, opened 2023-01-23
Introduce mutation testing

martintmk:

This PR adds the mutation testing that verifies the quality of the unit tests written.

See https://stryker-mutator.io/ for more details.

No new tests added, just introdution of mutation gate. The current mutation score is 72, and new PRs that add features cannot fall below this threshold. Over time the score should increase by adding more through tests. 

### Confirm the following

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,6 +52,12 @@ jobs:
         files: ./src/Polly.Specs/coverage-reports/Cobertura.xml,./src/Polly.Core.Tests/coverage-reports/Cobertura.xml
         flags: ${{ matrix.os_name }}
 
+    - name: Upload Mutation Report
+      uses: actions/upload-artifact@v3
+      with:
+        name: mutation-report-${{ matrix.os_name }}
+        path: StrykerOutput
+
     - name: Publish NuGet packages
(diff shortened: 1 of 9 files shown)


martincostello:
I'm on vacation until Monday, I'll take a look then.

.github/workflows/build.yml
```diff
@@ -52,6 +52,11 @@ jobs:
         file: ./src/Polly.Specs/coverage-reports/Cobertura.xml
         flags: ${{ matrix.os_name }}
 
+    - uses: actions/upload-artifact@v3
+      with:
+        name: mutation-report
```
martincostello, line 57: Should we make this name OS-specific so that we know exactly which leg uploaded the report? Or we could only upload it from one OS.

image

martintmk: I think it's ok for just picking one, for example (windows + NET7.0). We do the same for our projects. Otherwise, too many reports might overwhelm the users.

martintmk: @martincostello In the end I decided to upload all reports since we are executing the mutation testing on them anyway.

martincostello (approved):

PR #1056, opened 2023-03-15
Introduce IResilienceStrategy and core primitives for V8

martintmk:

In this PR I am transferring some core primitives from v8-poc branch. This includes:

- Adding documentation. 
- Adding code-guards.
- Polishing and improving the API.

To keep thing simple I am omitting the following:

- Pooling of `ResilienceContext` instances.
- Additional `Execute` extensions as there will be many of these (to be done in follow-up).

I'll include the tests once we agree on API shape.

### Confirm the following


src/Polly.Core/ResilienceContext.cs
```diff
@@ -0,0 +1,90 @@
...
+
+    /// <summary>
+    /// Gets or sets the <see cref="CancellationToken"/> associated with the execution.
+    /// </summary>
+    public CancellationToken CancellationToken { get; set; }
+
+    /// <summary>
+    /// Gets a value indicating whether the execution is synchronous.
+    /// </summary>
+    public bool IsSynchronous { get; private set; }
```
martintmk, line 24: These properties are set by the `Execute` extensions and should not be mutable across the execution.

src/Polly.Core/ResilienceContext.cs
```diff
@@ -0,0 +1,90 @@
...
+
+    /// <summary>
+    /// Gets the type of the result associated with the execution.
+    /// </summary>
+    public Type ResultType { get; private set; } = typeof(UnknownResult);
```
martintmk, line 29: I was considering wrapping these 4 properties:

- `ResultType`
- `IsVoid`
- `ContinueOnCapturedContext`
- `IsSynchronous`

Into something like `IExecutionMode` and exposing it as `ResilienceContext.ExecutionMode`. However, it felt it does not add much value as we will be passing `ResilienceContext` everywhere anyway.

geeknoid:
Note that the default object pool in .NET has terrible scaling characteristics in scenarios where many pooled objects exist and are accessed in a concurrent fashion. The default pool was originally designed for very short-term object rentals and it just doesn't work well for long-term rentals.

I fixed this for .NET 8. In the meantime, you may wish to grab the .NET 8 implementation here and use it instead of the default .NET object pool.

https://github.com/dotnet/aspnetcore/blob/main/src/ObjectPool/src/DefaultObjectPool.cs

martintmk:
> 

My thinking is just to copy that code here since the revisited version is pretty simple. And we will avoid extra dependency.

src/Polly.Core/NullResilienceStrategy.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// Gets the singleton instance of the <see cref="NullResilienceStrategy"/>.
+    /// </summary>
+    public static readonly NullResilienceStrategy Instance = new();
+
+    private NullResilienceStrategy()
+    {
+    }
+
+    /// <inheritdoc/>
+    public ValueTask<T> ExecuteCoreAsync<T, TState>(Func<ResilienceContext, TState, ValueTask<T>> execution, ResilienceContext context, TState state)
```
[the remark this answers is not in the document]
joelhulen: I don't know if the key word 'core' denotes an indirect method compared to something like 'internal'. Perhaps we can call it `ExecuteInternalAsync`?

joelhulen (approved):

src/Polly.Core/NullResilienceStrategy.cs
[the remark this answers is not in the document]
martintmk: I gave this a shot. What do you think about:

- making `IResilienceStrategy.ExecuteInternalAsync` so it won't be even visible anywhere.
- only Execute/ExecuteAsync extensions are visible
- the custom strategies need to derive from `DelegatingResilienceStrategy` anyway where `ExecuteCoreAsync` can be overriden
 
The only downside is the reduced direct mockability of `IResilienceStrategy` but anyone can use `DelegatingResilienceStrategy` or `NullResilienceStrategy` in tests.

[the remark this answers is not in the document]
joelhulen: Actually making it internal is even better, in my opinion. I think the key here is to be opinionated about how to execute these strategies, which is to do so through extensions. The `NullResilienceStrategy` could be used almost like a noop in tests, I assume.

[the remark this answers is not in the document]
martintmk: I'll go ahead with this API and start writing tests tomorrow.

src/Polly.Core/ResilienceContext.cs
```diff
@@ -0,0 +1,90 @@
+namespace Polly;
+
+/// <summary>
+/// Context that carries with a single execution through a <see cref="IResilienceStrategy"/>.
```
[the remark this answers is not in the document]
martintmk: Updated the description and added more details.

src/Polly.Core/NullResilienceStrategy.cs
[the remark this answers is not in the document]
martintmk: Done, 100% coverage and mutation score.

src/Polly.Core/IResilienceStrategy.cs
```diff
@@ -0,0 +1,22 @@
...
+/// <summary>
+/// A resilience strategy that is used to execute the user provided execution callbacks.
+/// </summary>
+/// <remarks>
+/// You do not execute any methods on this interface directly. Instead, use the dedicated extensions defined in <see cref="ResilienceStrategyExtensions"/> to execute your callbacks.
+/// </remarks>
+public interface IResilienceStrategy
+{
+    /// <summary>
+    /// Executes the <paramref name="execution"/> callback.
```
martincostello, line 12:
```suggestion
    /// Executes the specified callback.
```

src/Polly.Core/IResilienceStrategy.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// </summary>
+    /// <typeparam name="TResult">The type of result returned by the execution callback.</typeparam>
+    /// <typeparam name="TState">The type of state associated with the execution.</typeparam>
+    /// <param name="execution">The execution callback.</param>
+    /// <param name="context">The context associated with the execution.</param>
+    /// <param name="state">The state associated with the execution.</param>
+    /// <returns>An instance of <see cref="ValueTask"/> that represents an asynchronous execution.</returns>
+    /// <remarks>This method should not be used directly. Instead, use the dedicated extensions to execute the user provided callback.</remarks>
```
martincostello, line 20:
```suggestion
    /// <remarks>This method should not be used directly. Instead, use the dedicated extensions to execute the user-provided callback.</remarks>
```

src/Polly.Core/NullResilienceStrategy.cs
```diff
@@ -0,0 +1,27 @@
+namespace Polly;
+
+/// <summary>
+/// A resilience strategy that just executes the execution callback without any additional logic.
```
martincostello, line 4:
```suggestion
/// A resilience strategy that executes an execution callback without any additional logic.
```

src/Polly.Core/ResilienceContext.cs
```diff
@@ -0,0 +1,102 @@
+using System.Diagnostics.CodeAnalysis;
+
+namespace Polly;
+
+/// <summary>
+/// An unique assigned to a single execution of <see cref="IResilienceStrategy"/>. It is created manually or automatically
```
martincostello, line 6: I don't follow what _"An unique assigned to a"_ is referring to. Seems like it's maybe missing a word?

src/Polly.Core/ResilienceContext.cs
```diff
@@ -0,0 +1,102 @@
...
+/// when the user calls the various extensions on top of <see cref="IResilienceStrategy"/>. After every execution the context  should be discarded and returned to the pool.
```
martincostello, line 7:
```suggestion
/// when the user calls the various extensions on top of <see cref="IResilienceStrategy"/>. After every execution the context should be discarded and returned to the pool.
```

src/Polly.Core/ResilienceContext.cs
```diff
@@ -0,0 +1,102 @@
...
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the execution should continue on the captured context.
+    /// </summary>
+    public bool ContinueOnCapturedContext { get; set; }
+
+    /// <summary>
+    /// Gets a value indicating whether the context is initialized.
+    /// </summary>
+    internal bool IsInitialized => ResultType != typeof(UnknownResult);
```
martincostello, line 49: `is not` rather than `!=`?

src/Polly.Core/NullResilienceStrategy.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+    }
+
+    /// <inheritdoc/>
+    ValueTask<TResult> IResilienceStrategy.ExecuteInternalAsync<TResult, TState>(Func<ResilienceContext, TState, ValueTask<TResult>> execution, ResilienceContext context, TState state)
+    {
+        Guard.NotNull(execution);
+        Guard.NotNull(context);
+
+        context.AssertInitialized();
```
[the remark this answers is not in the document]
martincostello: Oh, I see now that this is just a DEBUG-only assertion.

src/Polly.Core/ResilienceStrategyExtensions.Async.Task.cs
```diff
@@ -0,0 +1,72 @@
+using System.Threading;
+using Polly;
+
+namespace Polly;
+
+public static partial class ResilienceStrategyExtensions
+{
+    /// <summary>
+    /// Executes the <paramref name="execution"/> callback.
```
martincostello, line 9:
```suggestion
    /// Executes the specified callback.
```

src/Polly.Core/ResilienceStrategyExtensions.Async.Task.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// </summary>
+    /// <typeparam name="TState">The type of state associated with the execution.</typeparam>
+    /// <param name="strategy">Instance of <see cref="IResilienceStrategy"/>.</param>
+    /// <param name="execution">The execution callback.</param>
+    /// <param name="context">The context associated with the execution.</param>
+    /// <param name="state">The state associated with the execution.</param>
+    /// <returns>The instance of <see cref="Task"/> that represents the asynchronous execution.</returns>
```
martincostello, line 16: Should we also document the exceptions that would be thrown by methods using the `Guard` methods?

src/Polly.Core/ResilienceStrategyExtensions.Async.Task.cs
```diff
@@ -0,0 +1,72 @@
...
+        TState state)
+    {
+        Guard.NotNull(strategy);
+        Guard.NotNull(execution);
+        Guard.NotNull(context);
+
+        InitializeAsyncContext(context);
+
+        await strategy.ExecuteInternalAsync(
+            static async (c, s) =>
```
martincostello, line 30: Could we use more meaningful names here than `c` and `s`? With this being a new API and me not looking at it in an IDE, it isn't super clear to me what the two things are (I'm assuming `context` and `strategy`?).

src/Polly.Core/ResilienceStrategyExtensions.Async.Task.cs
```diff
@@ -0,0 +1,72 @@
...
+            {
+                await s.execution(c, s.state).ConfigureAwait(c.ContinueOnCapturedContext);
+                return VoidResult.Instance;
+            },
+            context,
+            (execution, state)).ConfigureAwait(context.ContinueOnCapturedContext);
+    }
+
+    /// <summary>
+    /// Executes the <paramref name="execution"/> callback.
```
martincostello, line 40:
```suggestion
    /// Executes the specified callback.
```

src/Polly.Core/ResilienceStrategyExtensions.Async.Task.cs
martincostello: Ah I see in the method after they're `(context, state)`.

src/Polly.Core/ResilienceStrategyExtensions.Async.Task.cs
martincostello: There's a few of these, I won't comment on them all.

src/Polly.Core/ResilienceStrategyExtensions.Async.ValueTask.cs
```diff
@@ -0,0 +1,81 @@
+using System.Threading;
+using Polly;
+
+namespace Polly;
+
+/// <summary>
+/// The extensions for <see cref="IResilienceStrategy"/>. These extensions are used to execute the user callbacks in various execution modes.
```
martincostello, line 7:
```suggestion
/// Extensions for <see cref="IResilienceStrategy"/>. These extension methods are used to execute user callbacks in various execution modes.
```

src/Polly.Core/ResilienceStrategyExtensions.Async.ValueTask.cs
```diff
@@ -0,0 +1,81 @@
...
+/// For example, you can use the same <see cref="IResilienceStrategy"/> instance to execute a synchronous callback,
+/// an asynchronous callback, or a callback that returns a void result.
+/// </summary>
+public static partial class ResilienceStrategyExtensions
+{
+    /// <summary>
+    /// Executes the <paramref name="execution"/> callback.
+    /// </summary>
+    /// <typeparam name="TState">The type of s associated with the execution.</typeparam>
```
martincostello, line 16: `state` rather than `s`?

src/Polly.Core/ResilienceStrategyExtensions.Async.ValueTask.cs
```diff
@@ -0,0 +1,81 @@
...
+    /// <param name="strategy">Instance of <see cref="IResilienceStrategy"/>.</param>
+    /// <param name="execution">The execution callback.</param>
+    /// <param name="context">The c associated with the execution.</param>
+    /// <param name="state">The s associated with the execution.</param>
```
martincostello, line 20: Same here for the `c` and the `s`.

src/Polly.Core/ResilienceStrategyExtensions.Async.ValueTask.cs
martincostello: Same comments in the other places this comes up rather than adding too much comment noise.

src/Polly.Core.Tests/Utils/SynchronousExecutionHelperTests.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Threading.Tasks;
+using FluentAssertions;
+using Polly.Utils;
+using Xunit;
+
+namespace Polly.Core.Tests.Utils;
+
+public class SynchronousExecutionHelperTests
+{
+    public SynchronousExecutionHelperTests() => Trace.Listeners.Clear();
```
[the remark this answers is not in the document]
martintmk: Is there a day to do this globally? We don't want listeners in tests as it causes failures when running tests locally in debug mode.

src/Polly.Core/ResilienceStrategyExtensions.Async.ValueTask.cs
martintmk: True, done for all extensions.

src/Polly.Core/ResilienceStrategyExtensions.Async.ValueTask.cs
martintmk: Thanks for spotting this. Fixed.

src/Polly.Core/ResilienceContext.cs
martintmk: Added "context" there

src/Polly.Core/ResilienceStrategyExtensions.Async.Task.cs
martintmk: Done.

src/Polly.Core/ResilienceStrategyExtensions.Async.Task.cs
martintmk: I am split on this. I feel it just adds noise to the documentation. If you think we need this I can add it everywhere.

src/Polly.Core/Utils/Guard.cs
```diff
@@ -0,0 +1,17 @@
...
+
+internal static class Guard
+{
+    public static T NotNull<T>(T value, [CallerArgumentExpression("value")] string argumentName = "")
+        where T : class
+    {
+        if (value is null)
+        {
+            throw new ArgumentNullException(argumentName);
+        }
```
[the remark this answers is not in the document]
joelhulen: I agree with keeping it as it is.

src/Polly.Core/ResilienceStrategyExtensions.Async.Task.cs
martincostello: I'm happy to not, I just thought I'd suggest it.

src/Polly.Core.Tests/Utils/SynchronousExecutionHelperTests.cs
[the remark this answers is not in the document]
martintmk: It's just that the first test clears this collection and others are no-op. Should cause no problems.

PR #1063, opened 2023-03-20
Drop DelegatingResilienceStrategy

martintmk:

When writing additional tests, I realized that the `DelegatingResilienceStrategy` doesn't work really well when you want to share the `ResilienceStrategy` across different pipelines. The problem is with the mutable `Next` property. It also creates ambiguity when implementing a new strategy as we now have two base classes to derive from. 

To fix this problem and simplify the API we are dropping `DelegatingResilienceStrategy` and making it an internal implementation detail that is used when we are linking multiple strategies together. 

This simplifies the API surface and also removes some infra around freezing the pipelines which won't be needed anymore.

The downside is slightly slower execution; however, we should focus on clean API first.

### Confirm the following


src/Polly.Core/Builder/ResilienceStrategyPipeline.cs
```diff
@@ -1,55 +1,46 @@
...
-            }
-        }).ToList();
+        // convert all strategies to delegating ones (except the last one as it's not required)
+        var delegatingStrategies = strategies
+            .Take(strategies.Count - 1)
+            .Select(strategy => new DelegatingResilienceStrategy(strategy))
+            .ToList();
 
+        // link the last one
+        delegatingStrategies.Last().Next = strategies[strategies.Count - 1];
```
martincostello, line 36: Can we use indexer over LINQ here?

```suggestion
        delegatingStrategies[^1].Next = strategies[strategies.Count - 1];
```

martintmk: Dropped the ling and on .NET 6+ used the simpler syntax

martincostello (approved):

Issue #1066, opened 2023-03-20
Infrastructure for telemetry


martintmk:
We want to expose an infrastructure for telemetry. This way, individual resilience strategies can report important events in a unified way. The Polly consumers can hook to these events and build monitoring and dashboard on top.

We can consider something like:

``` csharp
public abstract class ResilienceTelemetry
{
    void Report(string eventName, ResilienceContext context);

    void Report<T>(string eventName, T result, ResilienceContext context);

    void Report(string eventName, Exception exception, ResilienceContext context);
}

public abstract class ResilienceTelemetryFactory
{
    ResilienceTelemetry Create(ResilienceTelemetryContext context);
}

public class ResilienceStrategyBuilderContext
{
    // new property
    public ResilienceTelemetry Telemetry { get; }
}

public class ResilienceStrategyBuilderOptions
{
    // new property
    public ResilienceTelemetryFactory TelemetryFactory { get; set; } = NullResilienceTelemetryFactory.Instance;
}
```

Customer can provide their own implementation of `ResilienceTelemetryFactory`. We can consider having built-in `ResilienceTelemetryFactory` based on [.NET Meter](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics-instrumentation) and `ILoggerFactory`. 

The `ResilienceTelemetryFactory.Create` is called for each resilience strategy individually and with different context.

martintmk:
The API was simplified in #1112 and it is now using `DiagnosticSource` internally.

PR #1073, opened 2023-03-21
Introduce ResilienceTelemetryFactory

martintmk:

#1066 

This API exposes simple `ResilienceTelemetry` API that the `ResilienceStrategy` authors can use to report important events in a unified way. For example:

``` csharp
ResilienceTelemetry telemetry = ...;

telemetry.Report("OnTimeout", resilienceContext);
 
telemetry.Report("OnRetry", failedResult, resilienceContext);
 
telemetry.ReportException("OnRetry", failedException, resilienceContext);
```

The consumers of `ResilienceStrategy` can provide their own implementation of `ResilienceTelemetryFactory` to provide additional telemetry with each event (extracted from `ResilienceContext` or from reported result or exception).

By default, we use `NullResilienceTelemetryFactory` that does nothing. At later stages we might consider introducing `Polly.Telemetry` package that implements `ResilienceTelemetryFactory` and does logging using `ILoggerFactory` and metering with the new `Meter` API.

### Confirm the following

### DIFF
--- a/src/Polly.Core/Builder/ResilienceStrategyBuilderOptions.cs
+++ b/src/Polly.Core/Builder/ResilienceStrategyBuilderOptions.cs
@@ -1,4 +1,5 @@
 using System.ComponentModel.DataAnnotations;
+using Polly.Telemetry;
 
 namespace Polly.Builder;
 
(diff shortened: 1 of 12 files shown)


src/Polly.Core/Builder/ResilienceStrategyBuilderOptions.cs
```diff
@@ -18,4 +19,10 @@ public class ResilienceStrategyBuilderOptions
     /// Gets the custom properties attached to builder options.
     /// </summary>
     public ResilienceProperties Properties { get; } = new();
+
+    /// <summary>
+    /// Gets or sets an instance of <see cref="TelemetryFactory"/>.
+    /// </summary>
+    [Required]
```
martincostello, line 26: Is this required if it has a default implementation?

martintmk: Just in case someone set's it to `null` explicitly.

martincostello (approved):

PR #1075, opened 2023-03-21
Polly now references Polly.Core

martintmk:
### The issue or feature being addressed

We will be moving some types `from Polly` to `Polly.Core` so Polly needs reference the core project. As a part of this we are also dropping support for legacy `netstandard1.1` target.

### Confirm the following

- [x]  I started this PR by branching from the head of the default branch
- [x]  I have targeted the PR to merge into the default branch
- [x]  I have included unit tests for the issue/feature
- [x]  I have successfully run a local build

PR #1076, opened 2023-03-22
Fix coverage upload

martincostello:
- Output a report in Cobertura format for coverlet.
- Write the reports to a stable directory path for coverlet upload.
- Build and test in the order of the dependency tree since #1075.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
     - uses: codecov/codecov-action@v3
       name: Upload coverage to Codecov
       with:
-        files: ./artifacts/coverage-reports/Polly.Core.Tests/Cobertura.xml,./artifacts/coverage-reports/Polly.Specs/Cobertura.xml
+        files: ./artifacts/coverage-reports/**/Cobertura.xml
         flags: ${{ matrix.os_name }}
 
     - name: Upload Mutation Report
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -49,7 +49,7 @@ jobs:
     - uses: codecov/codecov-action@v3
       name: Upload coverage to Codecov
       with:
-        files: ./src/Polly.Specs/coverage-reports/Cobertura.xml,./src/Polly.Core.Tests/coverage-reports/Cobertura.xml
+        files: ./artifacts/coverage-reports/Polly.Core.Tests/Cobertura.xml,./artifacts/coverage-reports/Polly.Specs/Cobertura.xml
```
martintmk, line 52: We will have more projects later, is it possible to have a mask here?

martincostello: I did try a directory at first https://github.com/App-vNext/Polly/pull/1076/commits/83c9f1df1f7403597ffdafca54783c9f32328e7b, but it [didn't seem to work](https://github.com/App-vNext/Polly/actions/runs/4488732817/jobs/7893695475#step:5:29).

I'll try using a wildcard and see if that works.

martincostello: Looks like that [didn't work](https://github.com/App-vNext/Polly/actions/runs/4490361008/jobs/7897328721?pr=1076#step:5:27) 😞

I'll revert to the previous change and we can look to improve it in the future.

PR #1078, opened 2023-03-23
Always attempt to upload mutation report

martintmk:

We have seen mutants that were not killed in some builds. Because the mutation report was not uploaded, we couldn't determine the cause. With this change, we always try to upload the mutation reports.

### Confirm the following

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,12 +51,15 @@ jobs:
       with:
         files: ./artifacts/coverage-reports/Polly.Core.Tests/Cobertura.xml,./artifacts/coverage-reports/Polly.Specs/Cobertura.xml
         flags: ${{ matrix.os_name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -51,12 +51,15 @@ jobs:
       with:
         files: ./artifacts/coverage-reports/Polly.Core.Tests/Cobertura.xml,./artifacts/coverage-reports/Polly.Specs/Cobertura.xml
         flags: ${{ matrix.os_name }}
+        if-no-files-found: warn
```
martincostello, line 54: This is the default, so we can omit this unless you want it to not do anything or fail the build.

martintmk: Thanks, cleaned this up.

martincostello (approved):

Issue #1071, opened 2023-03-20
Migrate the Timeout Policy


martintmk:
Migrate the [Timeout Policy](https://github.com/App-vNext/Polly/tree/main/src/Polly/Timeout) to V8 API.

**Prerequisites** 
#1070 
#1067

The details on how to introduce the new resilience strategies for V8 are [here](https://github.com/App-vNext/Polly/blob/main/src/Polly.Core/README.md#extensibility).

The steps in general:

- Move exceptions and enums and classes that are shared between V7 and V8 into `Polly.Core`
- Use [Type Forwarding](https://learn.microsoft.com/en-us/dotnet/standard/assembly/type-forwarding) so the consumers still using V7 won't get runtime errors. 
- Introduce `TimeoutOptions`
- Expose new extensions for the timeout strategy. 
- Forward the V7 implementation to V8.

**API**

``` csharp
public class TimeoutStrategyOptions
{
   public TimeoutStrategy Strategy { get; set; }

   public Event<OnTimeoutArguments> OnTimeout { get; set; }

   public Generator<TimeoutGeneratorArguments, TimeSpan> TimeoutGenerator { get; set; }
}

public readonly struct OnTimeoutArguments
{
   public ResilienceContext ResilienceContext { get; }
   public TimeSpan ExecutionTime { get; }
}

public readonly struct TimeoutGeneratorArguments
{
   public ResilienceContext ResilienceContext { get; }
}

public static class TimeoutResilienceStrategyBuilderExtensions
{
   ResilienceStrategyBuilder AddTimeout(this ResilienceStrategyBuilder builder, TimeoutStrategyOptions options);
   ResilienceStrategyBuilder AddTimeout(this ResilienceStrategyBuilder builder, TimeSpan timeout);
   ResilienceStrategyBuilder AddTimeout(this ResilienceStrategyBuilder builder, TimeSpan timeout, TimeoutStrategy  strategy);

   // possibly more
}
```

**API Usage**

``` csharp

// advanced setup
var options = new TimeoutStrategyOptions();

// add an event that's called when timeout occurrs
options.OnTimeout.Add(args => Console.WriteLine("Timeout occurred!"))

// add callback to result generator, you can return different timeouts based on the result type or data from resilience context
options.TimeoutGenerator.Add(args => TimeSpan.FromSeconds(5));
var strategy = new ResilienceStrategyBuilder().AddTimeout(options).Build();

// simple setup
[...]

Issue #1068, opened 2023-03-20
Introduce the pooling for ResilienceContext


martintmk:
The `ResilienceContext.Get` and `ResilienceContext.Return` should use object pool to reduce the allocations. 

We can just copy the .NET implementation to Polly to avoid unnecessary dependencies:
https://github.com/dotnet/aspnetcore/blob/main/src/ObjectPool/src/DefaultObjectPool.cs

PR #1079, opened 2023-03-23
Introduce Timeout Strategy

martintmk:

#1071 

General notes:

- We recognize two types of callbacks for timeout. 
- The `TimeoutGenerator` provides the timeout for a given execution. 
- The `OnTimeoutEvent` allows registering multiple callbacks that are called when timeout happens. 
- Each callback has associated "Arguments" class. Adding new members to it won't break existing adopters.
- No pooling of cancellation token sources yet (depends on #1068).]
- I opted for introducing the classes for callback as it allows us to evolve the API without breaking changes (as opposed if just raw callbacks were used).
- Only optimistic timeout initially. The pessimistic to be done later.

API Usage:

``` csharp
// simple
builder.AddTimeout(TimeSpan.FromSeconds(10));

// simple with callback
builder.AddTimeout(TimeSpan.FromSeconds(10), args => Console.WriteLine("Timeout occurred: {0}", args.Timeout));

// using options
builder.AddTimeout(new TimeoutStrategyOptions
{
    TimeoutGenerator = new TimeoutGenerator(TimeSpan.FromSeconds(10)),
    OnTimeout = new OnTimeoutEvent()
        .Add(args => Console.WriteLine("Timeout occurred 1: {0}", args.Timeout))
        .Add(args => Console.WriteLine("Timeout occurred 2: {0}", args.Timeout))
});
```

### Confirm the following


src/Polly.Core/ExecutionRejectedException.cs
```diff
@@ -8,7 +8,7 @@ namespace Polly;
 /// Exception thrown when a policy rejects execution of a delegate.
 /// <remarks>More specific exceptions which derive from this type, are generally thrown.</remarks>
 /// </summary>
-public abstract class ExecutionRejectedException : Exception
+public class ExecutionRejectedException : Exception
```
martincostello, line 11: Just curious, what's the benefit to a non-`abstract` class if there's no way to construct one without deriving from it anyway?

src/Polly.Core/Timeout/OnTimeoutEvent.cs
```diff
@@ -0,0 +1,95 @@
...
+    public OnTimeoutEvent Add(Func<OnTimeoutArguments, ValueTask> onTimeout)
+    {
+        Guard.NotNull(onTimeout);
+
+        _events.Add(onTimeout);
+        return this;
+    }
+
+    /// <summary>
+    /// Adds an synchronous on-timeout callback.
```
martincostello, line 25:
```suggestion
    /// Adds a synchronous on-timeout callback.
```

src/Polly.Core/Timeout/TimeoutGenerator.cs
```diff
@@ -0,0 +1,61 @@
+namespace Polly.Timeout;
+
+/// <summary>
+/// Timeout generates the timeout for a given execution of user-provided callback.
```
martincostello, line 4:
```suggestion
/// Generator for the timeout for a given execution of a user-provided callback.
```

src/Polly.Core/Timeout/TimeoutResilienceStrategy.cs
```diff
@@ -0,0 +1,108 @@
...
+        var previousToken = context.CancellationToken;
+        using var cancellationSource = new CancellationTokenSource();
+        _timeProvider.CancelAfter(cancellationSource, timeout);
+        context.CancellationToken = cancellationSource.Token;
+
+        CancellationTokenRegistration? registration = null;
+
+        if (previousToken.CanBeCanceled)
+        {
+            registration = previousToken.Register(static state => ((CancellationTokenSource)state!).Cancel(), cancellationSource, useSynchronizationContext: false);
```
martincostello, line 42: Is it safe to assume `state` is always a `CancellationToken` here?

src/Polly.Core/Timeout/TimeoutResilienceStrategy.cs
```diff
@@ -0,0 +1,108 @@
...
+
+            return await callback(context, state).ConfigureAwait(context.ContinueOnCapturedContext);
+        }
+        catch (OperationCanceledException e) when (cancellationSource.IsCancellationRequested && !previousToken.IsCancellationRequested)
+        {
+            _telemetry.Report(TimeoutConstants.OnTimeoutEvent, context);
+
+            await _onTimeout(new OnTimeoutArguments(context, e, timeout)).ConfigureAwait(context.ContinueOnCapturedContext);
+
+            throw new TimeoutRejectedException("The timeout resilience strategy did not complete within the timeout.", e);
```
martincostello, line 60: Would it be useful to pass the timeout value through to the exception for inspection by any code that catches it?

src/Polly.Core/Timeout/TimeoutResilienceStrategy.cs
```diff
@@ -0,0 +1,108 @@
...
+            return await completedTask.ConfigureAwait(context.ContinueOnCapturedContext);
+        };
+    }
+
+    private static Task<TResult> AsTask<TResult>(CancellationToken cancellationToken)
+    {
+        var tcs = new TaskCompletionSource<TResult>();
+
+        // A generalized version of this method would include a hot-path returning a canceled task (rather than setting up a registration)
+        // if (cancellationToken.IsCancellationRequested) on entry.  This is omitted, since we only start the timeout countdown in the token _after calling this method.
```
martincostello, line 95:
```suggestion
        // if (cancellationToken.IsCancellationRequested) on entry.  This is omitted, since we only start the timeout countdown in the token after calling this method.
```

src/Polly.Core/Timeout/TimeoutResilienceStrategyBuilderExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+/// <summary>
+/// Extension methods for adding timeouts to <see cref="ResilienceStrategyBuilder"/>.
+/// </summary>
+public static class TimeoutResilienceStrategyBuilderExtensions
+{
+    /// <summary>
+    /// Adds a timeout resilience strategy to the builder.
+    /// </summary>
+    /// <param name="builder">The builder instance.</param>
+    /// <param name="timeout">The timeout value. This value should be greater than <see cref="TimeSpan.Zero"/> and less than <see cref="System.Threading.Timeout.InfiniteTimeSpan"/>.</param>
```
martincostello, line 15: Should we validate this, rather than just give guidance?

src/Polly.Core/Timeout/TimeoutStrategy.cs
```diff
@@ -0,0 +1,18 @@
...
+/// </summary>
+public enum TimeoutStrategy
+{
+    /// <summary>
+    /// An optimistic <see cref="TimeoutStrategy"/>.  The timeout resilience strategy relies on a timing-out <see cref="CancellationToken"/> to cancel executed delegates by co-operative cancellation.
+    /// </summary>
+    Optimistic,
+
+    /// <summary>
+    /// An pessimistic <see cref="TimeoutStrategy"/>.  The timeout resilience strategy will assume the delegates passed to be executed will not necessarily honor
```
martincostello, line 14:
```suggestion
    /// A pessimistic <see cref="TimeoutStrategy"/>.  The timeout resilience strategy will assume the delegates passed to be executed will not necessarily honor
```

src/Polly.Core/Timeout/OnTimeoutEvent.cs
```diff
@@ -0,0 +1,95 @@
...
+        };
+    }
+
+    private static Func<OnTimeoutArguments, ValueTask> CreateHandler(Func<OnTimeoutArguments, ValueTask>[] callbacks)
+    {
+        return async args =>
+        {
+            foreach (var callback in callbacks)
+            {
+                await callback(args).ConfigureAwait(args.Context.ContinueOnCapturedContext);
```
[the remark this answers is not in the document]
martintmk: For event callbacks we probably want to execute all that are registered. 

This becomes interesting for retries and the `ShouldRetry`, `RetryDelayGenerator` and `RetriesGenerator` callback where it becomes more complicated.

Shall we stop after first `true` value is returned? 
What if another callback explicitly says no-retry? 
What's the priority of callbacks?

Dunno, with timeout I took the easy way, but for retries we will have to think about the behavior more :)

src/Polly.Core/Timeout/TimeoutResilienceStrategy.cs
martintmk: yup, will be always `CancellationTokenSource`

src/Polly.Core/Timeout/TimeoutResilienceStrategy.cs
martintmk: You mean to include it in the message? Or add a `Timeout` property there?

src/Polly.Core/Timeout/TimeoutResilienceStrategyBuilderExtensions.cs
martintmk: Yup definitely, if we are OK with the API shape I'll cover all edge-cases.

src/Polly.Core/Timeout/TimeoutResilienceStrategy.cs
martincostello: Both I guess 😄

src/Polly.Core/ExecutionRejectedException.cs
martincostello: That's weird IMHO as having the keyword is more explicit about intent, but I'm not going to die on the hill of not having it vs. suppressing it.

martincostello:
The shape of the API seems good to me.

src/Polly.Core/Timeout/TimeoutStrategyOptions.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// <summary>
+    /// Gets or sets the timeout event that notifies the timeout occurred.
+    /// </summary>
+    [Required]
+    public OnTimeoutEvent OnTimeout { get; set; } = new();
+
+    /// <summary>
+    /// Gets or sets the timeout strategy.
+    /// </summary>
```
geeknoid, line 30: Please note what the default is for this property.

src/Polly.Core/Timeout/TimeoutResilienceStrategyBuilderExtensions.cs
```diff
@@ -0,0 +1,100 @@
+using System;
+using Polly.Builder;
+
+namespace Polly.Timeout;
+
+/// <summary>
+/// Extension methods for adding timeouts to <see cref="ResilienceStrategyBuilder"/>.
```
geeknoid, line 7:
```suggestion
/// Extension methods for adding timeouts to a <see cref="ResilienceStrategyBuilder"/>.
```

src/Polly.Core/Strategy/IResilienceArguments.cs
```diff
@@ -0,0 +1,12 @@
+namespace Polly.Strategy;
+
+/// <summary>
+/// The recommended shape for arguments used by individual strategies.
+/// </summary>
+public interface IResilienceArguments
```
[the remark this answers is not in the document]
martintmk: The idea was to enforce some kind of common shape since we now ever argument class will have the `Context` property. By having public interface, we can have unified docs by using `<inheritdoc/>` thingy. 

This interface won't be used anywhere directly, just as a marker.

src/Polly.Core/Timeout/TimeoutResilienceStrategy.cs
```diff
@@ -0,0 +1,108 @@
...
+    private static Func<ResilienceContext, TState, ValueTask<TResult>> MakeCancellable<TState, TResult>(Func<ResilienceContext, TState, ValueTask<TResult>> callback)
+    {
+        return async (context, state) =>
+        {
+            var timeoutTask = AsTask<TResult>(context.CancellationToken);
+            var actionTask = callback(context, state).AsTask();
+
+            var completedTask = await Task.WhenAny(actionTask, timeoutTask).ConfigureAwait(context.ContinueOnCapturedContext);
+
+            return await completedTask.ConfigureAwait(context.ContinueOnCapturedContext);
```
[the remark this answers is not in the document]
martintmk: Does not work because of Task -> ValueTask conversion.

src/Polly.Core/ExecutionRejectedException.cs
martintmk: Made it abstract again and relaxed the rule. Felt weird to me too.

src/Polly.Core/Timeout/TimeoutResilienceStrategyBuilderExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// <summary>
+    /// Adds a timeout resilience strategy to the builder.
+    /// </summary>
+    /// <param name="builder">The builder instance.</param>
+    /// <param name="timeout">The timeout value. This value should be greater than <see cref="TimeSpan.Zero"/> and less than <see cref="System.Threading.Timeout.InfiniteTimeSpan"/>.</param>
+    /// <returns>The same builder instance.</returns>
+    public static ResilienceStrategyBuilder AddTimeout(this ResilienceStrategyBuilder builder, TimeSpan timeout)
+    {
+        Guard.NotNull(builder);
+
```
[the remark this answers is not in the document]
martintmk: Done.

src/Polly.Core/Timeout/TimeoutResilienceStrategyBuilderExtensions.cs
martintmk: Done.

src/Polly.Core/Timeout/TimeoutStrategyOptions.cs
```diff
@@ -16,6 +16,9 @@ public class TimeoutStrategyOptions : ResilienceStrategyOptions
     /// <summary>
     /// Gets or sets the timeout generator that generates the timeout for a given execution.
     /// </summary>
+    /// <remarks>
+    /// By default, the generator provides <see cref="System.Threading.Timeout.InfiniteTimeSpan"/> which effective disables the timeout strategy.
```
martincostello, line 20:
```suggestion
    /// By default, the generator provides <see cref="System.Threading.Timeout.InfiniteTimeSpan"/> which effectively disables the timeout strategy.
```

src/Polly.Core/Timeout/TimeoutRejectedException.cs
```diff
@@ -0,0 +1,80 @@
...
+    protected TimeoutRejectedException(SerializationInfo info, StreamingContext context)
+        : base(info, context)
+    {
+    }
+#endif
+
+    /// <summary>
+    /// Gets the timeout value that caused this exception.
+    /// </summary>
+    public TimeSpan Timeout { get; private set; } = TimeoutConstants.NoTimeout;
```
martincostello, line 79: For `NETSTANDARD2_0` we need to read the value of `Timeout` out of the serialization info to restore it in the constructor and write it out in `GetObjectData()`.

src/Polly.Core/Timeout/TimeoutResilienceStrategy.cs
```diff
@@ -0,0 +1,113 @@
...
+            return await callback(context, state).ConfigureAwait(context.ContinueOnCapturedContext);
+        }
+        catch (OperationCanceledException e) when (cancellationSource.IsCancellationRequested && !previousToken.IsCancellationRequested)
+        {
+            _telemetry.Report(TimeoutConstants.OnTimeoutEvent, context);
+
+            await OnTimeout(new OnTimeoutArguments(context, e, timeout)).ConfigureAwait(context.ContinueOnCapturedContext);
+
+            throw new TimeoutRejectedException(
+                $"The timeout resilience strategy did not complete within the timeout of {(int)timeout.TotalSeconds} seconds.",
```
martincostello, line 63: I would omit the `int` cast so that if a timeout of, say, 1.5 seconds is used, then the user will see the fractions in the message.

Maybe just format it as a TimeSpan string, rather than convert it to seconds?

src/Polly.Core/Timeout/TimeoutValidator.cs
```diff
@@ -0,0 +1,15 @@
+namespace Polly.Timeout;
+
+internal static class TimeoutValidator
+{
+    internal static void ValidateTimeout(TimeSpan timeout)
+    {
+        if (timeout <= TimeSpan.Zero || timeout == System.Threading.Timeout.InfiniteTimeSpan)
```
martincostello, line 7: Should this be this?

```suggestion
        if (timeout <= TimeSpan.Zero && timeout != System.Threading.Timeout.InfiniteTimeSpan)
```

This seems to throw if you specify an infinite timeout but then the message says that that's a valid value.

src/Polly.Core/Timeout/TimeoutRejectedException.cs
martincostello: True. But this is only partial support.

Either we remove the support entirely, like in the other PR, and suppress the warnings, or we should implement it fully.

I would err on the side of removing it.

src/Polly.Core/Timeout/TimeoutValidator.cs
martintmk: Yes indeed, fixed.

src/Polly.Core/Timeout/TimeoutResilienceStrategy.cs
martintmk: Done.

src/Polly.Core/Timeout/TimeoutRejectedException.cs
martintmk: I have fixed this to serialize delay correctly. We can drop the serialization support in separate PR.

src/Polly.Core/Timeout/TimeoutStrategyOptions.cs
martintmk: I have dropped the support for pessimistic timeout in this PR to make it simpler. The property is no longer there.

src/Polly.Core/Timeout/OnTimeoutEvent.cs
```diff
@@ -0,0 +1,78 @@
...
+        });
+        return this;
+    }
+
+    /// <summary>
+    /// Adds a synchronous on-timeout callback.
+    /// </summary>
+    /// <param name="onTimeout">The specified callback.</param>
+    /// <returns>This instance.</returns>
+    public OnTimeoutEvent Add(Action onTimeout)
```
geeknoid, line 46: There's no need/value for a similar Func overload?

```
public OnTimeoutEvent Add(Func<ValueTask> onTimeout)
```

src/Polly.Core/Timeout/OnTimeoutEvent.cs
```diff
@@ -0,0 +1,78 @@
...
+            return default;
+        });
+        return this;
+    }
+
+    internal Func<OnTimeoutArguments, ValueTask> CreateHandler()
+    {
+        return _events.Count switch
+        {
+            0 => _ => default,
```
geeknoid, line 62: You might want to check whether this allocates a delegate every time through or not. If it does allocate every time through, then you can just use a static field to hold an instance.

Alternatively, if the 0 case is common, it might be more efficient to return null here, and to have the call site do null checks before invoking the delegate. Calling into a delegate will be more expensive than the null check. If 0 is rare, then this approach is better.

src/Polly.Core/Timeout/TimeoutGenerator.cs
```diff
@@ -0,0 +1,68 @@
+namespace Polly.Timeout;
+
+/// <summary>
+/// Generator for the timeout for a given execution of an user-provided callback.
```
geeknoid, line 4:
```suggestion
/// Generator for the timeout for a given execution of a user-provided callback.
```

geeknoid (approved):

src/Polly.Core/Timeout/OnTimeoutEvent.cs
martintmk: My thinking was that for async callbacks not to encourage too much. If someone want to use them, they should also consume the arguments.

src/Polly.Core/Timeout/OnTimeoutEvent.cs
```diff
@@ -0,0 +1,78 @@
...
+        };
+    }
+
+    private static Func<OnTimeoutArguments, ValueTask> CreateHandler(Func<OnTimeoutArguments, ValueTask>[] callbacks)
+    {
+        return async args =>
+        {
+            foreach (var callback in callbacks)
+            {
+                await callback(args).ConfigureAwait(args.Context.ContinueOnCapturedContext);
```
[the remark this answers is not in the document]
martintmk: I would say yes, concurrent execution brings problems regarding the `AggregateException` and reduces performance (since here, in 99% of cases the callbacks will be synchronous.

I'll add a note here in docs.

src/Polly.Core/Timeout/OnTimeoutEvent.cs
martintmk: I have added the `EmptyEvent` static field that's used when there are no callbacks. This also improves the readability. 

> Calling into a delegate will be more expensive than the null check. If 0 is rare, then this approach is better.

Only focusing on core functionality now, we can do full optimalization pass later once we have the features in. Wdyt?

src/Polly.Core/Timeout/TimeoutResilienceStrategy.cs
```diff
@@ -0,0 +1,85 @@
...
+        catch (OperationCanceledException e) when (cancellationSource.IsCancellationRequested && !previousToken.IsCancellationRequested)
+        {
+            _telemetry.Report(TimeoutConstants.OnTimeoutEvent, context);
+
+            await OnTimeout(new OnTimeoutArguments(context, e, timeout)).ConfigureAwait(context.ContinueOnCapturedContext);
+
+            await DisposeRegistration(registration).ConfigureAwait(context.ContinueOnCapturedContext);
+
+            throw new TimeoutRejectedException(
+                $"The timeout resilience strategy did not complete within the '{timeout}' timeout.",
```
martincostello, line 62:
```suggestion
                $"The timeout resilience strategy did not complete within the timeout of '{timeout}'.",
```

src/Polly.Core/Timeout/OnTimeoutEvent.cs
geeknoid: > Only focusing on core functionality now, we can do full optimalization pass later once we have the features in. Wdyt?

I personally find it hard to come back and find/fix this kind of small stuff. But if you prefer doing the micro-optimizations later on, that also definitely works.

src/Polly.Core/Timeout/TimeoutResilienceStrategy.cs
geeknoid: How about just

The operation didn't complete within the allowed timeout of {timeout}.

martincostello: Sure, updated the comment. Either way reads better with the formatted timespan value in it compared to the current message.

src/Polly.Core/Timeout/OnTimeoutEvent.cs
martintmk: You are right here, just making the result callback nullable and returning null if no events are registered is simple speeds-up the code with minimal effort.

martintmk:
> The shape of the API seems good to me.

@martincostello, I think I am done with the PR. Do you think there is anything left to address?

We can still do some cleanups in follow-up.

src/Polly.Core/Timeout/TimeoutRejectedException.cs
```diff
@@ -0,0 +1,90 @@
...
+    public TimeoutRejectedException(string message, Exception innerException)
+        : base(message, innerException)
+    {
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="TimeoutRejectedException" /> class.
+    /// </summary>
+    /// <param name="timeout">The timeout value that caused this exception.</param>
+    public TimeoutRejectedException(TimeSpan timeout) => Timeout = timeout;
```
martincostello, line 45: Same here.

src/Polly.Core/Timeout/TimeoutStrategyOptions.cs
```diff
@@ -0,0 +1,53 @@
+using System;
+using System.ComponentModel.DataAnnotations;
+using Polly.Builder;
+
+namespace Polly.Timeout;
+
+/// <summary>
+/// The options for the timeout strategy.
```
martincostello, line 8:
```suggestion
/// Represents options for the timeout strategy.
```

src/Polly.Core/Timeout/TimeoutStrategyOptions.cs
```diff
@@ -0,0 +1,53 @@
...
+    public TimeSpan Timeout { get; set; } = InfiniteTimeout;
+
+    /// <summary>
+    /// Gets or sets the timeout generator that generates the timeout for a given execution.
+    /// </summary>
+    /// <remarks>
+    /// By default, the generator is empty and the <see cref="Timeout"/> is used by default.
+    /// If generator returns a <see cref="TimeSpan"/> value that is less or equal to <see cref="TimeSpan.Zero"/>
+    /// its value is ignored and <see cref="TimeSpan"/> is used instead.
+    ///
```
martincostello, line 42: (I think this is the right syntax)

```suggestion
    /// <para/>
    ///
```

martincostello (approved):
Just a few minor comments otherwise LGTM.

src/Polly.Core/Timeout/TimeoutRejectedException.cs
```diff
@@ -0,0 +1,90 @@
...
+/// </summary>
+#if !NETCOREAPP
+[Serializable]
+#endif
+public class TimeoutRejectedException : ExecutionRejectedException
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="TimeoutRejectedException" /> class.
+    /// </summary>
+    public TimeoutRejectedException()
```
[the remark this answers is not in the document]
martintmk: Yup, good idea.

src/Polly.Core/Timeout/TimeoutStrategyOptions.cs
martintmk: Done.

martincostello (approved):
:shipit:

PR #1109, opened 2023-04-11
Fix sample code for `Timeout` overload

prodehghan:
The overload takes a `Func<TimeSpan>`, and it is assumed that we already have a lambda variable called `myTimeoutProvider, but the variable is incorrectly passed to the overload.

### Confirm the following

### DIFF
--- a/README.md
+++ b/README.md
@@ -619,7 +619,7 @@ Policy
 
 // Configure variable timeout via a func provider.
 Policy
-  .Timeout(() => myTimeoutProvider)) // Func<TimeSpan> myTimeoutProvider
+  .Timeout(myTimeoutProvider)) // Func<TimeSpan> myTimeoutProvider
 
 // Timeout, calling an action if the action times out
 Policy


README.md
```diff
@@ -619,7 +619,7 @@ Policy
 
 // Configure variable timeout via a func provider.
 Policy
-  .Timeout(() => myTimeoutProvider)) // Func<TimeSpan> myTimeoutProvider
+  .Timeout(myTimeoutProvider)) // Func<TimeSpan> myTimeoutProvider
```
martincostello, line 622: Alternatively, it could also be this:

```suggestion
  .Timeout(() => myTimeoutProvider())) // Func<TimeSpan> myTimeoutProvider
```

martincostello (approved):

README.md
prodehghan: Yes, but why someone would need to do this? 😊

martincostello: IMHO it makes the intent clearer to the reader _outside of an IDE_ (i.e. in our documentation) that it's a delegate and not simple variable.

prodehghan: I also thought of that, but that's why there is a comment in front of the code. It is not always possible (or good) to make everything *super* easy. The reader already took the time to read the README file to that point and already looked at samples. At this point, there are four different hints for the reader: 1- This sample should have something new compared to the two samples above it, 2- The comment above the code says "via a func provider", 3- the variable name ends with *Provider*, 4- The comment in front states the variable type.

No one can help someone that just wants to copy a code snippet that he/she found, blindly, without paying attention.

Sorry for the long comment 😋. I hope I was able to delivery my point.

prodehghan: And the library is already complicated enough, that *not* paying attention will not help the reader.

PR #1111, opened 2023-04-12
Introduce ObjectPool and use it for ResilienceContext pooling

### LINKED ISSUE #955: Pooling of Context instances
**Is your feature request related to a specific problem? Or an existing feature? Please describe.**

We have noticed increased allocations related to creating the `Context` object. This can be easily avoided if Pollu provides built-in support for pooling or at least slightly change the internal implementation of `Context` so we can easily reset it to default state.

Currently, this is not achievable because of these read-only properties:

* `OperationId`
* `CorrelationId`

**Describe your proposed or preferred solution**:

We want to extend the `Context` API with the following static methods:

``` csharp
public static Context Get();
public static Context Get(string operationName);
public static void Return(Context context);
```
Alternatively, we can introduce a new `ContextPool` static class.

The pool can be also used internally to improve Polly performance.

**Describe any alternative options you've considered**:

Simpler solution is to make the following changes to `Context` class:

``` csharp
protected Guid? CorrelationIdField { get; protected set; }
public string OperationKey { get; protected set; }
```

We would just derive from `Context` and handle the pooling ourselves.

**Any additional info?**

For pooling support, we need implementation of the pool, and we can either:

* Include reference to [Microsoft.Extensions.ObjectPool](https://www.nuget.org/packages/Microsoft.Extensions.ObjectPool)
* Implement our own pool (copy + paste) if additional dependencies are a no-go

martintmk:

Closes #955 

The `ObjectPool` is just trimmed down version of [`DefaultObjectPool`](https://github.com/dotnet/aspnetcore/blob/main/src/ObjectPool/src/DefaultObjectPool.cs). 

I had to add more tests so the code passes our gates. I will re-run the benchmarks later once we add polling for `CancellationTokenSource` instances.

### Confirm the following

### DIFF
--- a/src/Polly.Core/Utils/ObjectPool.cs
+++ b/src/Polly.Core/Utils/ObjectPool.cs
@@ -1,0 +1,68 @@
+// Licensed to the .NET Foundation under one or more agreements.
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+
+namespace Polly.Utils;
+
+// copied from https://raw.githubusercontent.com/dotnet/aspnetcore/main/src/ObjectPool/src/DefaultObjectPool.cs
+internal sealed class ObjectPool<T>
+    where T : class
+{
+    internal static readonly int MaxCapacity = (Environment.ProcessorCount * 2) - 1; // the - 1 is to account for _fastItem
+
+    private readonly Func<T> _createFunc;
+    private readonly Func<T, bool> _returnFunc;
+
+    private readonly ConcurrentQueue<T> _items = new();
+
+    private T? _fastItem;
+    private int _numItems;
+
+    public ObjectPool(Func<T> createFunc, Func<T, bool> returnFunc)
+    {
+        _createFunc = createFunc;
+        _returnFunc = returnFunc;
+    }
+
+    public T Get()
+    {
+        var item = _fastItem;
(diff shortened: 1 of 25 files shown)


src/Polly.Core/Utils/ObjectPool.cs
```diff
@@ -0,0 +1,67 @@
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+
+namespace Polly.Utils;
+
+// copied from https://raw.githubusercontent.com/dotnet/aspnetcore/main/src/ObjectPool/src/DefaultObjectPool.cs
```
martincostello, line 7: Let's update this to point at a specific Git SHA so we can tell _when_ the code was taken too.

martincostello (commented):
Could you re-run the benchmarks with this change so we can see the difference it makes?

src/Polly.Core/Utils/ObjectPool.cs
martintmk: Done.

src/Polly.Core.Benchmarks/README.md
martincostello, line 1: Nice 😎

martincostello (approved):

PR #1123, opened 2023-04-14
Drop ResilienceTelemetryFactory

martintmk:

This PR drops the following public APIs:

- `ResilienceTelemetryFactory`
- `NullResilienceTelemetryFactory`
- `NullResilienceTelemetry`

The `ResilienceTelemetry` stays public and strategies can use it to report important resilience events. Its signature is expanded to allow passing additional arguments. Internally, `ResilienceTelemetry` uses `DiagnosticSource` to report events. In `Polly.Extensions` we will subscribe to this `DiagnosticSource` and report the events as logs and metrics.

### Confirm the following


src/Polly.Core/Polly.Core.csproj
```diff
@@ -24,6 +24,7 @@
     <PackageReference Include="System.Threading.Tasks.Extensions" Condition="!$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'netcoreapp3.1'))" />
     <PackageReference Include="System.ValueTuple" Condition="$([MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)')) == '.NETFramework'" />
     <PackageReference Include="System.ComponentModel.Annotations" Condition="!$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'netcoreapp3.1'))" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Condition="!$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'netcoreapp3.1'))" />
```
martincostello, line 27: Add to the dependabot ignores.

src/Polly.Core/Strategy/Outcome.TResult.cs
```diff
@@ -0,0 +1,88 @@
...
+
+    /// <summary>
+    /// Gets the exception that occurred during the operation, if any.
+    /// </summary>
+    public Exception? Exception { get; }
+
+    /// <summary>
+    /// Gets the result produced by the operation, if any.
+    /// </summary>
+    public TResult? Result { get; }
```
martincostello, line 35: Can we add "not null when" attributes to these that look to `HasResult`?

src/Polly.Core/Strategy/Outcome.TResult.cs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    internal Outcome AsOutcome()
+    {
+        if (Exception != null)
+        {
+            return new Outcome(typeof(TResult), Exception);
+        }
+
+        return new Outcome(typeof(TResult), Result);
```
martincostello, line 86: If the attributes work, then these could be changed to `if (HasResult)` and then the `?`s could be removed?

src/Polly.Core/Strategy/Outcome.cs
```diff
@@ -32,7 +41,12 @@ public Outcome(TResult result)
     /// <summary>
     /// Gets the result produced by the operation, if any.
     /// </summary>
-    public TResult? Result { get; }
+    public object? Result { get; }
```
martincostello, line 44: Same here regarding the nullable attributes.

src/Polly.Core/Telemetry/ResilienceTelemetrySource.cs
```diff
@@ -0,0 +1,4 @@
+namespace Polly.Telemetry;
+
+internal record class ResilienceTelemetrySource(string BuilderName, ResilienceProperties BuilderProperties, string StrategyName, string StrategyType);
```
martincostello, line 3: Sealed?

src/Polly.Core/Telemetry/TelemetryEventArguments.cs
```diff
@@ -0,0 +1,8 @@
+using Polly.Strategy;
+
+namespace Polly.Telemetry;
+
+internal record class TelemetryEventArguments(ResilienceTelemetrySource Source, string EventName, IResilienceArguments Arguments, Outcome? Outcome) : IResilienceArguments
```
martincostello, line 5: Sealed?

src/Polly.Core/Strategy/Outcome.TResult.cs
martintmk: I think that still would be misleading. We can still do:

``` csharp
new Outcome<MyClass>(null!)
```

`null` for some scenarios can be considered a valid result. So `HasResult` is true but the result is null.

Which actually means this signature is invalid:

``` csharp
    public bool TryGetResult([NotNullWhen(true)] out TResult? result)
```
I'll drop `NotNullWhen`.

src/Polly.Core/Polly.Core.csproj
martintmk: I checked, it's already there.

src/Polly.Core/Strategy/Outcome.TResult.cs
martintmk: Same problem as above, it's perfectly fine to return a null result and execute the strategy as:

``` csharp
strategy.Execute<object>(() => null!);
```

src/Polly.Core/Telemetry/TelemetryEventArguments.cs
martintmk: Done.

martincostello (approved):

PR #1125, opened 2023-04-14
Cleanup of callback API

martintmk:

- Turning events, predicates and generators from abstract to sealed classes. The extra complexity didn't warrant the added value.
- Rename `SimpleEvent` to `NoOutcomeEvent`
- Added non-generic methods to `OutcomeGenerator`. This allows adding new extensions for retry.
- Adding a new generic `OutcomePredicate<TArgs, TResult>` that only configures the predicates for a specific result-type.
- Deleting unnecessary classes.
- Prefixing methods in OutcomePredicate with `Handler`
- Renaming methods in OutcomeEvent from `Add` to `Register`.

In the follow-up PR I will also add non-generic overloads to `OutcomeEvent`

### Confirm the following


src/Polly.Core.Tests/Retry/RetryResilienceStrategyBuilderExtensionsTests.cs
```diff
@@ -4,25 +4,41 @@
...
-            builder.AddRetry(retry=>retry.Result(10), RetryBackoffType.Linear, 2, TimeSpan.FromSeconds(1));
+            builder.AddRetry(retry=>retry.HandleResult(10), RetryBackoffType.Linear, 2, TimeSpan.FromSeconds(1));
             AssertStrategy(builder, RetryBackoffType.Linear, 2, TimeSpan.FromSeconds(1));
         },
+        builder =>
+        {
+            builder.AddRetry(retry => retry.HandleResult(10), attempt =>
+            {
+                return TimeSpan.FromMilliseconds(attempt);
+            });
```
martincostello, line 33: Maybe use the terser syntax?
```suggestion
            builder.AddRetry(retry => retry.HandleResult(10), attempt => TimeSpan.FromMilliseconds(attempt));
```

src/Polly.Core/Retry/RetryResilienceStrategyBuilderExtensions.cs
```diff
@@ -28,6 +29,28 @@ public static class RetryResilienceStrategyBuilderExtensions
...
+    /// <param name="shouldRetry">A predicate that defines the retry conditions.</param>
+    /// <param name="retryDelayGenerator">The generator for retry delays. The argument is zero-based attempt number.</param>
+    /// <returns>The builder instance with the retry strategy added.</returns>
+    public static ResilienceStrategyBuilder AddRetry(
+        this ResilienceStrategyBuilder builder,
+        Action<OutcomePredicate<ShouldRetryArguments>> shouldRetry,
+        Func<int, TimeSpan> retryDelayGenerator)
+    {
+        Guard.NotNull(builder);
+        Guard.NotNull(shouldRetry);
```
martincostello, line 45: Guard for `retryDelayGenerator` as it's not nullable?

src/Polly.Core/Strategy/OutcomeGenerator.cs
```diff
@@ -3,26 +3,16 @@
...
-
 /// <summary>
 /// The base class for all generators that generate a value based on the <see cref="Outcome{TResult}"/>.
 /// </summary>
-/// <typeparam name="TGeneratedValue">The type of generated value.</typeparam>
 /// <typeparam name="TArgs">The arguments the generator uses.</typeparam>
-/// <typeparam name="TSelf">The class that implements <see cref="OutcomeGenerator{TGeneratedValue, TArgs, TSelf}"/>.</typeparam>
-public abstract partial class OutcomeGenerator<TGeneratedValue, TArgs, TSelf>
+/// <typeparam name="TGeneratedValue">The type of generated value.</typeparam>
+public sealed partial class OutcomeGenerator<TArgs, TGeneratedValue>
```
martincostello, line 11: Maybe `TResult` or `TOutcome` instead of `TGeneratedValue`?

src/Polly.Core/Strategy/OutcomePredicate.TResult.cs
```diff
@@ -0,0 +1,222 @@
...
+using Polly.Strategy;
+
+namespace Polly.Strategy;
+
+/// <summary>
+/// Predicate that uses <see cref="Outcome{TResult}"/> and <typeparamref name="TResult"/> as an input.
+/// </summary>
+/// <typeparam name="TArgs">The type of arguments the predicate uses.</typeparam>
+/// <typeparam name="TResult">The result type that this predicate handles.</typeparam>
+public sealed class OutcomePredicate<TArgs, TResult>
```
martincostello, line 11: Maybe `TOutcome`? Whatever the name I think the name should match the type parameters on things like `OutcomeGenerator` like with the previous comment.

src/Polly.Core/Strategy/OutcomePredicate.TResult.cs
```diff
@@ -0,0 +1,222 @@
...
+    /// Adds an exception predicate for the specified exception type.
+    /// </summary>
+    /// <typeparam name="TException">The exception type to add a predicate for.</typeparam>
+    /// <returns>The current updated instance.</returns>
+    public OutcomePredicate<TArgs, TResult> HandleException<TException>()
+        where TException : Exception
+    {
+        return HandleOutcome((outcome, _) =>
+        {
+            if (outcome.Exception is TException typedException)
```
martincostello, line 31: We don't seem to use/need the actual value?

```suggestion
            if (outcome.Exception is TException)
```

src/Polly.Core/Strategy/OutcomePredicate.TResult.cs
```diff
@@ -0,0 +1,222 @@
...
+    /// Adds a result predicate for the specified result value.
+    /// </summary>
+    /// <param name="value">The result value to be retried.</param>
+    /// <returns>The current updated instance.</returns>
+    /// <remarks>
+    /// By default, the default equality comparer is used to compare the result value with the value specified in this method.
+    /// </remarks>
+    public OutcomePredicate<TArgs, TResult> HandleResult(TResult value)
+    {
+        return HandleResult((result, _) => new ValueTask<bool>(EqualityComparer<TResult>.Default.Equals(result!, value)));
```
martincostello, line 116: Should we provide an overload that allows for a user-specified equality comparer, or are the `Func<TResult, bool>` overloads enough to allow a user to do that themselves with: this?

```csharp
var myComparer = ...;
HandleResult((result, _) => new ValueTask<bool>(myComparer.Equals(result!, value)
```

Adding the overload makes it simpler and avoids the (extra?) closure, but having the overload(s) increases the API surface area/complexity.

src/Polly.Core/Strategy/OutcomeGenerator.cs
martintmk: `TResult` collides with generic methods that have `TResult` and I feel `TOutcome` collides with the `Outcome` class we already have. The generated value is not really an outcome.

martincostello: The naming just feels a bit weird to me.

The type is described as a _"generator of an outcome"_, so it feels to me that its output would be an _"outcome"_, not a _"generated"_. A bit like a _"foo factory"_ produces _"foo-s"_ rather than _"fabricated-s"_.

What if we drop the `Geneated` prefix, and just have `TValue`? Or will that clash elsewhere too?

src/Polly.Core/Strategy/OutcomePredicate.TResult.cs
martintmk: I have added optional comparer parameter. I think it's fine and provides greater flexibility.

src/Polly.Core.Tests/Retry/RetryResilienceStrategyBuilderExtensionsTests.cs
martintmk: Done.

src/Polly.Core/Strategy/OutcomeGenerator.cs
martintmk: Nope, `TValue` is fine. We can use it.

src/Polly.Core/Strategy/OutcomePredicate.TResult.cs
martintmk: Done.

src/Polly.Core/Strategy/OutcomePredicate.TResult.cs
martintmk: TResult is consistent with naming in `OutcomeGenerator`:

``` csharp
OutcomeGenerator<TArgs, TValue> SetGenerator<TResult>(Func<Outcome<TResult>, TArgs, TValue> generator)
```

We also use the same name at various places:

``` csharp
public readonly struct Outcome<TResult>

public class ResilienceStrategy
{
    public TResult Execute<TResult>(
        Func<CancellationToken, TResult> callback,
        CancellationToken cancellationToken = default)
}
```

martincostello (approved):

PR #1134, opened 2023-04-15
Add publish workflow

martincostello:
Add a workflow to publish NuGet packages to NuGet.org - the packages are pushed only when a tag is created in this repository.

As currently configured, anyone with write access to the repository would be able to publish a package as they would be able to use the secret.

If this isn't desirable for some reason, we could have the NuGet publish token as an _environment secret_, and have that require approval by specific GitHub accounts by tying the NuGet publish step to that environment.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,7 @@ name: build
 on:
   push:
     branches: [ main ]
+    tags: [ '*' ]
   pull_request:
     branches: [ main ]
   workflow_dispatch:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -95,4 +95,5 @@ jobs:
       shell: pwsh
       run: |
         Write-Host "This is where the packages would be published to NuGet.org"
-        <# dotnet nuget push "artifacts\nuget-packages\*.nupkg" --api-key ${{ secrets.NUGET_TOKEN }} --skip-duplicate --source https://api.nuget.org/v3/index.json #>
+        ls
+        <# dotnet nuget push "*.nupkg" --api-key ${{ secrets.NUGET_TOKEN }} --skip-duplicate --source https://api.nuget.org/v3/index.json #>
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,15 @@ on:
     branches: [ main ]
   workflow_dispatch:
 
+env:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -73,4 +74,52 @@ jobs:
...
+    - name: Download packages
+      uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
+      with:
+        name: packages-windows
+
+    - name: Setup .NET SDK
+      uses: actions/setup-dotnet@607fce577a46308457984d59e4954e075820f10a # v3.0.3
+
+    - name: Push NuGet packages to NuGet.org
+      run: dotnet nuget push "*.nupkg" --api-key ${{ secrets.NUGET_TOKEN }} --skip-duplicate --source https://api.nuget.org/v3/index.json
```
martincostello, line 125: Assuming we're happy with this workflow, we just need a token adding to the repository secrets so this will actually work.

.github/workflows/build.yml
```diff
@@ -74,4 +77,52 @@ jobs:
...
```
martintmk, line 128: Maybe push to some temporary source while we are validating the workflow? Jus tin case we produce some invalid package :)

martintmk (approved):
LGTM,

.github/workflows/build.yml
martincostello: I was thinking about adding GitHub Packages [like this](https://github.com/justeattakeaway/JustSaying/blob/c5e1b001a1e272326bc51a7dfb7b0babcde32d55/.github/workflows/build.yml#L108-L128), unless we wanted to use something like MyGet instead.

martintmk: Maybe just disable this step until we are sure (or put the nuget publishing step behind variable)?

martincostello: It's effectively behind a variable already because the job is conditional on a tag:

https://github.com/App-vNext/Polly/pull/1134/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R116

joelhulen (approved):
Good to go. I'm adding the environment stuff now.

.github/workflows/build.yml
joelhulen: I just added the NUGET_TOKEN repository secret.

PR #1140, opened 2023-04-18
Introduce TelemetryResilienceStrategy

### LINKED ISSUE #1113: Introduce TelemetryResilienceStrategy
The `TelemetryResilienceStrategy` logs and meters the execution times of resilience strategies. It needs to be explicitly added to `ResilienceStrategyBuilder`. 

It uses enrichers introduced by #1112 to enrich the metered dimensions.

Metering Properties:

Meter Name: `Polly`
Instrument Type: Histogram (tracks execution time)
Instrument Name: `strategy-execution-duration`

Dimensions: 

- `builder-name`: The builder name or null
- `strategy-key`: The strategy key or null
- `result-type`: The type of result this strategy is being executed for Int, String, VoidResult
- `exception-name`: System.InvalidOperationException, System.Net.HttpRequestException or `N/A`
- `execution-health`: Healthy/Unhealthy. The result is reported as healthy when the number of resilience events that occurred during the execution is zero. (no retries, timeouts, rate-limits etc.) 

For instrument and dimensions names we are following the [conventions](https://github.com/dotnet/aspnetcore/blob/main/src/Servers/Kestrel/Core/src/Internal/Infrastructure/KestrelMetrics.cs#L272) from ASP.NET Core.

martintmk:

Closes #1113 

This PR enables telemetry support when executing individual strategies. The `TelemetryResilienceStrategy` is automatically added when `EnableTelemetry` is called without exposing any new API.

Additional changes:

- Introducing `ResilienceContext.ResilienceEvents` to track all important events that happened during execution.
- General cleanup around tests by introducing `Polly.TestUtils` project.

This PR should finalize the telemetry story.

### Confirm the following


src/Polly.Core/ResilienceContext.cs
```diff
@@ -55,6 +58,14 @@ private ResilienceContext()
     /// </summary>
     public ResilienceProperties Properties { get; } = new();
 
+    /// <summary>
+    /// Gets the collection of resilience events that occurred while executing the resilience strategy.
+    /// </summary>
+    /// <remarks>
+    /// If the number of resilience events is greater than zero it's an indication that the execution was unhealthy.
+    /// </remarks>
```
martincostello, line 66: Might people write their own extensions that report successful events? For example I was thinking earlier that maybe for when I look at caching I'd see at adding cache hit/miss telemetry.

src/Polly.Extensions/Telemetry/EnrichmentUtil.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        foreach (var enricher in enrichers)
+        {
+            enricher(context);
+        }
+
+        foreach (var pair in context.Tags)
+        {
+            tags.Add(pair.Key, pair.Value);
+        }
```
martincostello, line 31:
```suggestion
        if (context.Tags > 0)
        {
            foreach (var pair in context.Tags)
            {
                tags.Add(pair.Key, pair.Value);
            }
        }
```

src/Polly.Extensions/Telemetry/Log.cs
```diff
@@ -0,0 +1,77 @@
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Polly.Extensions.Telemetry;
+
+#pragma warning disable S107 // Methods should not have too many parameters
+#pragma warning disable SA1118 // Parameter should not span multiple lines
+
+internal static class Log
```
martincostello, line 9: Can we conditionally compile this class so we can use the Logging Source Generator for .NET 6+ for better performance?

src/Polly.Core/ResilienceContext.cs
martintmk: Inside your strategy you can use `ResilienceStrategyTelemetry` to report individual events and these will be reported here. However, these should really be treated as "warnings". 

In custom strategies you can add your own collection into `ResilienceProperties` and use that to report cache events.

For example:
``` csharp
var events = resilienceContext.Properties.GetValue(new ResiliencePropertyKey<List<string>>("cache-events"), new List<string>());
events.Add("hit");
```

src/Polly.Extensions/Telemetry/Log.cs
martintmk: Done.

src/Polly.Extensions/Telemetry/EnrichmentUtil.cs
martintmk: This seems artificial, once enricher is registered it will add a tag to the collection. Otherwise, there is no point registering it. 

So the expectation is that `context.Tags` is never empty when we have enrichers.

src/Polly.Core/ResilienceContext.cs
martintmk: Also, maybe we can expand the `ReportedResilienceEvent` with `Severity` later?

That way you can also report `informational` events and use the common-infra.

martincostello: That sounds like a good idea.

martincostello (approved):

src/Polly.Core/ResilienceContext.cs
martintmk: Created a new issue: #1143

Issue #1141, opened 2023-04-18
Validate v8 build with .NET 8


martincostello:
Add a build configuration to validate in CI that Polly v8 works as expected with .NET 8 with a `net8.0` TFM using the prerelease packages.

Once preview 4 is available, remove the internal `TimeProvider` copy and use the version shipping in a NuGet package instead.

Once .NET 8 ships as stable in November, we can then roll `net8.0` into the main build setup and ship a package that explicitly supports it.

martintmk:
As a part of this task we should also unify all `Microsoft.Extensions.*` and `System.*` packages to version 8.0.0. These versions should fix known compatibility problems seen in versions. Would simplify our setup too as we don't have to deal with different versions for each framework anymore.

martincostello:
Yep I'll sort that out once it's available (that's p4 too right?). For now I've just extended the multi-targeting while I'm playing with it locally.

martintmk:
> that's p4 too right?

Hopefully yes, we just need to wait until the `System.Bcl.TimeProvider` package appears in nuget :)

martintmk:
@martincostello I came across this post:

https://devblogs.microsoft.com/dotnet/package-validation/

I think it's something we can do for V8 too! ;)

martincostello:
Yep it would be good to add that too - I already use it in a number of my OSS projects. We should be able to enable it separate from the .NET 8 upgrade.

martincostello:
Added package validation in #1172. There's also some additional checking already as part of #1134.

martincostello:
Completed by #1144.

PR #1144, opened 2023-04-18
Add support for .NET 8

### LINKED ISSUE #1070: Use the new TimeProvider API from .NET 8
Instead of writing our own variant of `IClock` that allows mocking time in tests we can now leverage the [TimeProvider](https://github.com/dotnet/runtime/issues/36617#issue-619777517) API that will be introduced in .NET 8. 

The `TimeProvider` will be also available in .NET Framework by referencing the new nuget package. 

Until the package is ready we can just copy the implementaiton into Polly, mark all classes as internal and once the official package is ready expose `ResilienceStrategyBuilderOptions.TimeProvider` property. 

Wit this approach we will be using unified time abstraction and also allow easy unit-testing of strategies working with time.

### LINKED ISSUE #1786: [Feature request]: Consider dowgrading Microsoft.Extensions.Options dependency for older targets

I need to support older code that targets .NET Framework. I am using Polly for resiliency and would like to bump it to v8. But `Polly.Extensions` depends on `Microsoft.Extensions.Options` v2.2.0 which is out of support. Please consider downgrading to v2.1.x, which is still [supported on .NET Framework](https://dotnet.microsoft.com/en-us/platform/support/policy/aspnet/2.1-packages).

`Polly.Extensions` should depend on `Microsoft.Extensions.Options` >= 2.1.0

### Additional context

_No response_

martincostello:
- Add `net8.0` TFM.
- Bump packages to .NET 8 versions where relevant.
- Use the [new `System.TimeProvider` abstraction](https://devblogs.microsoft.com/dotnet/announcing-dotnet-8-preview-4/#introducing-time-abstraction) to resolve #1070.

Relates to #1141, depends on #1738, resolves #1786.


martintmk:
> @martintmk Do you want to push your changes to use `System.TimeProvider` into this branch?

Yup, but first let me create another PR that's prerequisite for this. Then I'll push changes directly to yours.

eng/Analyzers.targets
```diff
@@ -13,5 +12,9 @@
     <RunAnalyzersDuringBuild>true</RunAnalyzersDuringBuild>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <AnalysisLevel>latest</AnalysisLevel>
+    <EnableNETAnalyzers>true</EnableNETAnalyzers>
+    <!--
+    <AnalysisMode>AllEnabledByDefault</AnalysisMode>
```
martintmk, line 17: What does this thing does? Do we still get all analyzers with this change?

martincostello: It enables all of the in-box analysers - I've commented it out for now as it adds _a lot_ of new errors (whether they're valid or not).

martintmk: OK, it's something we can address in follow-up. 

Btw, have you thought about enabling the new "artifacts" folder? Looks neat.

martincostello: Yep, I was just going to wait for your changes (and #1134) first.

I've been turning it on in various other repos of mine this morning after I had issues with it testing with preview 3.

src/Polly.Core/Polly.Core.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net7.0;net6.0;netstandard2.0;net472;net462</TargetFrameworks>
+    <TargetFrameworks>net8.0;net7.0;net6.0;netstandard2.0;net472;net462</TargetFrameworks>
```
martintmk, line 4: We should also add conditional reference to `System.Bcl.TimeProvider` on .NET < 8.0

src/Directory.Packages.props
```diff
@@ -1,14 +1,20 @@
 <Project>
+  <PropertyGroup>
+    <MicrosoftExtensionsVersion>8.0.0-preview.4.23259.5</MicrosoftExtensionsVersion>
```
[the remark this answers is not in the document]
martintmk: Yup, the sooner we switch to these the sooner we can report any compatibility issues. These will be then backported to 8.0.0 version.

martincostello:
@martintmk Could you rebase this onto `main` so we can update to preview 5?

martintmk:
> @martintmk Could you rebase this onto `main` so we can update to preview 5?

Rebased. Few tests are failing, we can address that later.

martincostello:
I'll look at bumping this to preview 5 and fixing the tests today.

martincostello:
I tweaked a few assertions with what I thought was the correct change, but I wasn't 100% what the intended behaviour change was for the other failures.

martincostello:
I'll try and update this again once #1355 is merged.

martincostello:
Blocked on adopting `Microsoft.Extensions.TimeProvider.Testing` until the fix for https://github.com/dotnet/runtime/issues/88000 is in a publicly available preview of .NET 8, which is probably not going to be until preview 7.

martincostello:
@martintmk Would be good to get some feedback on the changes in 7be0f937710ffcc055d372fc9f08f2057b711aba. I've removed the Moq-based implementation so we're not introspecting on the implementation details of the clock too much, but I'm not 100% sure the test refactoring is correct, even though they pass locally.

I also simplified the `TimeProviderExtensions` code as comparing it to the code in the NuGet version, we seem to duplicate a fair number of the conditions already in that code. If we just call it then we can remove the _"is it a mock or not?"_ branching - assuming I understood it correctly.

martintmk:
> @martintmk Would be good to get some feedback on the changes in [7be0f93](https://github.com/App-vNext/Polly/commit/7be0f937710ffcc055d372fc9f08f2057b711aba). I've removed the Moq-based implementation so we're not introspecting on the implementation details of the clock too much, but I'm not 100% sure the test refactoring is correct, even though they pass locally.
> 
> I also simplified the `TimeProviderExtensions` code as comparing it to the code in the NuGet version, we seem to duplicate a fair number of the conditions already in that code. If we just call it then we can remove the _"is it a mock or not?"_ branching - assuming I understood it correctly.

@martincostello, the extra code in `TimeProviderExtensions` for the synchronous path was done to avoid the sync-over-async pattern. But honestly, I am pretty fine with this change.  It makes out code simpler and it only happens when resilience event occurs, so not on the hot path. 

Your test refactorings are fine, otherwise some mutants would have survived. But I suggest to make those changes in the main branch too.

martincostello:
Cool, thanks. Yep, today I'll tidy up this branch a bit and then cherry-pick some of the changes to go into main now rather than wait on 8.0 RTM.

martincostello:
Will rebase once #1360 is merged.

martincostello:
Just tried the testing time provider again locally, but it's still broken in preview 6. I guess the fix for the issue I found didn't make it in in time and we'll have to wait for preview 7.

martintmk:
> Just tried the testing time provider again locally, but it's still broken in preview 6. I guess the fix for the issue I found didn't make it in in time and we'll have to wait for preview 7.

@martincostello What issue do you mean? Our tests are passing so it shouldn't affecting us.

martincostello:
This one: https://github.com/dotnet/runtime/issues/88000

martincostello:
@martintmk I've updated to the new time provider testing package, but there's a few hedging tests that use the custom hedging time provider that are now failing. Could you take a quick look when you get a moment?

martintmk:
> @martintmk I've updated to the new time provider testing package, but there's a few hedging tests that use the custom hedging time provider that are now failing. Could you take a quick look when you get a moment?

Looking at this.

martincostello:
Looks like something's changed for the metering tests.

src/Polly.Extensions/Polly.Extensions.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net7.0;net6.0;netstandard2.0;net472;net462</TargetFrameworks>
+    <TargetFrameworks>net8.0;net7.0;net6.0;netstandard2.0;net472;net462</TargetFrameworks>
```
martintmk, line 3: Based on #1534 

Do we really need to add net8.0 target if we are not using anything in it?

martincostello: I thought it made the TimeProvider usage better/best? Without looking at the diff (on phone atm), I think we're using it to pass it through to some new APIs in System for Task.Delay?

martincostello:
Not sure why the tests are failing now - they didn't locally in Visual Studio.

martincostello:
> Not sure why the tests are failing now - they didn't locally in Visual Studio.

Actually they are failing locally, but only for net481, net6.0 and net7.0.

src/Polly.Core/Polly.Core.csproj
```diff
@@ -25,6 +27,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Condition="!$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'netcoreapp3.1'))" />
+    <PackageReference Include="Microsoft.Bcl.TimeProvider" />
```
martincostello, line 30: Prior to RC.1 this was conditional, but it seemed to break locally after updating unless this was present for all TFMs.

test/Polly.Core.Tests/Timeout/TimeoutResilienceStrategyTests.cs
```diff
@@ -131,7 +132,12 @@ public async Task Execute_Timeout_EnsureStackTrace()
         ResilienceContextPool.Shared.Get(),
         "state");
         outcome.Exception.Should().BeOfType<TimeoutRejectedException>();
-        outcome.Exception!.StackTrace.Should().Contain("Execute_Timeout_EnsureStackTrace");
+
+#if NET8_0_OR_GREATER
+        outcome.Exception!.StackTrace.Should().Contain(nameof(Execute_Timeout_EnsureStackTrace));
+#else
+        outcome.Exception!.StackTrace.Should().Contain(nameof(StrategyHelper));
+#endif
```
martincostello, line 140: This fixes the tests, but I don't fully understand what's going on some I'm not sure this is the right thing to do.

src/Polly.Core/Polly.Core.csproj
martintmk: This is odd, we shouldn't be forced to add this for .NET 8

test/Polly.Core.Tests/Timeout/TimeoutResilienceStrategyTests.cs
martintmk: For some reason when .NET 8 is not involved the calling function `Execute_Timeout_EnsureStackTrace` is not included in the stack trace string.

src/Polly.Core/Polly.Core.csproj
martintmk: This was fixed by adding .NET 8 target to `Polly.Testing`. 

`Microsoft.Bcl.TimeProvider` is still not included on .NET 8

test/Polly.Core.Tests/Timeout/TimeoutResilienceStrategyTests.cs
martintmk: Happens only on ubuntu

martincostello: Is that a bug we should report, or is it an OS-specific implementation detail we just need to accept?

martintmk: Still investigating, however, to me it seems like a bug.

martintmk:
For some reason the mutation testing is broken in this PR :/

martincostello:
> [09:04:18 WRN] Could not find a project analysis for the chosen target framework net7.0. 
The available target frameworks are: net6.0,net8.0,net462,net472,netstandard2.0. 
The first available framework will be selected, which is net6.0.

Maybe explicitly change it to `net6.0`?

martincostello:
> Maybe explicitly change it to `net6.0`?

Nope, that didn't work either 😞

martincostello:
Rebased after #1734 but it's a bit broken. Will re-visit this to tidy it up and fix it properly over the next few days.

src/Polly.Core/ResiliencePipelineBuilderBase.cs
```diff
@@ -71,10 +71,9 @@ private protected ResiliencePipelineBuilderBase(ResiliencePipelineBuilderBase ot
     /// Gets or sets a <see cref="TimeProvider"/> that is used by strategies that work with time.
     /// </summary>
     /// <value>
-    /// The default value is <see cref="TimeProvider.System"/>.
+    /// The default value is <see langword="null"/> and unless set, the <see cref="TimeProvider.System"/> will be used.
```
martincostello, line 74:
```suggestion
    /// The default value is <see langword="null"/> and unless set, <see cref="TimeProvider.System"/> will be used.
```

src/Polly.Core/ResiliencePipelineBuilderBase.cs
```diff
@@ -129,13 +128,15 @@ internal PipelineComponent BuildPipelineComponent()
 
         var source = new ResilienceTelemetrySource(Name, InstanceName, null);
 
-        return PipelineComponentFactory.CreateComposite(components, new ResilienceStrategyTelemetry(source, TelemetryListener), TimeProvider);
+        return PipelineComponentFactory.CreateComposite(components, new ResilienceStrategyTelemetry(source, TelemetryListener), GetTimeProvider());
     }
 
+    internal TimeProvider GetTimeProvider() => TimeProvider ?? TimeProvider.System;
```
martincostello, line 134: What about `Resolve` instead of `Get`? Or just having a property called `TimeProviderInternal` or something?

martincostello:
Build failures are due to a known issue in the RC2 SDK when a newer version of a global tool is available than in the config file.

martincostello:
Looks like there's a flaky test when testing for .NET Framework related to the metrics stuff.

burger100:
Does it means Polly doesn't support .net version prior to .net 8.0 anymore ?
After an update to version 8.2.0 I got build error saying that TimeProvider is ambigous between our internal TimeProvider and the new System.TimeProvider added in .net8.0.  Our project is still targetting .net7.0.

PR #1156, opened 2023-04-25
Disable NuGet package caching

### LINKED ISSUE #1154: The build is broken in Windows because of testhost connection
**Summary**: 

Any [windows build](https://github.com/App-vNext/Polly/actions/runs/4783994471/jobs/8504927828?pr=1153) fails with the following error:

```
Starting test execution, please wait...
A total of 1 test files matched the specified pattern.
vstest.console process failed to connect to testhost process after 90 seconds. This may occur due to machine slowness, please set environment variable VSTEST_CONNECTION_TIMEOUT to increase timeout.
Data collector 'Blame' message: The specified inactivity time of 10 seconds has elapsed. Collecting hang dumps from testhost and its child processes.
Data collector 'Blame' message: Dumping 4816 - testhost.
Data collector 'Blame' message: All tests finished running, Sequence file will not be generated.
```

Seems to be a problem with nuget cache:
https://github.com/microsoft/vstest/issues/2108

We might disable this in Windows in the meantime to unblock.

@martincostello

martincostello:
Disable NuGet package caching as it appears to be causing build instability.

Resolves #1154.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,12 +37,12 @@ jobs:
     - name: Setup .NET SDK
       uses: actions/setup-dotnet@607fce577a46308457984d59e4954e075820f10a # v3.0.3
 
-    - name: Setup NuGet cache
-      uses: actions/cache@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
-      with:
-        path: ~/.nuget/packages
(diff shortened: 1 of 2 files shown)


martincostello:
Maybe I misunderstood the original issue - as this is still failing on Windows 😅

martintmk:
> Maybe I misunderstood the original issue - as this is still failing on Windows 😅

@martincostello My mistake, I hoped this would fix the problem, but seems it's indeed something else.

PR #1157, opened 2023-04-25
Introduce unit-tests for issues fixed in v8

martintmk:

Here we add unit tests for issues addressed in v8. This demonstrates the issue reporters the problem is fixed in v8.

### Confirm the following

### DIFF
--- a/src/Polly.Extensions.Tests/Issues/IssuesTests.OnCircuitBreakWithServiceProvider_796.cs
+++ b/src/Polly.Extensions.Tests/Issues/IssuesTests.OnCircuitBreakWithServiceProvider_796.cs
@@ -1,0 +1,70 @@
+using Microsoft.Extensions.DependencyInjection;
+using Polly.CircuitBreaker;
+using Polly.Extensions.DependencyInjection;
+using Polly.Registry;
+
+namespace Polly.Core.Tests.Issues;
+
+public class IssuesTests
+{
+    [Fact]
+    public async Task OnCircuitBreakWithServiceProvider_796()
+    {
+        var contextChecked = false;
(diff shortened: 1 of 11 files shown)


martintmk:
I think we can just merge this PR and keep the branch. It just adds more tests which is useful to have in main.

src/Polly.Extensions.Tests/Issues/IssuesTests.OnCircuitBreakWithServiceProvider_796.cs
```diff
@@ -0,0 +1,70 @@
+using Microsoft.Extensions.DependencyInjection;
+using Polly.CircuitBreaker;
+using Polly.Extensions.DependencyInjection;
+using Polly.Registry;
+
+namespace Polly.Core.Tests.Issues;
+
+public class IssuesTests
```
martincostello, line 8: Shall we make this partial too so we don't need to add it later?

martintmk: I tried but it complains about https://rules.sonarsource.com/csharp/RSPEC-2333.

Should be no problem to just add it later when we add more issues.

martincostello (approved):

src/Polly.Extensions.Tests/Issues/IssuesTests.OnCircuitBreakWithServiceProvider_796.cs
martintmk: @martincostello I just tried to add empty `IssuesTests` class as in other cases. It worked so it's non issue anyway.

martincostello (approved):

PR #1172, opened 2023-05-02
Add NuGet package validation

martincostello:
Add [NuGet package validation analysers](https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview).

Baselines were generated by setting [`GenerateCompatibilitySuppressionFile=true`](https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview#suppress-compatibility-errors) locally.

See https://github.com/App-vNext/Polly/issues/1141#issuecomment-1531059423.


martintmk (approved):
Looks good as it is.

PR #1174, opened 2023-05-02
Use records for resilience arguments

martintmk:

Just switching to records for all implementations of `IResilienceArguments` as it simplifies the implementation and reduces the unnecessary suppressions.

### Confirm the following


martincostello (approved):
Nice. The only downside is the `internal` constructor being removed means adding new properties would become a breaking change.

PR #1186, opened 2023-05-16
Preliminary API Review (2)

martintmk:

Just addressing some of the issues found when doing internal API review. We want to do some initial pass before we engage with broader audience.

- Fixed some docs.
- Dropping custom delegates and using `Func<> instead`.
- Dropping some APIs from `Outcome`.
- Embracing `System.Threading.Timeout.InfiniteTimeSpan` instead of custom API.

### Confirm the following


src/Polly.Core/Hedging/HedgingHandler.TResult.cs
```diff
@@ -23,8 +23,13 @@ public sealed class HedgingHandler<TResult>
     /// Gets or sets the hedging action generator that creates hedged actions.
     /// </summary>
     /// <remarks>
+    /// Make sure that the action returned by the generator represents a real asynchronous work, otherwise the hedging
+    /// engine will be blocked and parallel hedged actions won't ever be executed. The generator can return a <see langword="null"/> function.
+    /// In such case the hedging is not executed for that attempt.
```
martincostello, line 28:
```suggestion
    /// Make sure that the action returned by the generator represents real asynchronous work, otherwise the hedging
    /// engine will be blocked and parallel hedged actions won't ever be executed. The generator can return a <see langword="null"/> function.
    /// In such a case the hedging is not executed for that attempt.
```

src/Polly.Core/Hedging/VoidHedgingHandler.cs
```diff
@@ -22,8 +22,13 @@ public sealed class VoidHedgingHandler
     /// Gets or sets the hedging action generator that creates hedged actions.
     /// </summary>
     /// <remarks>
+    /// Make sure that the action returned by the generator represents a real asynchronous work, otherwise the hedging
+    /// engine will be blocked and parallel hedged actions won't ever be executed. The generator can return a <see langword="null"/> function.
+    /// In such case the hedging is not executed for that attempt.
```
martincostello, line 27:
```suggestion
    /// Make sure that the action returned by the generator represents real asynchronous work, otherwise the hedging
    /// engine will be blocked and parallel hedged actions won't ever be executed. The generator can return a <see langword="null"/> function.
    /// In such a case the hedging is not executed for that attempt.
```

martincostello (approved):

PR #1200, opened 2023-05-19
[Proposal] Use delegates instead of callback API

martintmk:

Going through the code together internally we are investigating the possibility of replacing the Callback API with simple delegates. 

So the syntax for callbacks will be:

Generic:

- Generators: `Func<Outcome<T>, Args, ValueTask<TValue>>`
- Predicates: `Func<Outcome<T>, Args, ValueTask<bool>>`
- Events: `Func<Outcome<T>, Args, ValueTask>`

Non-Generic:

- Generators: `Func<Outcome, Args, ValueTask<TValue>>`
- Predicates: `Func<Outcome, Args, ValueTask<bool>>`
- Events: `Func<Outcome, Args, ValueTask>`

So instead of:

``` csharp
var options = new RetryStrategyOptions();
options.ShouldRetry
   .HandleException<HttpRequestException>()
   .HandleResult<HttpResponseMessage>(response => !response.IsSuccessStatusCode);
```
We can embrace simple delegates with switch expresions:

``` csharp
var options = new RetryStrategyOptions();
options.ShouldRetry = (outcome, args) => outcome switch
{
    { Exception: HttpRequestException error } => PredicateResult.True,
    { Result: HttpResponseMessage response } when !response.IsSuccessStatusCode => PredicateResult.True,
    _ => PredicateResult.False
}
```

Note that `PredicateResult.False` is just a shorthand to `new ValueTask<bool>(false)`.

Upsides:

- We can drop the callback API and save considerable amount of code, meaning it's less maintenance for us in the future.
- Using the native C# features.
- API simplicity, no hidden magic.

Downsides:

- No more fluent API for callbacks. For users used to Polly V7 this is completely new.
- The configuration is not as readable (for some).
- Every delegate returns async `ValueTask` which is not used very often; however, the caller needs to return it.
- No extensibility for delegates. Once we expose the delegate it's fixed. The only extensibility can be done by adding new member to arguments. The customer cannot create convenience extensions for their specific scenarios. 

@martincostello , @joelhulen , @juraj-blazek , @geeknoid 

Your opinions please? Shall we keep the callback API or drop it and replace with simple delegates?

### Confirm the following

### DIFF
--- a/src/Polly.Core.Benchmarks/Internals/Helper.CircuitBreaker.cs
+++ b/src/Polly.Core.Benchmarks/Internals/Helper.CircuitBreaker.cs
@@ -13,8 +13,8 @@ internal static partial class Helper
         {
             PollyVersion.V7 =>
                 Policy
-                    .HandleResult(10)
-                    .Or<InvalidOperationException>()
+                    .HandleResult<HttpResponseMessage>(m => !m.IsSuccessStatusCode)
+                    .Or<HttpRequestException>()
                     .AdvancedCircuitBreakerAsync(0.5, TimeSpan.FromSeconds(30), 10, TimeSpan.FromSeconds(5)),
 
             PollyVersion.V8 => CreateStrategy(builder =>
--- a/src/Polly.Core.Benchmarks/Internals/Helper.Pipeline.cs
+++ b/src/Polly.Core.Benchmarks/Internals/Helper.Pipeline.cs
@@ -24,7 +24,12 @@ internal static partial class Helper
                 })
                 .AddTimeout(TimeSpan.FromSeconds(10))
                 .AddRetry(
-                    predicate => predicate.HandleException<InvalidOperationException>().HandleResult(10),
+                    (outcome, _) => outcome switch
+                    {
+                        (int result, _) when result == 10 => true,
+                        (_, InvalidOperationException) => true,
+                        _ => false
+                    },
                     RetryBackoffType.Constant,
                     3,
                     TimeSpan.FromSeconds(1))
--- a/src/Polly.Core.Benchmarks/Internals/Helper.Retry.cs
+++ b/src/Polly.Core.Benchmarks/Internals/Helper.Retry.cs
@@ -1,4 +1,5 @@
 using System;
+using Polly.Strategy;
 
 namespace Polly.Core.Benchmarks;
 
@@ -13,22 +14,24 @@ internal static partial class Helper
             PollyVersion.V7 =>
                 Policy
                     .HandleResult(10)
-                    .Or<InvalidOperationException>()
+                    .Or<HttpRequestException>()
                     .WaitAndRetryAsync(3, attempt => delay, (_, _) => Task.CompletedTask),
 
             PollyVersion.V8 => CreateStrategy(builder =>
             {
-                var options = new RetryStrategyOptions
+                builder.AddRetry(new RetryStrategyOptions
                 {
                     RetryCount = 3,
                     BackoffType = RetryBackoffType.Constant,
-                    BaseDelay = delay
-                };
-
-                options.ShouldRetry.HandleOutcome<int>((outcome, _) => outcome.Result == 10 || outcome.Exception is InvalidOperationException);
-                options.OnRetry.Register<int>(() => { });
-
-                builder.AddRetry(options);
+                    BaseDelay = delay,
+                    ShouldRetry = (outcome, _) => outcome switch
+                    {
+                        { Exception: HttpRequestException error } => PredicateResult.True,
+                        { Result: HttpResponseMessage response } when !response.IsSuccessStatusCode => PredicateResult.True,
+                        _ => PredicateResult.False
+                    },
+                    OnRetry = (_, _) => default
+                });
             }),
             _ => throw new NotSupportedException()
         };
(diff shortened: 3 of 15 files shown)


juraj-blazek:
While the existing API is closer to v7 and probably easier to use, it brings performance costs on hot path. If there are multiple predicate handlers registered, each of them might need to be evaluated to get the result, comparing to a single call to the delegate.

martincostello:
At first glance, this seems a lot less intuitive/simple to use for the _""Average .NET Developer™️""_ as it uses quite a few more-recent C# features that people may not be (as) familiar with compared to "vanilla" C#.

I'll have a think about it further, but I'd be worried that we'd push the performance slider too far and make Polly harder to adopt for simple use cases by the majority of the .NET community. This is just based on my personal perception of things - I might not be in tune with how up-to-date the _actual_ average community developer is.

martintmk:
> At first glance, this seems a lot less intuitive/simple to use for the _""Average .NET Developer™️""_ as it uses quite a few more-recent C# features that people may not be (as) familiar with compared to "vanilla" C#.
> 
> I'll have a think about it further, but I'd be worried that we'd push the performance slider too far and make Polly harder to adopt for simple use cases by the majority of the .NET community. This is just based on my personal perception of things - I might not be in tune with how up-to-date the _actual_ average community developer is.

One thing to note is that callback API allows high-performance scenarios:

``` csharp
var options = new RetryStrategyOptions<HttpResponseMessage>();

options.ShouldRetry.HandleOutcome((outcome, args) => outcome switch
{
    { Exception: HttpRequestException error } => true,
    { Result: HttpResponseMessage response } when !response.IsSuccessStatusCode => false,
    _ => false
});
```

I'll wait until we receive more feedback. Potentially, we should also engage the community for their opinion.

geeknoid:
I definitely prefer the delegate model since it is more idiomatic.

The existing approach is about providing some fan out semantics for these callbacks, and I believe this fan out should not be part of the infra itself. The infra needs to get an answer to a question and it does so by invoking a callback and that's that. If the callback needs to fan out in some way, that's its problem.

When reading through the current API, I was very confused by what I was looking at. Some of that was the naming of the types involved, but most was the model which seemed unexpected to me.

In terms of compat with V7, would it be possible to compose the fan out thing as a utility library, whose output would be the single delegate needed by the core infra.

joelhulen:
I agree with Martin Costello in that, for the "average .NET developer," moving to delegates in this fashion away from the familiar v7- callback APIs might not be intuitive. However, I think we can solve this with suitable examples. Still, I fear we're drifting away from our initial goal of making a clear and easy migration path from existing policies people have built to v8+ idioms. I think this might be solved via a helper library, as @geeknoid suggested.

martintmk:
@martincostello 

How about writing the predicate in a more simpler way?

Instead of:

``` csharp
var options = new RetryStrategyOptions();
options.ShouldRetry = (outcome, args) => outcome switch
{
    { Exception: HttpRequestException error } => PredicateResult.True,
    { Result: HttpResponseMessage response } when !response.IsSuccessStatusCode => PredicateResult.True,
    _ => PredicateResult.False
}
```

We can just use basic features:
``` csharp
var options = new RetryStrategyOptions();
options.ShouldRetry = (outcome, args) => outcome switch
{
   if (outcome.Exception is HttpRequestException)
   {
      return PredicateResult.True;
   }

   if (outcome.TryGetResult(out var response) && !response.IsSuccessStatusCode)
   {
      return PredicateResult.True;
   }

   return PredicateResult.False
}
```

We can also bake some basic API directly to `Outcome`. For example:

``` csharp
outcome.HasException<InvalidOperationException>();
outcome.HasException<InvalidOperationException>(exception => ...);
outcome.HasResult(result => ... );
```

This would help simplify the usage and give flexibility to either use switch expression or more basic features of C#.

martintmk:
> I think I'd be happy with that if there's a way for existing/less-experienced users to be able to migrate in a "v7-y" way where there aren't particularly interested in the high-perf API surface and want it to "Just Work™️" to solve their LOB app needs.
> 
> I'm just keen to avoid a "migration barrier" where people will avoid upgrading/adopting v8 because it "looks difficult" or is too much churn on their existing code.

How about just exposing convenience extensions for `ResilienceStrategyBuilder`, these can accept some simplified version of predicate builders and could map very similarly to V7? Basically, the API shape will be similar to what is now in V7. These extensions will just create delegate from these builders and store it directly to options.

The options will be still delegate based.

joelhulen:
> > I think I'd be happy with that if there's a way for existing/less-experienced users to be able to migrate in a "v7-y" way where there aren't particularly interested in the high-perf API surface and want it to "Just Work™️" to solve their LOB app needs.
> > I'm just keen to avoid a "migration barrier" where people will avoid upgrading/adopting v8 because it "looks difficult" or is too much churn on their existing code.
> 
> How about just exposing convenience extensions for `ResilienceStrategyBuilder`, these can accept some simplified version of predicate builders and could map very similarly to V7? Basically, the API shape will be similar to what is now in V7. These extensions will just create delegate from these builders and store it directly to options.
> 
> The options will be still delegate based.

I think this is a good balance of ease of migration and not hindering the performance gains of delegates. I think we'll just need to be clear in documenting the new way and have several examples of migrating from v7- strategies to v8+, and the benefits of doing so.

PR #1204, opened 2023-05-22
[Proposal] Introduce ResilienceStrategyBuilder<TResult>

martintmk:
# Introducing `ResilienceStrategyBuilder<T>` and `ResilienceStrategy<T>`

This PR introduces the `ResilienceStrategyBuilder<T>` and `ResilienceStrategy<T>` classes. They are designed for scenarios that require a strategy for handling a single result type. The infrastructure introduced in V8 supports this feature.

The `ResilienceStrategyBuilder<T>` class serves as a simple wrapper over the `ResilienceStrategyBuilder` class. Similarly, the `ResilienceStrategy<T>` class is a basic wrapper for `ResilienceStrategy` and doesn't support inheritance.

The core infrastructure is still centered on `ResilienceStrategy`, which prevents implementation fragmentation as we observed in V7. The V8 infrastructure facilitates the integration of non-generic strategies with generic ones. It also enables scenarios, such as sharing a single circuit breaker that handles multiple result types with numerous `ResilienceStrategy<T>` instances.

Upsides:
- Strong contract for resilience strategies that work with a single result type. (common use-case)

Downsides:
- Code duplication, although very limited. Mostly on some extensions that need to target both generic and non-generic resilience strategies.

## Usage

Here is a simple usage example:

```csharp
ResilienceStrategy<HttpResponseMessage> builder = new ResilienceStrategyBuilder<HttpResponseMessage>()
    .AddRetry(shouldRetry => shouldRetry.HandleException<HttpRequestException>())
    .AddTimeout()
    .Build();
```

## Additional Changes

- Strategy extensions employing generic options now operate with `ResilienceStrategyBuilder<T>`.
- Several extensions originally targeting `ResilienceStrategyBuilder` have been shifted to `ResilienceStrategyBuilder<T>`.
- All convenience extensions now target `ResilienceStrategyBuilder<T>`.
- Extensions for strategies that do not work with results (like `Timeout`,`RateLimiter`) have been duplicated to target both `ResilienceStrategyBuilder` and `ResilienceStrategyBuilder<T>`.
- All extensions to `ResilienceStrategyBuilder` now solely accept options. The goal behind the use of the non-generic `ResilienceStrategyBuilder` is for advanced scenarios where additional typing does not create significant inconvenience.

### DIFF
--- a/src/Polly.Core.Benchmarks/Benchmarks/HedgingBenchmark.cs
+++ b/src/Polly.Core.Benchmarks/Benchmarks/HedgingBenchmark.cs
@@ -6,7 +6,7 @@ namespace Polly.Benchmarks;
 
 public class HedgingBenchmark
 {
-    private ResilienceStrategy? _strategy;
+    private ResilienceStrategy<string>? _strategy;
 
     [GlobalSetup]
     public void Setup()
--- a/src/Polly.Core.Benchmarks/Internals/Helper.CircuitBreaker.cs
+++ b/src/Polly.Core.Benchmarks/Internals/Helper.CircuitBreaker.cs
@@ -13,13 +13,13 @@ internal static partial class Helper
         {
             PollyVersion.V7 =>
                 Policy
-                    .HandleResult(10)
+                    .HandleResult(Failure)
                     .Or<InvalidOperationException>()
                     .AdvancedCircuitBreakerAsync(0.5, TimeSpan.FromSeconds(30), 10, TimeSpan.FromSeconds(5)),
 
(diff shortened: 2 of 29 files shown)


martintmk:
I'll fix the build and add some tests if we agree with this proposal.

martincostello:
Seems reasonable to me 👍

juraj-blazek:
This makes sense, it will make the most common use cases easier to implement (like `HttpResponseMessage`-based resilience pipeline)

martincostello (approved):

PR #1233, opened 2023-05-26
Polly V8: Public API Review

martintmk:
### Details on the issue fix or feature implementation

**Great news!** We've been working hard and we're excited to finally introduce Polly V8 for public API review. The purpose of this PR is to outline the API signatures in the upcoming release. We welcome your comments, suggestions, and improvements to refine the API. Once we reach a consensus on the outcome, we will move forward to publish alpha packages for V8. 

## Summary of Changes

In the following table, we provide a comparison between V7 and V8:

| V7 | V8 | Remarks | 
| --- | --- | --- |
| `IAsyncPolicy` |  `ResiliencePipeline` | `ResiliencePipeline` can be utilized in both synchronous and asynchronous scenarios. |
| `ISyncPolicy` |  `ResiliencePipeline` | `ResiliencePipeline` can be utilized in both synchronous and asynchronous scenarios. |
| `IAsyncPolicy<T>` |  `ResiliencePipeline<T>` | `ResiliencePipeline<T>` can be utilized in both synchronous and asynchronous scenarios. |
| `ISyncPolicy<T>` |  `ResiliencePipeline<T>` | `ResiliencePipeline<T>` can be utilized in both synchronous and asynchronous scenarios. |
| `Context` |  `ResilienceContext` |  |
| `Policy` |  `ResiliencePipelineBuilder` | Utilized when the non-generic resilience strategy is required. |
| `Policy` |  `ResiliencePipelineBuilder<T>` | Utilized when the generic resilience strategy is required.  |

### Building `ResiliencePipeline<TResult>` that handles single type of the result

Below is an example of how to create a `ResiliencePipeline<HttpResponseMessage>` using convenience API:

```csharp
ResiliencePipeline<HttpResponseMessage> strategy = new ResiliencePipelineBuilder<HttpResponseMessage>()
    .AddRetry(builder => builder
        .Handle<HttpRequestException>()
        .HandleResult(r => r.IsSuccessStatusCode))
    .AddTimeout(TimeSpan.FromSeconds(5))
    .Build();
```

In addition, for maximum flexibility, you can utilize options as shown below:

```csharp
ResiliencePipeline<HttpResponseMessage> strategy = new ResiliencePipelineBuilder<HttpResponseMessage>()
    .AddRetry(new RetryStrategyOptions<HttpResponseMessage>
    {
        ShouldRetry = args =>
        {
            if (args .Exception is HttpRequestException)
            {
                return PredicateResult.True;
            }

[...]

PR #1235, opened 2023-05-29
Drop Outcome and cleanup the options

martintmk:

The `Outcome` is a remnant from Callback API and does not add much value now. Dropping it allows us to cleanup and simplify the non-generic options.

These can now just implement their generic counterparts.

https://github.com/App-vNext/Polly/pull/1233/files#r1206898556

### Confirm the following


src/Polly.Core/Strategy/Outcome.cs
```diff
@@ -76,18 +76,24 @@ public bool TryGetResult(out object? result)
...
-        : new Outcome<TResult>((TResult)Result!);
+    internal TResult GetResultOrRethrow()
+    {
+        ExceptionDispatchInfo?.Throw();
+        return Result!;
+    }
+
+    internal Outcome<object> AsObjectOutcome() => Convert<object>();
+
+    internal Outcome<T> Convert<T>() => (ExceptionDispatchInfo != null)
```
martincostello, line 96: I'm not sure what we should call it, but Convert doesn't seem the right name to me for this. It seems to suggest it turns it into a TResult.

martintmk: Maybe just `AsOutcome` and `AsOutcome<T>`?

martincostello: That seems reasonable to me.

martincostello (approved):

PR #1236, opened 2023-05-30
[Proposal] Introduce ResilienceStrategyBuilderBase

martintmk:

One thing I wasn't a big fan of was the duplications of extensions for non-reactive strategies such as `Timeout` or `RateLimiter`.

In this PR I am introducing `ResilienceStrategyBuilderBase` class. The extensibility author can use is to target extensions that work for both generic and non-generic strategies. 

This way full set of extensions are always exposed for both generic and non-generic builders and there are no duplications.

The extension bellow will be visible for both builder types:

``` csharp
public static TBuilder AddTimeout<TBuilder>(this TBuilder builder, TimeoutStrategyOptions options)
    where TBuilder : ResilienceStrategyBuilderBase
{
    builder.AddStrategy(context => new TimeoutResilienceStrategy(options, context.TimeProvider, context.Telemetry), options);
    return builder;
}
```

I have dropped a bunch of duplicate extensions.

What do you think about this idea?

### Confirm the following


martincostello:
This seems reasonable to me.

PR #1244, opened 2023-06-01
[Proposal] Drop internal visibility between assemblies

martintmk:

The internals of `Polly.Core` are visible by `Polly.Extensions`. This feels like cheating because these are two independent packages that can have different version and therefore use diferent APIs. Using internals hides this fact and doesn't protect us against accidental API changes.

We have tooling in place to fail when any incompatible changes are introduced (#1172). I think it's better to expose APIs used by `Polly.Extensions` as public and hide it from common users by using `[EditorBrowsable]` attribute.

This way, any changes to this pub-internal API will be reported as a breaking change and will fail the build.

### Confirm the following


src/Polly.Extensions/Telemetry/TelemetryUtil.cs
```diff
@@ -0,0 +1,6 @@
+namespace Polly.Extensions.Telemetry;
+
+internal class TelemetryUtil
```
martincostello, line 3:
```suggestion
internal static class TelemetryUtil
```

martincostello (approved):
Seems reasonable to me.

PR #1254, opened 2023-06-06
Enable dynamic reloads

### LINKED ISSUE #1241: `ResilienceStrategyRegistry` should support reloadable resilience strategies
The [ResilienceStrategyRegistry](https://github.com/App-vNext/Polly/blob/main/src/Polly.Core/Registry/ResilienceStrategyRegistry.cs) should support reloadable resilience strategies. 

It means that the strategy that was dynamically retrieved using registered builder can be recreated if external signal changes:

``` csharp 
var timeout = TimeSpan.FromSeconds(1);
var registry = new ResilienceStrategyRegistry<string>();
var timeoutChanged = new CancellationTokenSource();

registry.TryAddBuilder("my-timeout", (key, builder) =>
{
    builder
        .AddTimeout(new TimeoutStrategyOptions
        {
            // provide dynamic value for the property
            Timeout = timeout
        });

    // register a reload token provider that is raised when a value changes
    builder.Properties.Set(
        new ResiliencePropertyKey<Func<CancellationToken>>("Polly.ReloadTokenProvider"),
        () => timeoutChanged.Token);
});

timeout = TimeSpan.FromSeconds(5);
timeoutChanged.Cancel();
```
This will be opt-in feature not directly exposed because it's for advanced scenarios only. Practical usage is microservice that uses resilience and allows changing the resilience configuration at runtime.

We could also expose convenience `EnableReloads` method when using `AddResilienceStrategy` method:

``` csharp 
serviceCollection.AddResilienceStrategy("dummy", (builder, context) =>
{
    var options = context.ServiceProvider.GetRequiredService<IOptionsMonitor<TimeoutStrategyOptions>>();

    // this call enables dynamic reloads for the dummy strategy
    context.EnableReloads(() =>
    {
        var source = new CancellationTokenSource();
        IDisposable? disposable = null;
        disposable = options.OnChange(_ =>
        {
            source.Cancel();
            disposable?.Dispose();
        });

        // returns a token that is cancelled whenever these options are changed
        return source.Token;
    });

    builder.AddTimeout(options.CurrentValue);
});

var strategy = serviceCollection.BuildServiceProvider().GetRequiredService<ResilienceStrategyProvider<string>>().Get("dummy");
```

In the above example the resulting `strategy` supports dynamic reloads and is automatically reloaded whenever the `TimeoutStrategyOptions` are changed.

martintmk:

Closes #1241

This PR adds opt-in support for reloadable pipelines. Dynamic reloads are enabled when new  `ConfigureBuilderContext.EnableReloads` method is called. 

Caller needs to provide a factory that returns a `CancellationToken`. When the token is cancelled the strategy is reloaded.

I have also considered using [`IChangeToken`](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/change-tokens?view=aspnetcore-7.0) which is more natural to propagate changes, however, we would have to add a new dependency to `Microsoft.Extensions.Primitives` which seemed overkill for such small addition.

### Confirm the following


src/Polly.Core/Registry/ConfigureBuilderContext.cs
```diff
@@ -28,4 +28,21 @@ internal ConfigureBuilderContext(TKey strategyKey, string builderName, string st
     /// Gets the string representation of strategy key for the strategy being created.
     /// </summary>
     public string StrategyKeyString { get; }
+
+    internal Func<CancellationToken>? ReloadTokenProducer { get; private set; }
+
+    /// <summary>
+    /// Enables the dynamic reloads for the strategy produced by current builder.
```
martincostello, line 35:
```suggestion
    /// Enables dynamic reloading for the strategy produced by current builder.
```

src/Polly.Core/Registry/ConfigureBuilderContext.cs
```diff
@@ -28,4 +28,21 @@ internal ConfigureBuilderContext(TKey strategyKey, string builderName, string st
...
+    /// </summary>
+    /// <param name="reloadTokenProducer">The producer of <see cref="CancellationToken"/> that is triggered when change occurs.</param>
+    /// <remarks>
+    /// The <paramref name="reloadTokenProducer"/> should always return a new <see cref="CancellationToken"/> instance when invoked otherwise
+    /// the reload infra will stop listening for changes.
```
martincostello, line 40:
```suggestion
    /// the reload infrastructure will stop listening for changes.
```

src/Polly.Core/Strategy/ReloadableResilienceStrategy.cs
```diff
@@ -0,0 +1,74 @@
+using System;
+using System.Threading.Tasks;
+
+namespace Polly.Strategy;
+
+internal class ReloadableResilienceStrategy : ResilienceStrategy
```
martincostello, line 6:
```suggestion
internal sealed class ReloadableResilienceStrategy : ResilienceStrategy
```

src/Polly.Core/Registry/ConfigureBuilderContext.cs
```diff
@@ -28,4 +28,21 @@ internal ConfigureBuilderContext(TKey strategyKey, string builderName, string st
     /// Gets the string representation of strategy key for the strategy being created.
     /// </summary>
     public string StrategyKeyString { get; }
+
+    internal Func<CancellationToken>? ReloadTokenProducer { get; private set; }
+
+    /// <summary>
+    /// Enables dynamic reloading of strategy retrieved from <see cref="ResilienceStrategyRegistry{TKey}"/>.
```
martincostello, line 35:
```suggestion
    /// Enables dynamic reloading of the strategy retrieved from <see cref="ResilienceStrategyRegistry{TKey}"/>.
```

martincostello (approved):

PR #1263, opened 2023-06-08
Specify default action generator for hedging

martintmk:

I was investigating the [original hedging engine](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Resilience/Polly/Hedging/HedgingEngine.cs) and [standard hedging pipeline](https://github.com/dotnet/extensions/blob/main/src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/HedgingHttpClientBuilderExtensions.cs#L17) to find a ways to simplify the hedging configuration. A lot of boilerplate in hedging is around specifying the hedging action generator. 

However, we can actually allow the default for this generator. In this case the hedging just executes the original callback again ( similar to retries). This can be done by providing the original callback to the generator arguments. 

Standard hedging handler can seamlessly integrate into this enhancement by:

- Passing a callback that reads the data from resilience context. (for example cloned request message)
- Taking advantage of the enhanced `HedgingActionGeneratorArguments`

The simple API addition above will greatly simplify the code in `Microsft.Extensions.Http.Resilience` library.

On top of that I have also changed the signature of the generator action from `Task<TResult>` to `Task<Outcome<TResult>>`. This plays nicely with the overal infrastructure we have. Hedged actions can be optimized to not throw exceptions.

### Confirm the following

### DIFF
--- a/src/Polly.Core.Tests/Hedging/HedgingActions.cs
+++ b/src/Polly.Core.Tests/Hedging/HedgingActions.cs
@@ -32,31 +32,34 @@ internal class HedgingActions
         };
     }
 
-    public Func<HedgingActionGeneratorArguments<string>, Func<Task<string>>?> Generator { get; }
+    public Func<HedgingActionGeneratorArguments<string>, Func<ValueTask<Outcome<string>>>?> Generator { get; }
 
-    public Func<HedgingActionGeneratorArguments<string>, Func<Task<string>>?> EmptyFunctionsProvider { get; } = args => null;
+    public Func<HedgingActionGeneratorArguments<string>, Func<ValueTask<Outcome<string>>>?> EmptyFunctionsProvider { get; } = args => null;
 
-    public List<Func<ResilienceContext, Task<string>>> Functions { get; }
+    public List<Func<ResilienceContext, ValueTask<Outcome<string>>>> Functions { get; }
 
-    private async Task<string> GetApples(ResilienceContext context)
+    private async ValueTask<Outcome<string>> GetApples(ResilienceContext context)
     {
         await _timeProvider.Delay(TimeSpan.FromSeconds(10), context.CancellationToken);
-        return "Apples";
+        return "Apples".AsOutcome();
     }
 
-    private async Task<string> GetPears(ResilienceContext context)
+    private async ValueTask<Outcome<string>> GetPears(ResilienceContext context)
     {
         await _timeProvider.Delay(TimeSpan.FromSeconds(3), context.CancellationToken);
-        return "Pears";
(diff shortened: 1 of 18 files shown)


src/Polly.Core.Tests/Hedging/HedgingActions.cs
```diff
@@ -32,31 +32,34 @@ public HedgingActions(TimeProvider timeProvider)
...
     {
         await _timeProvider.Delay(TimeSpan.FromSeconds(2), context.CancellationToken);
-        return "Oranges";
+        return "Oranges".AsOutcome();
     }
 
-    public static Func<HedgingActionGeneratorArguments<string>, Func<Task<string>>?> GetGenerator(Func<ResilienceContext, Task<string>> task) => args => () => task(args.Context);
+    public static Func<HedgingActionGeneratorArguments<string>, Func<ValueTask<Outcome<string>>>?> GetGenerator(Func<ResilienceContext, ValueTask<Outcome<string>>> task)
+    {
+        return args => () => task(args.Context);
```
martincostello, line 61:
```suggestion
        return args => static () => task(args.Context);
```

src/Polly.Core/Hedging/HedgingActionGeneratorArguments.TResult.cs
```diff
@@ -1,9 +1,11 @@
...
 /// <summary>
-/// Represents arguments used in <see cref="HedgingHandler{TResult}.HedgingActionGenerator"/>.
+/// Represents arguments used in the hedging resilience strategy.
 /// </summary>
 /// <typeparam name="TResult">The type of the result.</typeparam>
 /// <param name="Context">The context associated with the execution of a user-provided callback.</param>
 /// <param name="Attempt">The zero-based hedging attempt number.</param>
-public readonly record struct HedgingActionGeneratorArguments<TResult>(ResilienceContext Context, int Attempt);
+/// <param name="Callback">The callback passed to hedging strategy.</param>
+///
```
martincostello, line 10:
```suggestion
```

martintmk:
In the follow-up i will do some internal cleanup around hedging. As we can cleanup some internals.

src/Polly.Core/Outcome.cs
```diff
@@ -1,5 +1,6 @@
 #pragma warning disable CA1815 // Override equals and operator equals on value types
 
+using System;
```
martincostello, line 3: Needed?

martintmk: Nope, but I moved this class to TestUtils

martincostello (approved):

PR #1273, opened 2023-06-09
Fix CI version in main

### LINKED ISSUE #1270: Fix CI version in main to not use stable versions if not a tag
Builds in the main branch are all producing a package version of `8.0.0`, rather than adding some sort of prerelease suffix ([example](https://github.com/App-vNext/Polly/actions/runs/5213437307/jobs/9408416968#step:5:250)).

Need to investigate why this is and resolve.

martincostello:
- Use `v8alpha` tag for pre-release builds in `main`.
- Use the default file name for GitVersion configuration.
- Include the build version in the assembly file version so we can tell build-to-build.
- Remove logging of the informational version that isn't what we use.
- Add `.vscode/settings.json` file the new C# Dev Kit keeps creating.
- Use Minimal MSBuild verbosity for `dotnet clean` to match other targets.

Resolves #1270.


GitVersion.yml
```diff
@@ -0,0 +1,4 @@
+next-version: 8.0.0
+branches:
+  main:
+    tag: v8alpha
```
martincostello, line 4: I've updated this based on our conversation yesterday, though the `v8` prefix seems redundant to me.

martincostello:
Hmm - the package version is still [8.0.0](https://github.com/App-vNext/Polly/actions/runs/5220862538/jobs/9424448825#step:5:407)...

PR #1314, opened 2023-06-16
Enhance `OnHedgingArguments`

martintmk:

`OnHedgingArguments` is now a class because and also contains contains missed `Context` property. The class is fine in this case because it's not on a hot path and we can still enable pooling later. 

The main reason for this change is to allow setting properties on `OnHedgingArguments` later for more advanced scenarios.
The action generator argument now also contains the `PrimaryContext` property which allows passing data between primary and secondary hedged tasks.
 
### Confirm the following


src/Polly.Core/Hedging/HedgingActionGeneratorArguments.cs
```diff
@@ -4,10 +4,15 @@ namespace Polly.Hedging;
 /// Represents arguments used in the hedging resilience strategy.
 /// </summary>
 /// <typeparam name="TResult">The type of the result.</typeparam>
-/// <param name="Context">The context associated with the execution of a user-provided callback.</param>
+/// <param name="PrimaryContext">The primary resilience context.</param>
+/// <param name="Context">The context associated with the execution of a user-provided callback that is cloned from the primary context.</param>
```
martincostello, line 8: Could we prefix this to give it a complimentary name to `PrimaryContext`? (`SecondaryContext`?)

PR #1319, opened 2023-06-19
Alpha fixes and improvements

martintmk:

Fixes and improvements based on some testing and playing with the alpha1 package.

- Timeout strategy disposing `CancellationTokenSource` too late
- `ResultFormatter` not being used for resilience events.
- The hedging should not dispose the resources it holds in background. It should do it as a post-execute action.
- Unifying the predicate arguments, all have the same `<Strategy>PredicateArguments` pattern. The delegate's name is always `ShouldHandle`.
- Setting default `ShouldHandle` callback for fallback.
- Changing argument structs for resilience events to classes. These were boxed anyway and we can still turn the pooling later one. All these are not executed on the hot path, only when resilience event occurs.
- Stack trace was missing for some exceptions.

Contributes to #1301

### Confirm the following


src/Polly.Core/Hedging/Controller/HedgingExecutionContext.cs
```diff
@@ -94,7 +94,7 @@ public void Complete()
         // We are intentionally doing the cleanup in the background as we do not want to
         // delay the hedging.
         // The background cleanup is safe. All exceptions are handled.
-        _ = CleanupInBackgroundAsync();
+        return CleanupInBackgroundAsync();
```
martincostello, line 97: Should we remove this comment and rename the method of it's no longer done in the background?

src/Polly.Core/Utils/ExceptionUtilities.cs
```diff
@@ -0,0 +1,36 @@
+﻿namespace Polly.Utils;
+
+#pragma warning disable S3011 // Reflection should not be used to increase accessibility of classes, methods, or fields
+
+internal static class ExceptionUtilities
+{
+#if !NET6_0_OR_GREATER
+    private static readonly FieldInfo StackTraceString = typeof(Exception).GetField("_stackTraceString", BindingFlags.NonPublic | BindingFlags.Instance)!;
+    private static readonly Type TraceFormat = Type.GetType("System.Diagnostics.StackTrace")!.GetNestedType("TraceFormat", BindingFlags.NonPublic)!;
+    private static readonly MethodInfo TraceToString = typeof(StackTrace).GetMethod("ToString", BindingFlags.NonPublic | BindingFlags.Instance, null, new[] { TraceFormat }, null)!;
```
martincostello, line 10: Are these two StackTrace types different? Otherwise can just use `typeof` twice?

src/Polly.Core/Utils/ExceptionUtilities.cs
```diff
@@ -0,0 +1,36 @@
...
+#else
+        SetStackTrace(exception, new StackTrace());
+#endif
+        return exception;
+    }
+
+#if !NET6_0_OR_GREATER
+    private static void SetStackTrace(this Exception target, StackTrace stack)
+    {
+        var getStackTraceString = TraceToString.Invoke(stack, new[] { Enum.GetValues(TraceFormat).GetValue(0) });
```
martincostello, line 32: Could this array be cached for re-use across invocations?

martincostello (commented):
Any tests we can add/extend to verify stack traces don't get lost?

src/Polly.Core/Hedging/Controller/HedgingExecutionContext.cs
martintmk: I have inlined `CleanupInBackgroundAsync` here, it was small anyway.

src/Polly.Core/Utils/ExceptionUtilities.cs
martintmk: Yes, simplified.

src/Polly.Core/Hedging/HedgingResilienceStrategy.cs
```diff
@@ -37,6 +39,7 @@ internal sealed class HedgingResilienceStrategy<T> : ResilienceStrategy
 
     public EventInvoker<OnHedgingArguments>? OnHedging { get; }
 
+    [ExcludeFromCodeCoverage] // coverlet issue
```
martincostello, line 42: Is there a specific issue we could link to?

martincostello (approved):

Issue #1329, opened 2023-06-20
Simplifying predicate configuration for resilience strategies in V8 


martintmk:
In Polly V8, resilience strategies are configured using options and delegates.

Here are some patterns you can use to accomplish this:

``` csharp
// Switch-based pattern (recommended)
builder.AddRetry(new RetryStrategyOptions
{
    ShouldHandle = args => args.Exception switch
    {
        InvalidOperationException => PredicateResult.True,
        _ => PredicateResult.False
    }
});

// Simplified pattern
builder.AddRetry(new RetryStrategyOptions
{
    ShouldHandle = args =>
    {
        if (args.Exception is InvalidOperationException)
        {
            return PredicateResult.True;
        }

        return PredicateResult.False;
    }
});

// Inline pattern
builder.AddRetry(new RetryStrategyOptions
{
    ShouldHandle = args => new ValueTask<bool>(args.Exception is InvalidOperationException)
});
```

For handling results, the pattern is slightly more complex:

``` csharp
builder.AddRetry(new RetryStrategyOptions<HttpResponseMessage>
{
    ShouldHandle = args => args switch
    {
        { Exception: InvalidOperationException } => PredicateResult.True,
        { Result: HttpResponseMessage response } when !response.IsSuccessStatusCode => PredicateResult.True,
        _ => PredicateResult.False
    }
});
```

I'm currently exploring ways to streamline the configuration of predicates. Here are some options I've considered:

# Option 1: Introduction of new Constructors

This would involve adding new sets of constructors for every generic option:

``` csharp
public class RetryStrategyOptions<TResult> : ResilienceStrategyOptions
{
    public RetryStrategyOptions()
    {
        // ...
    }

    public RetryStrategyOptions(PredicateBuilder<TResult> shouldHandle)
    {
        // ...
    }

    public RetryStrategyOptions(Func<OutcomeArguments<TResult, RetryPredicateArguments>> shouldHandle)
    {
        // ...
    }
```

With these, you can now configure strategies like so:

``` csharp
// Simple
builder.AddRetry(new RetryStrategyOptions<HttpResponseMessage>(args => args.Exception is InvalidOperationException));

// With results
builder.AddRetry(new RetryStrategyOptions<HttpResponseMessage>(args =>
{
    if (args.Exception is InvalidOperationException)
    {
        return true;
    }

    if (args.Result != null && !args.Result.IsSuccessStatusCode)
    {
        return false;
    }

    return false;
}));

// Using predicate builder
[...]

PR #1332, opened 2023-06-21
Allow implicit conversion of `PredicateBuilder` to delegates

martintmk:

Contributes to #1329

### Confirm the following


bench/Polly.Core.Benchmarks/PredicateBenchmark.cs
```diff
@@ -0,0 +1,47 @@
...
+            .HandleResult(r => r.StatusCode == HttpStatusCode.InternalServerError)
+            .Handle<HttpRequestException>()
+            .Handle<InvalidOperationException>(e => false)
+    };
+
+    [Benchmark(Baseline = true)]
+    public ValueTask<bool> Predicate_SwitchExpression()
+    {
+        return _delegate.ShouldHandle(_args);
+
```
martincostello, line 39:
```suggestion
```

src/Polly.Core/PredicateBuilder.TResult.cs
```diff
@@ -0,0 +1,140 @@
...
+        return Add(outcome => outcome.Exception?.InnerException is TException innerException && predicate(innerException));
+    }
+
+    /// <summary>
+    /// Adds a predicate for handling results.
+    /// </summary>
+    /// <param name="predicate">The predicate function to use for handling the result.</param>
+    /// <returns>The same instance of the <see cref="PredicateBuilder{TResult}"/> for chaining.</returns>
+    public PredicateBuilder<TResult> HandleResult(Func<TResult, bool> predicate)
+        => Add(outcome => outcome.TryGetResult(out var result) && predicate(result!));
```
martincostello, line 69: Can we annotate `TryGetResult` so the compiler knows `result` won't be `null` if it returns `true`? Then we won't need the `!`.

src/Polly.Core/PredicateBuilder.TResult.cs
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    /// <summary>
+    /// Builds the predicate.
+    /// </summary>
+    /// <returns>An instance of predicate delegate.</returns>
+    /// <exception cref="InvalidOperationException">Thrown when no predicates were configured using this builder.</exception>
+    /// <remarks>
+    /// The returned predicate will return <see langword="true"/> if any of the configured predicates return <see langword="true"/>.
+    /// Please be aware of performance penalty if you register too many predicates with this builder. In such case, it's better to create your own predicate
```
martincostello, line 91: Here and other presences.

```suggestion
    /// Please be aware of the performance penalty if you register too many predicates with this builder. In such case, it's better to create your own predicate
```

src/Polly.Core/PredicateBuilder.TResult.cs
martintmk: The problem is that we should also allow `nulls` there.

You can do `Outcome.FromResult<object?>(null)`. It's perfectly valid result.

martincostello (approved):

PR #1334, opened 2023-06-21
Validate NuGet package signature

martincostello:
A step to validate the NuGet package signature before pushing to NuGet.org with `nuget verify`.

Will remove the changes I made to the workflow for testing once verified as working.


martincostello:
Forgot to do this in a branch so the secrets were available...

Issue #1333, opened 2023-06-21
The new nuget package Polly 7.2.4 is invalid or cannot be verified on this platform.


valentinmarinro:
Summary:
On Azure Pipelines build step I get this error for 7.2.4 Nuget package Polly 7.2.4 The package signature is invalid or cannot be verified on this platform.
But version 7.2.3 works correctly.

Expected behavior:
Signed package.

Actual behaviour:
Not signed package.

Steps / Code to reproduce the problem:
While trying to restore nuget package , I'm getting the following error:

Package 'Polly 7.2.4' from source '[https://api.nuget.org/v3/index.json'](https://api.nuget.org/v3/index.json%27): The package signature is invalid or cannot be verified on this platform.

martincostello:
Duplicate of https://github.com/App-vNext/Polly/issues/1323#issuecomment-1598521023.

On what operating system and tooling are you trying to download the package on and then validate the package with?

valentinmarinro:
Hi, it is on Windows and the agent in pipeline is defines like this:

pool:
  name: Default
  demands:
  - Agent.Name -equals AgentName
  - msbuild
  - visualstudio
  - vstest

valentinmarinro:
We also check if the certificates are installed, and they are for both NuGet.org and .NET Foundation

martincostello:
And how exactly are you attempting to validate the package signature?

Note that certificates won't be _installed_ into a computer by default as they are not trusted root certificates like those used for TLS certificates, but [Windows Authenticode certificates](https://learn.microsoft.com/windows-hardware/drivers/install/authenticode).

Do you validate any other NuGet packages' signatures in the same workflow, such as the .NET ones (e.g. [System.Text.Json](https://www.nuget.org/packages/System.Text.Json/8.0.0-preview.5.23280.8)), as they are similarly Authenticode signed?

![image](https://github.com/App-vNext/Polly/assets/1439341/100dab09-d8b0-462e-a631-55ae3a1f02c9)

valentinmarinro:
Yes, we are using System.Text.Json (latest stable version) on all our microservices that uses the same process and pipelines, also we have the flag -warnaserror enabled

valentinmarinro:
These are some examples of the libraries we have in the same process where Polly fails that are signed, one from Microsoft and one thirdparty from BouncyCastle:

[...]

Issue #1323, opened 2023-06-19
The new nuget package Polly 7.2.4 is not signed


elina-zh:
**Summary**: 
Nuget package Polly 7.2.4 is not signed

**Expected behavior**: 
Signed package.

**Actual behaviour**:
Not signed package.

**Steps / Code to reproduce the problem**: 
While trying to restore nuget package from the artifactory, I'm getting the following error:

Package 'Polly 7.2.4' from source '[https://api.nuget.org/v3/index.json'](https://api.nuget.org/v3/index.json%27): The package signature is invalid or cannot be verified on this platform.

martincostello:
What operating system are you trying to validate the package signature on?

martincostello:
As shown in the below screenshot, if I inspect the NuGet package with [NuGet Package Explorer](https://github.com/NuGetPackageExplorer/NuGetPackageExplorer) on Windows, the package is signed with a valid signature:

![image](https://github.com/App-vNext/Polly/assets/1439341/0e0eb490-8468-46e4-b239-5afd9b21ac56)

elina-zh:
> What operating system are you trying to validate the package signature on?

Linux

martincostello:
I think you have an Artifactory issue here - the NuGet libraries only support package signature validation on Windows when a package is Authenticode signed. The exception is being [thrown from the NuGet client](https://github.com/mzalves/NuGet.Client/blob/b4c39f96856a62e1805f1236044778f4ca48f9f8/src/NuGet.Core/NuGet.Packaging/Signing/Verification/PackageSignatureVerifier.cs#L66-L70) because the code to validate package signatures is [conditionally complied](https://github.com/NuGet/NuGet.Client/blob/488693efb590056b7f9e07d3a376d17c60d7f9e6/src/NuGet.Core/NuGet.Packaging/PackageArchiveReader.cs#L392-L473) and returns `false`/`null` as appropriate on non-Windows operating systems.

See https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/1219 for some more context on this.

The package itself _is_ signed.

martincostello:
FYI, older versions of the .NET SDK fail to validate the package signature. More recent versions do not. See here for more information: https://github.com/App-vNext/Polly/issues/1333#issuecomment-1600955911

PR #1335, opened 2023-06-21
Validate NuGet package signatures

martincostello:
A step to validate the NuGet package signature before pushing to NuGet.org with `nuget verify`.

Added in response to #1323 and #1333.


martincostello:
`dotnet nuget verify $package` is [behaving as expected](https://github.com/App-vNext/Polly/actions/runs/5333046464/jobs/9663166900?pr=1335#step:6:81).

martintmk (approved):

PR #1336, opened 2023-06-21
Add net6.0 and net7.0 to Polly

martincostello:
Add `net6.0` and `net7.0` TFMs to Polly to match the other assemblies.


martintmk (approved):
LGTM, we might actually trim some targets before official release.

PR #1359, opened 2023-06-27
Simplify the logging

martintmk:

I am dropping some #ifdef for older framework since in #1144 we are unifying the package dependencies anyway.

### Confirm the following

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -44,6 +44,6 @@
   </ItemGroup>
   <ItemGroup Condition="!$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'netcoreapp3.1'))">
     <PackageVersion Include="Microsoft.Extensions.Options" Version="2.2.0" />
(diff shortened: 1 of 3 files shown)


Directory.Packages.props
```diff
@@ -44,6 +44,6 @@
   </ItemGroup>
   <ItemGroup Condition="!$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'netcoreapp3.1'))">
     <PackageVersion Include="Microsoft.Extensions.Options" Version="2.2.0" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.0" />
```
martincostello, line 47: This won't cause an issue for any of the older TFMs?

martincostello (approved):

Directory.Packages.props
martintmk: It shouldn't and if we find so it will be fixed in 8.0.0 version that we will reference on every TFM anyway.

Issue #1365, opened 2023-06-28
Polly v8 API feedback


martincostello:
I've spent a few hours today upgrading an internal application which uses Polly from the v7 policies to v8 resilience strategies and encountered a number of rough edges and/or difficulties. This issue catalogs those at a high-level for discussion as to whether or not these are issues or missing functionality that we would wish to fix/add as part of the v8 release. I can add more concrete details/repro if needed for the specific items.

1. `ResilienceProperties` has a dictionary indexer for reading and writing properties, but it is an explicit interface definition so requires casting to a dictionary to use. It would be much more convenient to make it implicit so you can do things like `arguments.Properties["foo"]`.
2. It isn't possible to share a `CircuitBreakerManualControl` across multiple strategies - we have a use case where we manually isolate circuit breakers and it isn't possible to have a single instance available to any circuit breaker strategy to observe (thing similarly to sharing a `CancellationToken` on multiple operations). I've tried pooling them and iterating through the set when I want to open/close them, but this doesn't seem to work with what I've tried so far.
3. In the above case, if I try and manually isolate the manual control but there doesn't happen to have been a strategy created that uses the builder yet, it throws an exception saying it hasn't been initialized.
4. Creating a strategy of policies composed of other potentially existing policies seems to be a bit cumbersome. A rough example of what I'm trying to do is below, and needing to create a new builder to get a policy (because I can't make them directly because of internal constructors) is a bit weird.

```csharp
// This is just illustrative, it doesn't compile as-is

private readonly ConcurrentDictionary<string, ResilienceStrategy> _registry = new();
private readonly ConcurrentDictionary<string, CircuitBreakerManualControl> _isolators = new();

public ResilienceStrategy GetStrategy(string token, ApiEndpointOption endpoint, string resource)
{
    var builder = new ResilienceStrategyBuilder();

    ApplyRetries(builder, endpoint);
    ApplyCircuitBreaker(builder, endpoint, resource);
    ApplyTimeout(builder, endpoint);

    return builder.Build();
}

private void ApplyRetries(ResilienceStrategyBuilder builder, ApiEndpointOption endpoint)
{
    if (endpoint.Retries > 0)
    {
        var strategy = _registry.GetOrAdd($"{endpoint.Name}-Retry", (key) => new ResilienceStrategyBuilder().AddRetry(new()
        {
            BackoffType = RetryBackoffType.ExponentialWithJitter,
            BaseDelay = endpoint.RetryDelaySeed,
[...]

PR #1380, opened 2023-07-03
Introduce `ResilienceContext.OperationKey`

martintmk:

This PR introduces `ResilienceContext.OperationKey` that can be assigned to each `ResilienceContext`. It's immutable property that is specified at the time of the `ResilienceContext` creation:

``` csharp
var context= ResilienceContext.Get("my-operation");
var key = context.OperationKey; // my-operation
```

Operation key flows to telemetry and is included in logs and metrics. The main scenario is to distinguish the particular operation at call site and to improve the telemetry. It should have low cardinality.

This also fills the feature gap we had between Polly V7 and Polly V8.

Contributes to #1365 

### Confirm the following


martincostello (approved):

martincostello:
Are you planning on making any other new additions today? If not (or after them) I can look at publishing an alpha.5 with this change and the manual control changes and re-test in my app.

martintmk:
> Are you planning on making any other new additions today? If not (or after them) I can look at publishing an alpha.5 with this change and the manual control changes and re-test in my app.

Nope, I think I am mostly done with V8 (until API review). I'll only address issues as they come based on the feedback.

PR #1383, opened 2023-07-04
Add support for `PartitionedRateLimiter`

martintmk:

Implements the following proposal that adds support for `PartitionedRateLimiter`:

https://github.com/App-vNext/Polly/issues/1365#issuecomment-1618382828

### Confirm the following


martincostello (approved):
The changes look ok, but from the tests I don't understand how the rate limiter would partition on the HTTP context, for example, where the limit is intended to be per client on a particular strategy pipeline.

martintmk:
> The changes look ok, but from the tests I don't understand how the rate limiter would partition on the HTTP context, for example, where the limit is intended to be per client on a particular strategy pipeline.

Let me create an more complex issue test that demonstrates this.

martintmk:
> > The changes look ok, but from the tests I don't understand how the rate limiter would partition on the HTTP context, for example, where the limit is intended to be per client on a particular strategy pipeline.

Let me create an more complex issue test that demonstrates this in follow-up.

PR #1379, opened 2023-07-03
Debugger proxies for resilience strategies

martintmk:
### Details on the issue fix or feature implementation

Simplifies the debugging. Contributes to #1367 

### Confirm the following

- [x]  I started this PR by branching from the head of the default branch
- [x]  I have targeted the PR to merge into the default branch
- [x]  I have included unit tests for the issue/feature
- [x]  I have successfully run a local build

PR #1384, opened 2023-07-04
Fix debugger proxies

martintmk:

Follow-up of #1379

### Confirm the following


martincostello (approved):
Any screenshots of the new behaviour with and without it expanded you can share?

martintmk:
> Any screenshots of the new behaviour with and without it expanded you can share?

I cannot print-screen either because VS keeps hiding it :D

But now the proxy is only applied to `ResilienceStrategyPipeline`, the others use default behavior so that endless expansion thing is gone.

Issue #1290, opened 2023-06-14
Cleanup the `Polly` and `Polly.Specs` codebase


martintmk:
In the #1287 we enabled a full set of analyzers for these projects. Many of the rules are currently suppressed in each respective `csproj` file with the `<NoWarn>` element. 

We can enable these rules one-be-one and cleanup the codebase.

davidCett:
Hi👋I would like to help with this issue

martincostello:
@davidCett Thanks for volunteering! Feel free to open a pull request to tackle one of the disabled warnings. I suggest you tackle one or two per pull request to prevent the diffs getting too big to easily review.

martincostello:
💡 Tips for future contributors working on this issue:

- Only the `Polly` and `Polly.Specs` projects are still in scope for this issue. Other projects with analyzer suppressions are intentional (we've intentionally turned them off, rather than it being _"let's fix it later"_) and represent cases where we specifically disable a rule for a project because it is not useful or code-level suppressions are too noisy (e.g. the Polly.Benchmarks project disables rules that are not compatible with BenchmarkDotNet's requirements for benchmark classes).
- Avoid fixing more than one analyzer ID in the same pull request. Unless the changes are trivial, raise a new pull request per analyzer rule - this makes code review much easier for the maintainers as otherwise you could potentially change dozens of files.
- Consider whether a specific change is useful. If fixing a specific analyzer rule impacts hundreds of files and is a subjective opinion, it may not be worth fixing and can just be left suppressed. If you're unsure, feel free to ask in a comment below before starting the work if you feel it would generate a large change.
- Do not change the public API in the case of the Polly project. If fixing an analyser rule would require a change to the public API surface (e.g. renaming a member, re-ordering arguments etc., things that require changing a `PublicAPI.*.txt` file), then do not make the suggested change - instead, suppress the warning(s) on a case-by-case basis with a pair of `#pragma warning {disable|restore} {AnalyzerId}` directives in the source file(s).
- If changes add new code to the Polly project (e.g. adding missing validation, rather than style-only changes), then ensure that test coverage is not impacted. This will require adding new unit tests.

sukreshmanda:
For the warnings [xUnit1031](https://xunit.net/xunit.analyzers/rules/xUnit1031), it says blocking calls should not be used in a test method. The below mentioned code uses _Task.WaitAll_ which is a blocking action.

Example :

``` csharp
[...]

PR #1393, opened 2023-07-10
Improve the code readability of health info method

jokoyoski:
In the refactored code:

The calculation of the total variable has been moved within the if condition to eliminate the need for a separate variable declaration.
Instead of comparing total with 0 in the if statement, we directly compare successes + failures with 0.
The calculation of failureRate has been separated into its own variable for clarity.
The result is returned as a HealthInfo object, with the successes + failures value and the failureRate value passed as arguments.
These changes improve readability and maintain the original functionality of the code.

### DIFF
--- a/src/Polly.Core/CircuitBreaker/Health/HealthInfo.cs
+++ b/src/Polly.Core/CircuitBreaker/Health/HealthInfo.cs
@@ -4,12 +4,12 @@ internal readonly record struct HealthInfo(int Throughput, double FailureRate)
 {
     public static HealthInfo Create(int successes, int failures)
     {
-        var total = successes + failures;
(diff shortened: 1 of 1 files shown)


jokoyoski:
@dotnet-policy-service agree

martincostello:
Thanks for your suggestion, but I don't agree that this improves the readability in this case.

In the worst case this performs the same addition three times now (which is why total was used to store the computation) so degrades performance, and also changes the indenting to not match the existing level of indentation used in the code style.

martintmk:
@jokoyoski 

Thanks for your contribution, although this change might not be merged, you can check #1290 which proposes similar cleanup in the `Polly` codebase.

Issue #1367, opened 2023-06-29
Introduce `Polly.Testing` package?


martintmk:
When adopting Polly V8 I thought having package as this might be useful for authors or adopters of Polly. 

It would simplify some of the common scenarios:

- Assertion of strategies in the pipeline (is my pipeline constructed correctly?)
- Assertion of telemetry enrichers. 

Basically, we could just rename and polish https://github.com/App-vNext/Polly/tree/main/test/Polly.TestUtils and publish it as a package.

Wdyt?

martincostello:
I'll ponder this further as I try v8 out more in some internal projects.

Typically I discourage people from testing the Polly implementation in of itself (like mocking things), and more encourage people to observe the effects of its usage with the real thing, as that often finds misconfigurations that mocking doesn't. Plus a no-op pipeline is easy to configure.

I'm not against the idea as such, but it's more to ship and maintain if the amount of things actually useful to others (compared to our own test suite) is negligible.

PeterCsalaHbo:
Most of the cases (what I have seen) the policy parameters (like `maxRetryCount`, `sleepDuration`, etc.) are coming from configuration files.

It is pretty easy to use the wrong configuration value (unintentionally) in a given policy/strategy setup. ([related issue](https://github.com/App-vNext/Polly/issues/756)) 

So, providing built-in assertions against policy/strategy parameters after they have be built IMHO is really valuable.

martincostello:
> The test relies on internal naming in Polly and that's why I thought we could provide something to simplify the scenario. Basically, allow some assertions to check that the retry/timeout, etc. is present in the pipeline.

That might be helpful though. In the code I'm migrating for #1365 the tests previously asserted on the property key to imply the types configured via the convention the code used.

Maybe there's a way we could provide a way to make it easier to introspect a configured pipeline in some way as a first-class API feature (not something test-specific)? It might make debugging easier too (e.g. custom debugger proxies or similar).

martintmk:
> Maybe there's a way we could provide a way to make it easier to introspect a configured pipeline in some way as a first-class API feature (not something test-specific)? It might make debugging easier too (e.g. custom debugger proxies or similar).

I am really hesitant to do this, because the consuming code should not care what and how many inner strategies are inside a pipeline. The only use-case for this are the tests. Ideally we would have the option to do this:

``` csharp
// build pipeline
var strategy = new ResilienceStrategyBuilder()
    .AddRetry(new())
[...]

PR #1394, opened 2023-07-11
Introduce `Polly.Testing` package

martintmk:

Contributes to #1367 

With this relatively small package the consumer of Polly can assert the composition of resilience strategies with ease:

``` csharp
// arrange
var strategy = new ResilienceStrategyBuilder<string>()
    .AddFallback(new()
    {
        FallbackAction = _ => Outcome.FromResultAsTask("dummy"),
    })
    .AddRetry(new())
    .AddAdvancedCircuitBreaker(new())
    .AddTimeout(TimeSpan.FromSeconds(1))
    .AddHedging(new())
    .AddConcurrencyLimiter(10)
    .AddStrategy(new CustomStrategy())
    .ConfigureTelemetry(NullLoggerFactory.Instance)
    .Build();

// act
var descriptor = strategy.GetInnerStrategies();

// assert
descriptor.HasTelemetry.Should().BeTrue();
descriptor.Strategies.Should().HaveCount(7);
descriptor.Strategies[0].Options.Should().BeOfType<FallbackStrategyOptions<string>>();
descriptor.Strategies[1].Options.Should().BeOfType<RetryStrategyOptions<string>>();
descriptor.Strategies[2].Options.Should().BeOfType<AdvancedCircuitBreakerStrategyOptions<string>>();
descriptor.Strategies[3].Options.Should().BeOfType<TimeoutStrategyOptions>();
descriptor.Strategies[3].Options
    .Should()
    .BeOfType<TimeoutStrategyOptions>().Subject.Timeout
    .Should().Be(TimeSpan.FromSeconds(1));

descriptor.Strategies[4].Options.Should().BeOfType<HedgingStrategyOptions<string>>();
descriptor.Strategies[5].Options.Should().BeOfType<RateLimiterStrategyOptions>();
descriptor.Strategies[6].StrategyType.Should().Be(typeof(CustomStrategy));
```

cc @PeterCsalaHbo

### Confirm the following


Polly.sln
```diff
@@ -52,6 +52,10 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{A6CC41B9-E
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{B7BF406B-B06F-4025-83E6-7219C53196A6}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Polly.Testing", "src\Polly.Testing\Polly.Testing.csproj", "{9AD2D6AD-56E4-49D6-B6F1-EE975D5760B9}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Polly.Testing.Tests", "test\Polly.Testing.Tests\Polly.Testing.Tests.csproj", "{D333B5CE-982D-4C11-BDAF-4217AA02306E}"
```
martincostello, line 57: 😅

src/Polly.Testing/Polly.Testing.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net7.0;net6.0;netstandard2.0;net472;net462</TargetFrameworks>
```
martincostello, line 4: I wonder if we need quite so many TFMs for this package given it's brand new and doesn't necessarily need the same backwards compatibility "baggage".

src/Polly.Testing/ResilienceStrategyDescriptor.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Polly.Testing;
+
+/// <summary>
+/// This class provides additional information about <see cref="ResilienceStrategy"/>.
```
martincostello, line 4:
```suggestion
/// This class provides additional information about a <see cref="ResilienceStrategy"/>.
```

src/Polly.Testing/ResilienceStrategyDescriptor.cs
```diff
@@ -0,0 +1,12 @@
...
+/// </summary>
+/// <param name="Options">The options used by the resilience strategy, if any.</param>
+/// <param name="Type">The type of strategy as an enum.</param>
```
martincostello, line 7:
```suggestion
/// <param name="Type">The type of strategy.</param>
```

src/Polly.Testing/ResilienceStrategyExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+        ReloadableResilienceStrategy => ResilienceStrategyType.Reload,
+        _ when strategy.GetType().FullName == "Polly.RateLimiting.RateLimiterResilienceStrategy" => ResilienceStrategyType.RateLimiter,
+        _ when strategy.GetType().FullName == "Polly.Extensions.Telemetry.TelemetryResilienceStrategy" => ResilienceStrategyType.Telemetry,
+        _ when strategy.GetType().IsGenericType && strategy.GetType().GetGenericTypeDefinition() == typeof(RetryResilienceStrategy<>) => ResilienceStrategyType.Retry,
+        _ when strategy.GetType().IsGenericType && strategy.GetType().GetGenericTypeDefinition() == typeof(CircuitBreakerResilienceStrategy<>) => ResilienceStrategyType.CircuitBreaker,
+        _ when strategy.GetType().IsGenericType && strategy.GetType().GetGenericTypeDefinition() == typeof(HedgingResilienceStrategy<>) => ResilienceStrategyType.Hedging,
+        _ when strategy.GetType().IsGenericType && strategy.GetType().GetGenericTypeDefinition() == typeof(FallbackResilienceStrategy<>) => ResilienceStrategyType.Fallback,
+
+        _ => ResilienceStrategyType.Custom
+    };
```
martincostello, line 77: My mains concern here are that this isn't really extendable for anything custom and it it might not be obvious to future maintainers that this separate code needs updating if we added any new policies (e.g. the Simmy-related work), particularly given we're using "magic strings" here as we're not depending on the types directly.

src/Polly.Testing/ResilienceStrategyType.cs
```diff
@@ -0,0 +1,52 @@
+﻿namespace Polly.Testing;
+
+/// <summary>
+/// The type of resilience strategy.
+/// </summary>
+public enum ResilienceStrategyType
```
martincostello, line 6: As per previous comment, I think having this as an enum makes things a bit too tightly bound and hard to extend, particularly with it not actually depending on some of the strategies directly but then directly exposing concepts related to them.

src/Polly.Testing/Polly.Testing.csproj
martintmk: I think just having `netstandard2.0` can be completely fine here.

src/Polly.Testing/ResilienceStrategyType.cs
martintmk: Dropped the enum.

PeterCsalaHbo:
@martintmk Could you please modify your assertions to include a sample how to check for the `AddConcurrencyLimiter`'s parameter as well?

src/Polly.Testing/ResilienceStrategyExtensions.cs
```diff
@@ -32,16 +27,23 @@ public static IReadOnlyList<ResilienceStrategyDescriptor> GetInnerStrategies<TRe
...
         Guard.NotNull(strategy);
 
         var strategies = new List<ResilienceStrategy>();
         strategy.ExpandStrategies(strategies);
 
-        return strategies.Select(s => new ResilienceStrategyDescriptor(s.Options, GetType(s), s.GetType())).ToList();
+        var innerStrategies = strategies.Select(s => new ResilienceStrategyDescriptor(s.Options, s.GetType())).ToList();
+
+        return new InnerStrategiesDescriptor(
+            innerStrategies.Where(s => !ShouldSkip(s.StrategyType)).ToList().AsReadOnly(),
```
martincostello, line 40: Do we need to force the list to be read-only?

src/Polly.Testing/InnerStrategiesDescriptor.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Polly.Testing;
+
+/// <summary>
+/// Describes the pipeline of resilience strategy.
```
martincostello, line 4:
```suggestion
/// Describes the pipeline of a resilience strategy.
```

martincostello:
I think this is lightweight enough to be useful to people - I don't think I'd use it myself personally, but this doesn't look like it would add much maintenance burden.

One thought that occurs to me: should be call this _Testing_ or maybe it's more like _Diagnostics_ or _Introspection_? Was just thinking to myself that someone could possible have a diagnostics endpoint in their application where they might want to check the configured pipeline at runtime - in that case having to have a dependency on a package called `.Testing` would look a bit suspect.

src/Polly.Testing/ResilienceStrategyExtensions.cs
martintmk: Yes, after the extension it should not be modifiable anymore.

martintmk:
> I think this is lightweight enough to be useful to people - I don't think I'd use it myself personally, but this doesn't look like it would add much maintenance burden.
> 
> One thought that occurs to me: should be call this _Testing_ or maybe it's more like _Diagnostics_ or _Introspection_? Was just thinking to myself that someone could possible have a diagnostics endpoint in their application where they might want to check the configured pipeline at runtime - in that case having to have a dependency on a package called `.Testing` would look a bit suspect.

It also shows people that we care about the testing and have native support for it. 

> One thought that occurs to me: should be call this _Testing_ or maybe it's more like _Diagnostics_ or _Introspection_? 

I am just following the other naming convention such as `Microsoft.AspNetCore.Mvc.Testing` or `Microsoft.Extensions.TimeProvider.Testing`. It's not really something we should encourage people to use at runtime. 

> Was just thinking to myself that someone could possible have a diagnostics endpoint in their application where they might want to check the configured pipeline at runtime - in that case having to have a dependency on a package called `.Testing` would look a bit suspect.

If there is use-case for this in the future we can emit built-in telemetry without exposing any new API.

martintmk:
> @martintmk Could you please modify your assertions to include a sample how to check for the `AddConcurrencyLimiter`'s parameter as well?

Unfortunate thing is that `RateLimiting` is external API and we do not have it under our control. Assertion such as that is not possible without reflection.

src/Polly.Testing/README.md
```diff
@@ -1,3 +1,28 @@
 # About Polly.Testing
 
+This package exposes API and utilities that can be used to assert the composition of resilience strategies.
```
martincostello, line 3:
```suggestion
This package exposes APIs and utilities that can be used to assert on the composition of resilience strategies.
```

martintmk:
I'll use this package to simplify tests such as these:

https://github.com/dotnet/extensions/blob/142fe1886a94e3f251e149f58e5e918161dcd7a5/test/Libraries/Microsoft.Extensions.Http.Resilience.Tests/Hedging/StandardHedgingTests.cs#L166

martincostello (approved):

PR #1360, opened 2023-06-27
Prepare for .NET 8

martincostello:
Port various changes from #1144:

- Simplify NuGet package version management.
- Simplify `TimeProviderExtensions`.
- Enable .NET analyzers.
- Remove `MockTimeProvider`.
- Minor code formatting clean-ups.

Also updates the Cake dependencies to their latest versions.

Issue #1110, opened 2023-04-11
Investigate native AOT compatibility


martincostello:
**Is your feature request related to a specific problem? Or an existing feature? Please describe.**

ASP.NET Core 8 is making investments in supporting native AOT scenarios. In the case of APIs using Polly, we should ensure that we are also native AOT-friendly.

**Describe your proposed or preferred solution**:

Polly is fully usable with native AOT (**or** with known, documented, limitations).

Potential changes we _may_ have to make include:

- annotations for trimming
- changes to any use of reflection

**Describe any alternative options you've considered**:

Do nothing.

**Any additional info?**

[ASP.NET Core blog post](https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-8-preview-3/#libraries-and-native-aot)

geeknoid:
The .NET 8 generator currently depends on Microsoft.Extensions.Options. This is because the geneator looks at the IValidateOptions<T> type to determine the type to implement validation for.

Do you have any ideas how the generator could work without taking a dependency on Microsoft.Extensions.Options?

martintmk:
@geeknoid Just thinking out loud. We could use this syntax:

``` csharp
[OptionsValidator]
public partial class RetryStrategyOptions
{
    // Can generate inner RetryStrategyOptionsValidator here 
}

internal static partial class PollyValidators
{
   // just implements the method
   [OptionsValidator]
   public static partial IEnumerable<ValidationResult> ValidateRetryOptions(RetryStrategyOptions options);
}
```

Essentially, we just need something that we pass the options instance to and it returns us the list of `ValidationResult`. If the list is empty, then there are no errors. Ideally, we would just reference the generator package as a build-time dependency without any other changes. It could recognize our internal `OptionsValidatorAttribute` and just output the code.

That way, the Polly stays dependency free on .NET3.1+ runtimes.

PR #1412, opened 2023-07-18
Introduce ResilienceStrategyBuilder.Validator

martintmk:

This API addition allows the library to inject their own validator and get rid of the reflection that is used in the default validator. 
This allows the library that consumes the Polly to be AOT compatible.

In ideal world we could just use the generators introduced by https://github.com/dotnet/runtime/issues/85475. 
However, this would pull the heavy `Microsoft.Extensions.Options` dependency into the core package. 

As a follow-up task after #1360  is merged we can introduce `ResilienceValidator.Validate` utility API in `Polly.Extensions` that uses generated validators for validation and doesn't use reflection at all (for all built-in options). 

Contributes to #1110 

### Confirm the following

### DIFF
--- a/src/Polly.Extensions/Telemetry/TelemetryResilienceStrategyBuilderExtensions.cs
+++ b/src/Polly.Extensions/Telemetry/TelemetryResilienceStrategyBuilderExtensions.cs
@@ -48,7 +48,7 @@ public static class TelemetryResilienceStrategyBuilderExtensions
         Guard.NotNull(builder);
         Guard.NotNull(options);
 
-        builder.Validator(new(options, "The 'TelemetryOptions' are invalid."));
+        builder.Validator(new(options, $"The '{nameof(TelemetryOptions)}' are invalid."));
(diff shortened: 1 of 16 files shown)


src/Polly.Extensions/Telemetry/TelemetryResilienceStrategyBuilderExtensions.cs
```diff
@@ -48,10 +48,8 @@ public static TBuilder ConfigureTelemetry<TBuilder>(this TBuilder builder, Telem
         Guard.NotNull(builder);
         Guard.NotNull(options);
 
-        ValidationHelper.ValidateObject(options, "The resilience telemetry options are invalid.");
-
+        builder.Validator(new(options, "The 'TelemetryOptions' are invalid."));
```
martincostello, line 51:
```suggestion
        builder.Validator(new(options, $"The {nameof(TelemetryOptions)} are invalid."));
```

martincostello (approved):

PR #1422, opened 2023-07-21
Convert records to classes

martintmk:

These records expose unnecessary APIs that we do not want to rely on in the future if it's not used by anything.

Contributes to #1365 

### Confirm the following


src/Polly.Core/Fallback/OnFallbackArguments.cs
```diff
@@ -3,4 +3,6 @@ namespace Polly.Fallback;
 /// <summary>
 /// Represents arguments used in fallback handling scenarios.
 /// </summary>
-public record OnFallbackArguments();
+public sealed class OnFallbackArguments
```
martincostello, line 6: Compared to the others, why is this one a class rather than a struct?

src/Polly.Core/Timeout/OnTimeoutArguments.cs
```diff
@@ -3,7 +3,33 @@ namespace Polly.Timeout;
...
+    /// <summary>
+    /// Initializes a new instance of the <see cref="OnTimeoutArguments"/> class.
+    /// </summary>
+    /// <param name="context">The context associated with the execution of a user-provided callback.</param>
+    /// <param name="exception">The original exception that caused the timeout.</param>
+    /// <param name="timeout">The timeout value assigned.</param>
+    public OnTimeoutArguments(ResilienceContext context, Exception exception, TimeSpan timeout)
+    {
+        Context = context;
+        Exception = exception;
```
martincostello, line 17: Should we null guard any of these non-nullable types being passed to these argument types, or is it fine if someone forces them to be created with a null?

src/Polly.Core/Fallback/OnFallbackArguments.cs
martintmk: We use classes for events that do not happen on a hot path. 

As opposite to predicate arguments which are all structs.

src/Polly.Core/Timeout/OnTimeoutArguments.cs
martintmk: I would say it's not needed as creation is fully under our control.

martincostello (approved):

PR #1423, opened 2023-07-21
Do not record null dimensions

martintmk:

Just investigating whether this is viable approach. It saves some CPU cycles and bandwidth in case the consumer does not care about individual dimensions. 

### Confirm the following


martintmk:
@geeknoid Wondering what's the recommended approach? Is this change ok?

geeknoid:
Yep, seems fine to me.

PR #1445, opened 2023-07-28
Allow jitter for all backoff types

martintmk:

For context:
https://github.com/App-vNext/Polly/pull/1233/files#r1251595389

If jitter is used, then the final delay is randomized between the factor of `1` ... `1.5`. 

This applies for constant and linear delay, for exponential delay the formula is more complicated. (no changes)

Contributes to #1301 

### Confirm the following

### DIFF
--- a/src/Polly.Core/Retry/RetryStrategyOptions.TResult.cs
+++ b/src/Polly.Core/Retry/RetryStrategyOptions.TResult.cs
@@ -35,7 +35,7 @@ public class RetryStrategyOptions<TResult> : ResilienceStrategyOptions
     public RetryBackoffType BackoffType { get; set; } = RetryConstants.DefaultBackoffType;
 
     /// <summary>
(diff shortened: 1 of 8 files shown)


src/Polly.Core/Retry/RetryStrategyOptions.TResult.cs
```diff
@@ -34,6 +34,18 @@ public class RetryStrategyOptions<TResult> : ResilienceStrategyOptions
     /// </value>
     public RetryBackoffType BackoffType { get; set; } = RetryConstants.DefaultBackoffType;
 
+    /// <summary>
+    /// Gets or sets a value indicating whether the jitter should be used when calculating the backoff delay between retries.
```
martincostello, line 38:
```suggestion
    /// Gets or sets a value indicating whether jitter should be used when calculating the backoff delay between retries.
```

martincostello (approved):

PR #1448, opened 2023-07-28
Rename `ResilienceStrategyBuilder` to `CompositeStrategyBuilder`

martintmk:

Per discussion in #1233

- We don't want to use term "Pipeline" because it's not used anywhere in the public API. 
- `CompositeStrategyBuilder` is good term because it indicates the result is composite strategy and has `Strategy` in the name.

Contributes to #1301 

### Confirm the following


src/Polly.Core/Utils/CompositeResilienceStrategy.cs
```diff
@@ -7,24 +7,24 @@ namespace Polly.Utils;
 /// <summary>
 /// A pipeline of strategies.
```
martincostello, line 8: This probably needs updating too.

martincostello (approved):

src/Polly.Core/Utils/CompositeResilienceStrategy.cs
martintmk: True, changed the wording.

martincostello (approved):

PR #1474, opened 2023-08-10
Add new metering tests to cover uncovered lines

martintmk:

Noticed this in #1144. 

### Confirm the following


martincostello (approved):

martincostello:
Would this maybe explain that random 99.75% coverage failure we get sometimes? Or is this something entirely new/different?

martintmk:
> Would this maybe explain that random 99.75% coverage failure we get sometimes? Or is this something entirely new/different?

It was revealed on .NET 8 so it's possible that this affected also .NET 7 and lower. Anyway let's see, this change can only help.

PR #1486, opened 2023-08-14
Introduce TelemetryListener

martintmk:

Introduce `TelemetryListener` API that is used to listen for telemetry events. As opposed to `DiagnosticSource` that only accepts object-based arguments, this listener allows generics.

This enables us to simplify the `Polly.Core` and convert all argument classes to structs and drop all pooling of these arguments. 

For context:
https://github.com/App-vNext/Polly/pull/1233#discussion_r1290138561

### Confirm the following


src/Polly.Core/CircuitBreaker/OnCircuitClosedArguments.cs
```diff
@@ -5,6 +5,9 @@ namespace Polly.CircuitBreaker;
 /// <summary>
 /// Arguments used by <see cref="CircuitBreakerStrategyOptions{TResult}.OnClosed"/> event.
 /// </summary>
+/// <remarks>
+/// Always use constructor when creating this struct, otherwise we do not guarantee the binary compatibility.
```
martincostello, line 9: Nit on these comments:

```suggestion
/// Always use the constructor when creating this struct, otherwise we do not guarantee binary compatibility.
```

martincostello (approved):

PR #1493, opened 2023-08-16
ResilienceContextPool improvements

martintmk:

- This change allows us to make the `ResilienceContext.ContinueOnCapturedContext` property read-only.
- Introduced `ResilienceContextCreationArguments` so we can add more members to it in the future without breaking changes.

### Confirm the following


martincostello (approved):
Would be interesting to see if this has any effect on the benchmark results.

PR #1495, opened 2023-08-17
Drop unused ResiliencePipelineRegistry APIs

martintmk:

This makes the registry append-only.

One reason the removals might not be required is the built-in dynamic reload support. Instead of removing, re-creating and adding the pipeline back, it's cleaner and safer to just tell the registry to reload the pipeline.

 If there is a feature request for it, we can add these APIs easily back.

Contributes to #1233

### Confirm the following

### DIFF
--- a/src/Polly.Core/Registry/ResiliencePipelineRegistry.TResult.cs
+++ b/src/Polly.Core/Registry/ResiliencePipelineRegistry.TResult.cs
@@ -28,10 +28,6 @@ public sealed partial class ResiliencePipelineRegistry<TKey> : ResiliencePipelin
             _instanceNameFormatter = instanceNameFormatter;
         }
 
-        public bool TryAdd(TKey key, ResiliencePipeline<TResult> strategy) => _strategies.TryAdd(key, strategy);
-
-        public bool Remove(TKey key) => _strategies.TryRemove(key, out _);
-
         public bool TryGet(TKey key, [NotNullWhen(true)] out ResiliencePipeline<TResult>? strategy)
         {
             if (_strategies.TryGetValue(key, out strategy))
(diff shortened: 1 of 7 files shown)


src/Polly.Core/Registry/ResiliencePipelineRegistry.TResult.cs
```diff
@@ -65,9 +61,5 @@ public ResiliencePipeline<TResult> GetOrAdd(TKey key, Action<ResiliencePipelineB
         }
 
         public bool TryAddBuilder(TKey key, Action<ResiliencePipelineBuilder<TResult>, ConfigureBuilderContext<TKey>> configure) => _builders.TryAdd(key, configure);
-
-        public bool RemoveBuilder(TKey key) => _builders.TryRemove(key, out _);
-
-        public void Clear() => _strategies.Clear();
```
martincostello, line 71: I think we should maybe keep clear at least. Relying on reloading to do that for you feels a bit unintuitive to me.

martintmk: The follow-up of this PR is the dispose support for the registry shich is something that's missing right now. It essentially frees resources held by individual resilience pipelines (`RateLimiter`, `CircuitBreaker`).

Now the problem is:

``` csharp
var registry = new ResiliencePipelineRegistry<string>();
var pipeline= registry.GetPipeline("my-pipeline");

// this call also disposes all resources held by "my-pipeline"
registry.ClearPipelines();

// ObjectDisposedException
pipeline.Execute(()=> {  });
```

For this reason I thought it might be better to just wait and see. If someone really needs it we can add it, but the consumer should be aware of side-effects.

martincostello (approved):

src/Polly.Core/Registry/ResiliencePipelineRegistry.TResult.cs
martintmk: The follow up for this PR is the disposal of resources that are held by the resilience pipelines.

Think R

PR #1506, opened 2023-08-22
API Review Feedback

martintmk:

- Drop `ResilienceContext.ResilienceEvents`
- Rename `ResiliencePipeline.Null` to `ResiliencePipeline.Empty`
- Hide some `Outcome<>` properties
- Drop `RateLimiterRejected.RetryAfter`

Contributes to #1233 

### Confirm the following


martincostello (approved):

martincostello:
At this point I'm finding #1233 too unwieldly to follow so I've stopped keeping up to date with it.

How many more changes are we expecting before we actually freeze the public API unless there's specific user feedback in an issue?

martincostello:
> Just thinking, would opening a new PR with something like `Polly V8: Public API Review (final polish)` actually help?

Possibly.

The main problem I have is that the diff is _massive_ so looking for inline comments when I've marked a file as reviewed is difficult, and there's so many comments on the PR that most of them are hidden if they're old:

image

martintmk:
> > Just thinking, would opening a new PR with something like `Polly V8: Public API Review (final polish)` actually help?
> 
> Possibly.
> 
> The main problem I have is that the diff is _massive_ so looking for inline comments when I've marked a file as reviewed is difficult, and there's so many comments on the PR that most of them are hidden if they're old:
> 
> image

Let me give a shot at this tomorrow. There are maybe 10 comments left to address. Some of these came just recently, however it might be good to have something where one is not immediately lost :)

PR #1511, opened 2023-08-24
Allow to dispose linked resources on pipeline disposal

martintmk:

Allows disposal of resources attached to the pipeline or to attach any cleanup actions:

``` csharp
var provider = new ServiceCollection()
    .AddResiliencePipeline("my-pipeline", (builder, context) =>
    {
        var limiter = new ConcurrencyLimiter(new ConcurrencyLimiterOptions
        {
            PermitLimit = 1,
            QueueLimit = 1
        });

        builder.AddRateLimiter(new RateLimiterStrategyOptions
        {
            RateLimiter = args => limiter.AcquireAsync(1, args.Context.CancellationToken)
        });

        // when the pipeline instance is disposed, limiter is disposed too
        context.OnPipelineDisposed(() => limiter.Dispose());
    });
```

This is not inclusive to rate limiters as any disposable resource created when configuring the pipeline can be now freed after the pipeline is disposed. This prevents memory leaks and improves the performance.

Contributes to https://github.com/App-vNext/Polly/pull/1507#discussion_r1302499181

### Confirm the following

### DIFF
--- a/src/Polly.Core/Utils/Pipeline/ComponentWithDisposeCallbacks.cs
+++ b/src/Polly.Core/Utils/Pipeline/ComponentWithDisposeCallbacks.cs
@@ -7,7 +7,7 @@ internal class ComponentWithDisposeCallbacks : PipelineComponent
     public ComponentWithDisposeCallbacks(PipelineComponent component, List<Action> callbacks)
     {
         Component = component;
-        _callbacks = callbacks.ToList();
+        _callbacks = callbacks;
     }
 
     internal PipelineComponent Component { get; }
--- a/src/Polly.Core/Utils/Pipeline/ComponentWithDisposeCallbacks.cs
+++ b/src/Polly.Core/Utils/Pipeline/ComponentWithDisposeCallbacks.cs
@@ -1,0 +1,43 @@
+﻿namespace Polly.Utils.Pipeline;
+
+internal class ComponentWithDisposeCallbacks : PipelineComponent
+{
+    private readonly List<Action> _callbacks;
+
+    public ComponentWithDisposeCallbacks(PipelineComponent component, List<Action> callbacks)
+    {
+        Component = component;
+        _callbacks = callbacks.ToList();
+    }
+
+    internal PipelineComponent Component { get; }
+
+    public override void Dispose()
+    {
+        ExecuteCallbacks();
+
+        Component.Dispose();
+    }
+
+    public override ValueTask DisposeAsync()
+    {
+        ExecuteCallbacks();
+
+        return Component.DisposeAsync();
+    }
+
+    internal override ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState>(
+        Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
+        ResilienceContext context,
+        TState state) => Component.ExecuteCore(callback, context, state);
+
+    private void ExecuteCallbacks()
+    {
+        foreach (var callback in _callbacks)
+        {
+            callback();
+        }
+
+        _callbacks.Clear();
+    }
+}
--- a/src/Polly.Core/Utils/Pipeline/PipelineComponentFactory.cs
+++ b/src/Polly.Core/Utils/Pipeline/PipelineComponentFactory.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using Polly.Telemetry;
+
 namespace Polly.Utils.Pipeline;
 
 internal static class PipelineComponentFactory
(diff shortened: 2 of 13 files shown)


src/Polly.Core/Utils/Pipeline/ComponentWithDisposeCallbacks.cs
```diff
@@ -0,0 +1,43 @@
+﻿namespace Polly.Utils.Pipeline;
+
+internal class ComponentWithDisposeCallbacks : PipelineComponent
+{
+    private readonly List<Action> _callbacks;
+
+    public ComponentWithDisposeCallbacks(PipelineComponent component, List<Action> callbacks)
+    {
+        Component = component;
+        _callbacks = callbacks.ToList();
```
martincostello, line 10: `ToList()` on a `List<T>` looks a bit weird to me 😄

```suggestion
        _callbacks = new(callbacks);
```

src/Polly.Core/Utils/Pipeline/ComponentWithDisposeCallbacks.cs
```diff
@@ -0,0 +1,43 @@
...
+        TState state) => Component.ExecuteCore(callback, context, state);
+
+    private void ExecuteCallbacks()
+    {
+        foreach (var callback in _callbacks)
+        {
+            callback();
+        }
+
+        _callbacks.Clear();
```
martincostello, line 41: Do we need to worry about a race here if it gets disposed while still disposing?

test/Polly.Core.Tests/Utils/Pipeline/PipelineComponentFactoryTests.cs
```diff
@@ -0,0 +1,26 @@
...
+        var component = Substitute.For<PipelineComponent>();
+        var result = PipelineComponentFactory.WithDisposableCallbacks(component, new List<Action>());
+        result.Should().BeSameAs(component);
+    }
+
+    [Fact]
+    public void PipelineComponentFactory_Should_Return_WrapperComponent_With_Callbacks()
+    {
+        var component = Substitute.For<PipelineComponent>();
+        List<Action> callbacks = new List<Action> { () => { } };
```
martincostello, line 20: Can we either use `var` or `new()` here?

src/Polly.Core/Utils/Pipeline/ComponentWithDisposeCallbacks.cs
martintmk: Not really, the pipeline is disposed only once and from a single thread.

src/Polly.Core/Utils/Pipeline/ComponentWithDisposeCallbacks.cs
martintmk: changed the code slightly so the copy happens outside of the class

src/Polly.Core/Utils/Pipeline/ComponentWithDisposeCallbacks.cs
```diff
@@ -7,7 +7,7 @@ internal class ComponentWithDisposeCallbacks : PipelineComponent
     public ComponentWithDisposeCallbacks(PipelineComponent component, List<Action> callbacks)
     {
         Component = component;
-        _callbacks = callbacks.ToList();
+        _callbacks = callbacks;
```
martincostello, line 10: Did you not need it to be a copy?

src/Polly.Core/Utils/Pipeline/PipelineComponentFactory.cs
```diff
@@ -14,14 +14,14 @@ internal static class PipelineComponentFactory
 
     public static PipelineComponent FromStrategy<T>(ResilienceStrategy<T> strategy) => new BridgeComponent<T>(strategy);
 
-    public static PipelineComponent WithDisposableCallbacks(PipelineComponent component, List<Action> callbacks)
+    public static PipelineComponent WithDisposableCallbacks(PipelineComponent component, IEnumerable<Action> callbacks)
     {
-        if (callbacks.Count == 0)
+        if (!callbacks.Any())
```
martincostello, line 19: Just curious on the rationale for this change (as it's internal), as the new .NET 8 analyzers for "prefer concrete types" and "don't use Any() when you have a count/length" would apply here if it actually was a list at runtime and be more performant than treating it as an enumerable.

src/Polly.Core/Utils/Pipeline/ComponentWithDisposeCallbacks.cs
martincostello: Just saw your comment on the original review comment.

src/Polly.Core/Utils/Pipeline/PipelineComponentFactory.cs
martintmk: I wanted to do the safe copy of the list without going over that `list.ToList()` thingy :) 

Here it looks more natural.

martincostello (approved):

PR #1522, opened 2023-08-29
Protect against retry delay overflows

martintmk:

If the calculated delay overflows just return `TimeSpan.MaxValue`.

### Confirm the following

### DIFF
--- a/src/Polly.Core/Retry/RetryHelper.cs
+++ b/src/Polly.Core/Retry/RetryHelper.cs
@@ -10,6 +10,18 @@ internal static class RetryHelper
 
     public static TimeSpan GetRetryDelay(RetryBackoffType type, bool jitter, int attempt, TimeSpan baseDelay, ref double state, Func<double> randomizer)
     {
+        try
+        {
+            return GetRetryDelayCore(type, jitter, attempt, baseDelay, ref state, randomizer);
+        }
+        catch (OverflowException)
+        {
+            return TimeSpan.MaxValue;
+        }
+    }
+
+    private static TimeSpan GetRetryDelayCore(RetryBackoffType type, bool jitter, int attempt, TimeSpan baseDelay, ref double state, Func<double> randomizer)
+    {
         if (baseDelay == TimeSpan.Zero)
         {
             return baseDelay;
--- a/test/Polly.Core.Tests/Retry/RetryHelperTests.cs
+++ b/test/Polly.Core.Tests/Retry/RetryHelperTests.cs
@@ -95,6 +95,14 @@ public class RetryHelperTests
         RetryHelper.GetRetryDelay(RetryBackoffType.Exponential, false, 2, TimeSpan.FromSeconds(1), ref state, _randomizer).Should().Be(TimeSpan.FromSeconds(4));
     }
 
+    [Fact]
+    public void GetRetryDelay_Overflow_ReturnsMaxTimeSpan()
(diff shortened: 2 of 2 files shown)


src/Polly.Core/Retry/RetryHelper.cs
```diff
@@ -9,6 +9,18 @@ internal static class RetryHelper
...
     public static TimeSpan GetRetryDelay(RetryBackoffType type, bool jitter, int attempt, TimeSpan baseDelay, ref double state, Func<double> randomizer)
+    {
+        try
+        {
+            return GetRetryDelayCore(type, jitter, attempt, baseDelay, ref state, randomizer);
+        }
+        catch (OverflowException)
+        {
+            return TimeSpan.MaxValue;
+        }
```
martincostello, line 20: Just wondering what's the use case for trapping the overflow, rather than throwing an exception and the user fixing their code/configuration?

martincostello (approved):

src/Polly.Core/Retry/RetryHelper.cs
martintmk: Good point, maybe this really should be handled outside of Polly?

For example similar to how done here:

https://github.com/dotnet/extensions/blob/f1175430d64364b2a90e0188ba5caa8042c339f3/src/Libraries/Microsoft.Extensions.Http.Resilience/Internal/ValidationHelper.cs#L49


One use-case is really high number of retries, at some point the retry strategy will will fail with overflow. 
Is it an user-error or Polly error?

peter-csala: A kinda weird use case just bumped into my mind. (It is slightly related to this overflow problem) On StackOverflow once I have seen a question where the OP wanted to use the retry policy to schedule jobs to run them periodically (say once in a day).

I highly discourage the OP to use Polly V7 for that. But what's the case for V8? Is the new implementation suitable for this use case? Or should be people directed to dedicated job scheduler solution?

martintmk: It does not seems to be the use-case of Polly. Yes, you can misuse retries to do "periodical" executions, but eyebrows would  be raised :)

peter-csala: What do you think would it make sense to an "anti-patterns" section on the wiki/readme?

martintmk: If there are enough of those, I would say yes.

martincostello: I'd agree - using Polly to do that sounds like a case of _"when you have a hammer, everything looks like a nail"_.

We could maybe document anti-patterns, but I would only do so for things that are common mistakes or there's subtlety as to why they are - otherwise the number of things you shouldn't use Polly for is basically infinite.

peter-csala: I will collect anti-patterns that I have seen on SO in the past 3 years and I'll create an issue to discuss them and come up with suggested alternatives for V8

martintmk:
@martincostello Is it ok to merge this one or should I abandon it?

My thinking is that even when user sets some crazy configuration, Polly should be able to handle it (by limiting the maximum delay to max, for example).

https://github.com/App-vNext/Polly/pull/1522#discussion_r1308351167

PR #1459, opened 2023-08-07
Simmy v8

vany0114:
### Migrates Simmy to use the new Polly v8 
This PR migrates all the Simmy functionality into the new Polly v8 version so that Simmy can also take advantage of all the improvements made in the v8 version. 

### Details on the issue fix or feature implementation
* Simmy now lives in the same Polly package so that resilience and chaos features are shipped together. 
    * Pros: 
        * Easier releases 
        * Easier to maintain
        * All the Polly internal stuff is available for Simmy too
        * Any bug fix or improvement is immediately available in Simmy, which allows us to keep Simmy up to date in regards to Polly.
        * Consumers don't need to worry about adding another dependency when it comes to handling chaos.
    * Cons: 
        * We cannot track Simmy usage statistics, like downloads, repos where it is being used, etc, since now it will be bundled in the same package as Polly.
* I dropped the `Inject` prefix from the name, e.g: `InjectBehaviorPolicy` -> `BehaviorChaosStrategy`
* I used the `ChaosStrategy` suffix rather than `ResilienceStrategy` to make it clearer
* The new strategies are:
    * `BehaviorChaosStrategy`
    * `LatencyChaosStrategy`
    * `OutcomeChaosStrategy`: this one provides two functionalities, inject faults as well as custom results.
* All of the chaos strategies inherit from the `MonkeyStrategy` implementation, which is also exposed so that people can extend it to create their own chaos strategies.

### TODO:
* Docs
* Examples

### Questions:
* Do you guys think would be better to create an `AddChaosStrategy` service collection extension to separate it from `AddResilienceStrategy` so that is clearer for the consumers and also to allow keeping them separate in different registries, etc? If so I guess this would require some extra effort to potentially use a `CompositeChaosStrategyBuilder` and/or generic constraints.

### Confirm the following

- [x]  I started this PR by branching from the head of the default branch
- [x]  I have targeted the PR to merge into the default branch
- [x]  I have included unit tests for the issue/feature
- [x]  I have successfully run a local build

PR #1571, opened 2023-09-08
API Review for Simmy in v8

martintmk:

Based on #1459 

### Confirm the following

### DIFF
--- a/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy.Outcomes/OutcomeStrategyOptions.TResult.cs
+++ b/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy.Outcomes/OutcomeStrategyOptions.TResult.cs
@@ -1,0 +1,15 @@
+// Assembly 'Polly.Core'
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace Polly.Simmy.Outcomes;
+
+public class OutcomeStrategyOptions<TResult> : MonkeyStrategyOptions
+{
+    public Func<OnOutcomeInjectedArguments<TResult>, ValueTask>? OnOutcomeInjected { get; set; }
+    public Func<OutcomeGeneratorArguments, ValueTask<Outcome<TResult>?>>? OutcomeGenerator { get; set; }
+    public Outcome<TResult>? Outcome { get; set; }
+    public OutcomeStrategyOptions();
+}
--- a/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy.Outcomes/OutcomeStrategyOptions.TResult.cs
+++ b/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy.Outcomes/OutcomeStrategyOptions.TResult.cs
@@ -1,6 +1,7 @@
 // Assembly 'Polly.Core'
 
 using System;
+using System.ComponentModel.DataAnnotations;
 using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 
@@ -9,7 +10,7 @@ namespace Polly.Simmy.Outcomes;
 public class OutcomeStrategyOptions<TResult> : MonkeyStrategyOptions
 {
     public Func<OnOutcomeInjectedArguments<TResult>, ValueTask>? OnOutcomeInjected { get; set; }
-    public Func<OutcomeGeneratorArguments, ValueTask<Outcome<TResult>?>>? OutcomeGenerator { get; set; }
-    public Outcome<TResult>? Outcome { get; set; }
+    [Required]
+    public Func<OutcomeGeneratorArguments, ValueTask<Outcome<TResult>?>> OutcomeGenerator { get; set; }
     public OutcomeStrategyOptions();
 }
--- a/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/BehaviorChaosPipelineBuilderExtensions.cs
+++ b/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/BehaviorChaosPipelineBuilderExtensions.cs
@@ -1,0 +1,14 @@
+// Assembly 'Polly.Core'
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Threading.Tasks;
+using Polly.Simmy.Behavior;
+
+namespace Polly.Simmy;
+
+public static class BehaviorChaosPipelineBuilderExtensions
+{
+    public static TBuilder AddChaosBehavior<TBuilder>(this TBuilder builder, bool enabled, double injectionRate, Func<ValueTask> behavior) where TBuilder : ResiliencePipelineBuilderBase;
+    public static TBuilder AddChaosBehavior<TBuilder>(this TBuilder builder, BehaviorStrategyOptions options) where TBuilder : ResiliencePipelineBuilderBase;
+}
--- a/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/EnabledGeneratorArguments.cs
+++ b/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/EnabledGeneratorArguments.cs
@@ -1,0 +1,11 @@
+// Assembly 'Polly.Core'
+
+using System.Runtime.CompilerServices;
+
+namespace Polly.Simmy;
+
+public readonly struct EnabledGeneratorArguments
+{
+    public ResilienceContext Context { get; }
+    public EnabledGeneratorArguments(ResilienceContext context);
+}
--- a/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/LatencyChaosPipelineBuilderExtensions.cs
+++ b/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/LatencyChaosPipelineBuilderExtensions.cs
@@ -1,0 +1,13 @@
+// Assembly 'Polly.Core'
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using Polly.Simmy.Latency;
+
+namespace Polly.Simmy;
+
+public static class LatencyChaosPipelineBuilderExtensions
+{
+    public static TBuilder AddChaosLatency<TBuilder>(this TBuilder builder, bool enabled, double injectionRate, TimeSpan latency) where TBuilder : ResiliencePipelineBuilderBase;
+    public static TBuilder AddChaosLatency<TBuilder>(this TBuilder builder, LatencyStrategyOptions options) where TBuilder : ResiliencePipelineBuilderBase;
+}
--- a/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/MonkeyStrategy.cs
+++ b/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/MonkeyStrategy.cs
@@ -1,0 +1,14 @@
+// Assembly 'Polly.Core'
+
+using System;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace Polly.Simmy;
+
+public abstract class MonkeyStrategy : ResilienceStrategy
+{
+    protected MonkeyStrategy(MonkeyStrategyOptions options);
+    protected ValueTask<bool> ShouldInjectAsync(ResilienceContext context);
+}
--- a/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/OutcomeChaosPipelineBuilderExtensions.cs
+++ b/ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/OutcomeChaosPipelineBuilderExtensions.cs
@@ -1,0 +1,20 @@
+// Assembly 'Polly.Core'
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using Polly.Simmy.Outcomes;
+
+namespace Polly.Simmy;
+
+public static class OutcomeChaosPipelineBuilderExtensions
+{
+    public static ResiliencePipelineBuilder AddChaosFault(this ResiliencePipelineBuilder builder, bool enabled, double injectionRate, Exception fault);
+    public static ResiliencePipelineBuilder AddChaosFault(this ResiliencePipelineBuilder builder, bool enabled, double injectionRate, Func<Exception?> faultGenerator);
+    public static ResiliencePipelineBuilder AddChaosFault(this ResiliencePipelineBuilder builder, OutcomeStrategyOptions<Exception> options);
+    public static ResiliencePipelineBuilder<TResult> AddChaosFault<TResult>(this ResiliencePipelineBuilder<TResult> builder, bool enabled, double injectionRate, Exception fault);
+    public static ResiliencePipelineBuilder<TResult> AddChaosFault<TResult>(this ResiliencePipelineBuilder<TResult> builder, bool enabled, double injectionRate, Func<Exception?> faultGenerator);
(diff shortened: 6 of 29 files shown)


martintmk:
cc @martincostello, @joelhulen, @vany0114

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/OutcomeChaosPipelineBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Diagnostics.CodeAnalysis;
+using Polly.Simmy.Outcomes;
+
+namespace Polly.Simmy;
+
+public static class OutcomeChaosPipelineBuilderExtensions
+{
+    public static ResiliencePipelineBuilder AddChaosFault(this ResiliencePipelineBuilder builder, bool enabled, double injectionRate, Exception fault);
+    public static ResiliencePipelineBuilder AddChaosFault(this ResiliencePipelineBuilder builder, bool enabled, double injectionRate, Func<Exception?> faultGenerator);
+    public static ResiliencePipelineBuilder AddChaosFault(this ResiliencePipelineBuilder builder, OutcomeStrategyOptions<Exception> options);
```
martintmk, line 13:
```suggestion
    public static TBuilder AddChaosFault<TBuilder>(this ResiliencePipelineBuilder builder, FaultStrategyOptions options) where TBuilder: ResiliencePipelieBuilderBase;
```
Using `OutcomeStrategyOptions<Exception>` feels odd as per this comment:
https://github.com/App-vNext/Polly/pull/1459#discussion_r1305198892

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy.Outcomes/OutcomeStrategyOptions.TResult.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace Polly.Simmy.Outcomes;
+
+public class OutcomeStrategyOptions<TResult> : MonkeyStrategyOptions
+{
+    public Func<OnOutcomeInjectedArguments<TResult>, ValueTask>? OnOutcomeInjected { get; set; }
+    public Func<OutcomeGeneratorArguments, ValueTask<Outcome<TResult>?>>? OutcomeGenerator { get; set; }
+    public Outcome<TResult>? Outcome { get; set; }
```
martintmk, line 13: Hide this property? 

This will cause problems with disposable outcomes such as `HttpResponseMessage`. Outcome strategy will return pre-defined outcome and caller will dispose it this making the `Outcome` invalid for all other calls. 

For simple scenarios people can use this extension:

``` 
public static ResiliencePipelineBuilder<TResult> AddChaosResult<TResult>(this ResiliencePipelineBuilder<TResult> builder, bool enabled, double injectionRate, TResult result)
```

And if using options:

``` csharp
new ResiliencePipelineBuilder<HttpResponseMessage>()
    .AddChaosResult(new()
    {
        OutcomeGenerator = _ =>
        {
            var outcome = Outcome.FromResult(new HttpResponseMessage(HttpStatusCode.InternalServerError));

            return new ValueTask<Outcome<HttpResponseMessage>?>(outcome);
        }
    })
    .Build();
```

@clpun, I know we had this issue in our code when using older Simmy and I am thinking if we can prevent this with new design.

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/MonkeyStrategy.cs
```diff
@@ -0,0 +1,14 @@
+// Assembly 'Polly.Core'
+
+using System;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace Polly.Simmy;
+
+public abstract class MonkeyStrategy : ResilienceStrategy
```
martintmk, line 10: > > Top Level Comment:
> > There are 3 terms in this PR that basically represent the same thing:
> > 
> > * Monkey
> > * Simmy
> > * Chaos
> > 
> > Do you think we can consolidate and align this somehow?
> 
> @martintmk I think all of them are still valid tho, I mean from a brand standpoint if you will, back when Simmy was conceived, kinda the reasoning or the slogan was "Simmy is a monkey to create chaos" and it was developed mapping that, meaning that the Monkey is the one in charge to unleash the chaos as the more basic chaos unit let's say, thus the public base class that consumer can extend to implement their own monkies, for instance, I might want a monkey called Caesar that implements a strategy to stop nodes in a K8s cluster. Now, Simmy is our monkey implementation that offers a set of common chaos strategies.
> 
> @joelhulen thoughts?

Moved the this from the #1459 PR. 

Alternative naming:

- `Polly.Simmy` -> `Polly.Chaos` (namespace)
- `MonkeyStrategy` -> `ChaosStrategy`
- `MonkeyStrategy<T>` -> `ChaosStrategy<T>`

Personally, I feel this is a cleaner naming, but Simmy would lose the branding and effective dissolve into Polly. 

@vany0114, @martincostello, @joelhulen

martincostello: I have no strong thoughts on branded vs. not, but I agree that the naming should be consistent whichever is chosen.

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/BehaviorChaosPipelineBuilderExtensions.cs
```diff
@@ -0,0 +1,14 @@
+// Assembly 'Polly.Core'
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Threading.Tasks;
+using Polly.Simmy.Behavior;
+
+namespace Polly.Simmy;
+
+public static class BehaviorChaosPipelineBuilderExtensions
```
martincostello, line 10: This class name is a bit of a mouthful 😄

If we don't make it simpler, maybe `ChaosBehaviorPipelineBuilderExtensions` would be better?

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/EnabledGeneratorArguments.cs
```diff
@@ -0,0 +1,11 @@
+// Assembly 'Polly.Core'
+
+using System.Runtime.CompilerServices;
+
+namespace Polly.Simmy;
+
+public readonly struct EnabledGeneratorArguments
```
martincostello, line 7: Should this contain a bool so you know if it was turned on or turned off (enabled/disabled)?

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/LatencyChaosPipelineBuilderExtensions.cs
```diff
@@ -0,0 +1,13 @@
+// Assembly 'Polly.Core'
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using Polly.Simmy.Latency;
+
+namespace Polly.Simmy;
+
+public static class LatencyChaosPipelineBuilderExtensions
```
martincostello, line 9: Similar thoughts on the class name here.

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/OutcomeChaosPipelineBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+using System;
+using System.Diagnostics.CodeAnalysis;
+using Polly.Simmy.Outcomes;
+
+namespace Polly.Simmy;
+
+public static class OutcomeChaosPipelineBuilderExtensions
+{
+    public static ResiliencePipelineBuilder AddChaosFault(this ResiliencePipelineBuilder builder, bool enabled, double injectionRate, Exception fault);
+    public static ResiliencePipelineBuilder AddChaosFault(this ResiliencePipelineBuilder builder, bool enabled, double injectionRate, Func<Exception?> faultGenerator);
```
martincostello, line 12: Do we need both of these given someone could use the second one and do `static (_) => new BlahException("whatever"))`?

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/BehaviorChaosPipelineBuilderExtensions.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    public static TBuilder AddChaosBehavior<TBuilder>(this TBuilder builder, bool enabled, double injectionRate, Func<ValueTask> behavior) where TBuilder : ResiliencePipelineBuilderBase;
```
martintmk, line 12:
```suggestion
    public static TBuilder AddChaosBehavior<TBuilder>(this TBuilder builder, double injectionRate, Func<ValueTask> behavior) where TBuilder : ResiliencePipelineBuilderBase;
```
Do we really need `enabled` parameter, why would I want to add disabled behavior if there is a no way to enable it later?

Applies to other convenience extensions too.

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/OutcomeChaosPipelineBuilderExtensions.cs
vany0114: I think you created this PR with an old API file bc I already changed that to use `FaultStrategyOptions`.

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/OutcomeChaosPipelineBuilderExtensions.cs
vany0114: It's just for convenience, having those two options for the consumers doesn't hurt I guess.

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/BehaviorChaosPipelineBuilderExtensions.cs
vany0114: We do, it is very important when it comes to enabling/disabling the chaos dynamically, Imagine you have a UI where you can configure the chaos settings, like a master knob to turn it on/off among other settings, I wrote a couple of post about it back then: 
* https://elvanydev.com/chaos-injection-with-simmy/#the-chaos-ui
* https://elvanydev.com/simmy-with-azure-app-configuration/

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/EnabledGeneratorArguments.cs
vany0114: That's actually the result of the `enabledGenerator`

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/BehaviorChaosPipelineBuilderExtensions.cs
vany0114: Actually, I think I'm gonna drop the `Chaos` prefix from the file name, the API is already clear as it is using the `Chaos` prefix: AddChaosBehavior, AddChaosLatency, AddChaosFault etc. 

Otherwise we would need to apply that change everywhere for consistency's sake, like:

* ChaosBehaviorPipelineBuilderExtensions 
* ChaosBehaviorActionArguments
* ChaosBehaviorConstants
* ChaosBehaviorStrategyOptions
* OnChaosBehaviorInjectedArguments

Same for other strategies, etc.

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/BehaviorChaosPipelineBuilderExtensions.cs
martintmk: I fully agree with the above, however consider that is more advanced scenario that is fully supported when using options. 

The convenience extensions are just shortcuts and should be (imho) as simple as possible. That's why I suggested dropping the enabled flag as it doesn't add much value, just an additional parameter that is mostly not necessary for the simple scenarios. They can "disable" the behavior by simply setting `injectionRate` to 0 when using these convenience extensions.

vany0114: Oh I see, my bad 🤦‍♂️ I thought you were talking about removing the `enabled` property entirely but you're talking about only the convenience extensions, sorry about it, so I think that makes sense, I'll address your suggestion.

vany0114:
@martintmk I think this PR is not a draft anymore right? I guess that's why I cannot mark comments as resolved 🤔

martintmk:
> @martintmk I think this PR is not a draft anymore right? I guess that's why I cannot mark comments as resolved 🤔 

Strange, maybe I should mark it as ready for review?

vany0114:
> > @martintmk I think this PR is not a draft anymore right? I guess that's why I cannot mark comments as resolved 🤔
> 
> Strange, maybe I should mark it as ready for review?

Yep I think so.

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/OutcomeChaosPipelineBuilderExtensions.cs
martincostello: Polly v7 had issues evolving due to the explosion of overloads we had that made adding a new one difficult without breaking stuff - limiting trivial convenience overloads might help us avoid that in the future.

It's still relatively simple here so not a hill I'll die on, but worth considering.

vany0114: I removed the `enabled` parameter FYI so the convenience extensions are now simpler.

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy.Outcomes/OutcomeStrategyOptions.TResult.cs
vany0114: This is done, don't know why I can't mark comments as resolved 🤔

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/BehaviorChaosPipelineBuilderExtensions.cs
vany0114: done

ApiReview/API.Polly.Core/NoDocs/Polly.Simmy/MonkeyStrategy.cs
joelhulen: My two cents is that Simmy has been around for a few years now, and there's fairly steady traffic to the repo (11 clones and 147 unique visitors in the past two weeks alone, not to mention almost 500 stars). I know that @vany0114 has done a bit of work raising awareness of the project. I also like the monkey logo :) All said, I think there is a bit of brand awareness that I don't think we necessarily need to throw away. However, I like the clarity of the proposed naming. But because of the previous points, I'm OK with sticking to the branded naming as long as, like @martincostello said, we stay consistent in the naming. We also should be very descriptive in the XML document and related published documentation to prevent any confusion.

vany0114: > Can you create a tasks to expose Simmy in 8.2.0?

@martintmk by "creating the tasks" you mean the PR(s) right?

Issue #1091, opened 2023-03-31
v8 API documentation


martincostello:
As part of the release of Polly v8 we should ensure that either before or soon after release we have reviewed all of the existing documentation to ensure that we have:

- Updated the README and wiki to cover the v8 API
- Add any relevant new documentation that isn't covered by the above (e.g. brand new features)
- Have a migration guide for moving from v7 to v8
- Have a guide on how to wrap the v7 API with with the v8 API (i.e. a partial migration)

martintmk:
Let's mention the order of executions when adding strategies which is changed compared to Wrap in V7:

https://github.com/App-vNext/Polly/issues/1365#issuecomment-1618794684

martintmk:
We should also prepare migration guide from V7 to V8.

martintmk:
Let's also make our code-snipets future-proof by injecting them into markdowns:

https://github.com/App-vNext/Polly/pull/1539#issuecomment-1702511134

cc @martincostello

PR #1598, opened 2023-09-19
Migration guide from v7 to v8

martintmk:

Migration guide from v7 to v8

Contributes to #1091 

### Confirm the following

### DIFF
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -269,6 +269,9 @@ new ResiliencePipelineBuilder().AddRetry(new RetryStrategyOptions
 ```
 <!-- endSnippet -->
 
+- To disable waiting between retries, set the `Delay` property to `TimeSpan.Zero`.
+- To retry forever, set the `MaxRetryAttempts` property to `int.MaxValue`.
+
 ### Retry and wait in v7
 
 <!-- snippet: migration-retry-wait-v7 -->
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -144,7 +144,7 @@ cancellationToken);
 Here are the primary changes:
 
 - `ResiliencePipelineBuilder` is used to start building the resilience pipeline, instead of the static `Policy.HandleException<TException>(...)`.
-- For building generic resilience pipelines, `ResiliencePipelineBuilder<T>` replaces the static `Policy.handleResult<HttpResponseMessage>(...)`.
+- For building generic resilience pipelines, `ResiliencePipelineBuilder<T>` replaces the static `Policy.HandleResult<HttpResponseMessage>(...)`.
 - Resilience strategies are configured using options, such as `RetryStrategyOptions` and `RetryStrategyOptions<T>`.
 - `PredicateBuilder` determines which exceptions or results are handled.
 - The final `ResiliencePipeline` is created with a `Build()` call.
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -346,9 +346,71 @@ new ResiliencePipelineBuilder().AddRetry(new RetryStrategyOptions
 
 It's important to remember that the configuration in v8 is options based, i.e. `RetryStrategyOptions` are used.
 
+## Migrating rate limit policy
+
+The rate limit policy is now replaced by the [rate limiter strategy](strategies/rate-limiter.md) which uses the [`System.Threading.RateLimiting`](https://www.nuget.org/packages/System.Threading.RateLimiting) package. Polly does not implement its own rate limiter anymore.
+
+### Rate limit in v7
+
+<!-- snippet: migration-rate-limit-v7 -->
+```cs
+// Create sync rate limiter
+ISyncPolicy syncPolicy = Policy.RateLimit(
+    numberOfExecutions: 100,
+    perTimeSpan: TimeSpan.FromMinutes(1));
+
+// Create async rate limiter
+IAsyncPolicy asyncPolicy = Policy.RateLimitAsync(
+    numberOfExecutions: 100,
+    perTimeSpan: TimeSpan.FromMinutes(1));
+
+// Create generic sync rate limiter
+ISyncPolicy<HttpResponseMessage> syncPolicyT = Policy.RateLimit<HttpResponseMessage>(
+    numberOfExecutions: 100,
+    perTimeSpan: TimeSpan.FromMinutes(1));
+
+// Create generic async rate limiter
+IAsyncPolicy<HttpResponseMessage> asyncPolicyT = Policy.RateLimitAsync<HttpResponseMessage>(
+    numberOfExecutions: 100,
+    perTimeSpan: TimeSpan.FromMinutes(1));
+```
+<!-- endSnippet -->
+
+### Rate limit in v8
+
+<!-- snippet: migration-rate-limit-v8 -->
+```cs
+// The equivalent to Polly v7's RateLimit is the SlidingWindowRateLimiter.
+//
+// Polly exposes just a simple wrapper to the APIs exposed by the System.Threading.RateLimiting APIs.
+// There is no need to create separate instances for sync and async flows as ResiliencePipeline handles both scenarios.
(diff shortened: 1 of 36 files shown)


src/Snippets/Docs/Migration.Bulkhead.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Net.Http;
+
+namespace Snippets.Docs;
+
+internal static partial class Migration
```
adamnova, line 5: I like the structure of these migrations. I have two small suggestions:

1. I am not sure if grouping multiple calls into one migration is the best way, because it might confuse those who don't really understand Polly that much and are just doing migration. So that when they look at the code they see that these 4 commands are replaced with 2, how what why? Suddenly there is no async, they might not be sure what to do if they happen to just look at the code. I understand that there will be future notes in the migration guide.
2. I would probably add simple comments to the Migration snippets like they are in Context. It makes the code easier to understand. It can be kind of reassuring that you know the intent behind the lines in the snippet. I am not certain that every line needs it, just a food for thought.

src/Snippets/Docs/Migration.Retry.cs
```diff
@@ -0,0 +1,138 @@
+﻿using Polly.Retry;
+using Snippets.Docs.Utils;
+
+namespace Snippets.Docs;
+
+internal static partial class Migration
+{
+    public static void Retry_V7()
```
adamnova, line 8: Could be kind of overwhelming with so many options, maybe could be broken down into individual snippets? Like if you use this it migrates to this. More work but probably easier to understand.

adamnova (commented):
I like the changes mostly, but I can see some of the migrations being harder to navigate with so many options and nothing particularly matching one option to another.

src/Snippets/Docs/Migration.Bulkhead.cs
martintmk: Good point I'll try to add comments for more context into the snippets

src/Snippets/Docs/Migration.Retry.cs
martintmk: True, split the samples into more sections

docs/migration-v8.md
```diff
@@ -1 +1,515 @@
...
+
+> [!NOTE]
+> The v7 API is still available and fully supported even when using the v8 version.
+
+## Major differences
+
+- **The term *Policy* is now replaced with *Strategy***: In previous versions, Polly used the term *policy* for retries, timeouts, etc. In v8, these are referred to as *resilience strategies*.
+- **Introduction of Resilience Pipelines**: The [resilience pipeline](pipelines/index.md) combines one or more resilience strategies. This is the foundational API for Polly v8, similar to the **Policy Wrap** in previous versions but integrated into the core API.
+- **Unification of interfaces**: Interfaces such as `IAsyncPolicy`, `IAsyncPolicy<T>`, `ISyncPolicy`, `ISyncPolicy<T>`, and `IPolicy` are now unified under `ResiliencePipeline` and `ResiliencePipeline<T>`. The resilience pipeline supports both synchronous and asynchronous executions.
+- **Native async support**: Polly v8 was designed with asynchronous support from the start.
```
adamnova, line 13: I would probably change this into "Unified sync/async flow" or something like that, because Polly had native async support even in v7, although you had to implement it separately from sync.

docs/migration-v8.md
```diff
@@ -1 +1,515 @@
...
+    // your code here
+    return await GetResponseAsync(cancellationToken);
+},
+cancellationToken);
+```
+<!-- endSnippet -->
+
+Here are the primary changes:
+
+- `ResiliencePipelineBuilder` is used to start building the resilience pipeline, instead of the static `Policy.HandleException<TException>(...)`.
```
adamnova, line 146: I think these things should be code comments as well, so that readers don't have to go back and forth to understand things. It's always easier to understand what has changed if you can see the differences in smaller chunks rather than at bulk.

docs/migration-v8.md
```diff
@@ -1 +1,515 @@
...
+```
+<!-- endSnippet -->
+
+### Policy wrap in v8
+
+In v8, there's no need to use policy wrap explicitly. Instead, policy wrapping is smoothly integrated into `ResiliencePipelineBuilder`, as shown in the example below:
+
+<!-- snippet: migration-policy-wrap-v8 -->
+```cs
+// The "PolicyWrap" is integrated directly. Strategies are executed in the same order as they were added:
```
adamnova, line 178: This is exactly the comment I think is the best for understanding the changes. If I just look at the code sample, it's enough to understand what has changed from it directly and easily.

docs/migration-v8.md
```diff
@@ -1 +1,515 @@
...
+```
+<!-- endSnippet -->
+
+- It's important to remember that the configuration in v8 is options based, i.e. `RetryStrategyOptions` are used.
+- Notice that to disable waiting between retries the `Delay` property is set to `TimeSpan.Zero`.
+- To retry forever set `MaxRetryAttempts` to `int.MaxValue`.
+
+## Migrating bulkhead policy
+
+The bulkhead policy is now replaced by the [rate limiter strategy](strategies/rate-limiter.md) which uses the [`System.Threading.RateLimiting`](https://www.nuget.org/packages/System.Threading.RateLimiting) package. The new counterpart to bulkhead is `ConcurrencyLimiter`.
```
adamnova, line 355: I would probably put an extra sentence here about what is the difference between Bulkhead and Rate-limiting. Why is it that its suddenly called something different, it feels like that there must be lots of changes but there are just 2 sentences that don't say much in that regard.

docs/migration-v8.md
```diff
@@ -1 +1,515 @@
...
+
+// Create generic sync bulkhead
+ISyncPolicy<HttpResponseMessage> syncPolicyT = Policy.Bulkhead<HttpResponseMessage>(maxParallelization: 100, maxQueuingActions: 50);
+
+// Create generic async bulkhead
+IAsyncPolicy<HttpResponseMessage> asyncPolicyT = Policy.BulkheadAsync<HttpResponseMessage>(maxParallelization: 100, maxQueuingActions: 50);
+```
+<!-- endSnippet -->
+
+### Concurrency limiter in v8
```
adamnova, line 375: I am not 100% sure, but it might be a good idea just for completeness put Bulkhead in the title as well, in case people don't read and just go by headers, so that it's immediately obvious they belong together.

docs/migration-v8.md
```diff
@@ -1 +1,515 @@
...
+// Return the context to the pool
+ResilienceContextPool.Shared.Return(context);
+```
+<!-- endSnippet -->
+
+For more details, refer to the [Resilience context](advanced/resilience-context.md) documentation.
+
+## Migrating no-op policy
+
+- For `Policy.NoOp` or `Policy.NoOpAsync`, now use `ResiliencePipeline.Empty`.
```
adamnova, line 458: nit: I would probably use the same `now use` or `switch to` for both since they are there just twice and I actually reread the first line again to make sure what I was supposed to do.

docs/migration-v8.md
```diff
@@ -1 +1,515 @@
...
+In v7, the following registry APIs are exposed:
+
+- `IPolicyRegistry<T>`
+- `IReadOnlyPolicyRegistry<T>`
+- `IConcurrentPolicyRegistry<T>`
+- `PolicyRegistry<T>`
+
+In v8, these have been replaced by:
+
+- `ResiliencePipelineProvider<TKey>`
```
adamnova, line 472: It might be good idea just to mention what this is good for.

docs/migration-v8.md
martintmk: Done.

docs/migration-v8.md
martintmk: Done.

docs/migration-v8.md
martintmk: Done.

adamnova (approved):

PR #1608, opened 2023-09-21
[Docs] Testing

martintmk:

Contributes to #1091 

### Confirm the following


docs/advanced/testing.md
```diff
@@ -0,0 +1,63 @@
...
+        MaxRetryAttempts = 4
+    })
+    .AddTimeout(TimeSpan.FromSeconds(1))
+    .Build();
+
+// Retrieve the descriptor.
+ResiliencePipelineDescriptor descriptor = pipeline.GetPipelineDescriptor();
+
+// Check the pipeline's composition with the descriptor.
+Assert.Equal(2, descriptor.Strategies.Count);
```
adamnova, line 32: I would maybe add a check here that the first strategy is `RetryStrategy`? It is true that there is options check, but to me it fit the description.

martintmk: Not really possible because strategy implementations are internal. Checking for options type fulfils that role.

adamnova: ok I thought that might be the case, because I remember some talks about it. Maybe just tweak the "composition" word in the comment? "composition" implied to me that we are checking the structure.

martintmk: But we are checking the composition. For example:

``` csharp
// Verify the retry settings.
var retryOptions = Assert.IsType<RetryStrategyOptions>(descriptor.Strategies[0].Options);
Assert.Equal(4, retryOptions.MaxRetryAttempts);
```

This checks that the first strategy is retry and it has correct configuration.

adamnova: ok I see, so basically it was like title header for everything past it. That makes sense.

PR #1609, opened 2023-09-21
Avoid capturing where possible.

IEvangelist:

All of these changes are only within the `Docs/Snippets` content.

When reviewing docs, I noticed that some of the example snippets were _lifting_ (capturing outer scope variables) when it clearly wasn't intended as the `state` parameter was passed an argument with the same name. Ideally, we shouldn't be capturing outer scope variables when using the `state` parameter.

In this PR:

- Some cases are simply resolved by adding the `static` modifier.
- Other cases involved renaming the delegate parameter for the inner scope to isolate the variable usage.
- And a few cases required passing `state`.

### Confirm the following

### DIFF
--- a/README_V8.md
+++ b/README_V8.md
@@ -48,7 +48,7 @@ ResiliencePipeline pipeline = new ResiliencePipelineBuilder()
     .Build(); // Builds the resilience pipeline
 
 // Execute the pipeline asynchronously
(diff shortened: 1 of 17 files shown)


martintmk (approved):
Please also run `dotnet mdsnippets` from the Polly root to update the relevant docs.

PR #1625, opened 2023-09-25
Fix the behavior of `OnHedging` event

martintmk:

- `OnHedging` was called even for last attempt, it caused that additional event to be reported.
- Now the `OnHedging` is called consistently just before the hedged task starts executing.
- Made changes to `OnHedgingArguments<T>`
  - Dropped `Outcome` and `Duration` properties as these are giving non-deterministic results based on what previous task finished first. If someone needs these we can add it later, however, with different API that allows deterministic results, for example access by attempt number.
  - Added `ActionContext` property to be consistent with `HedgingActionGeneratorArguments<T>`

### Confirm the following


src/Polly.Core/Hedging/Controller/TaskExecution.cs
```diff
@@ -137,6 +139,21 @@ public void Cancel()
...
+    private async Task HandleOnHedgingAsync(ResilienceContext primaryContext, int attemptNumber)
+    {
+        var args = new OnHedgingArguments<T>(
+            primaryContext,
+            Context,
+            attemptNumber);
+
+        _telemetry.Report(new(ResilienceEventSeverity.Warning, HedgingConstants.OnHedgingEventName), Context, args);
+
+        if (_handler.OnHedging is not null)
```
martincostello, line 151: What about capturing it into a local like `is {} callback` and then invoking that?

src/Polly.Core/Hedging/OnHedgingArguments.cs
```diff
@@ -14,36 +14,31 @@ namespace Polly.Hedging;
     /// <summary>
     /// Initializes a new instance of the <see cref="OnHedgingArguments{TResult}"/> struct.
     /// </summary>
-    /// <param name="outcome">The context in which the resilience operation or event occurred.</param>
-    /// <param name="context">The outcome of the resilience operation or event.</param>
+    /// <param name="primaryContext">The outcome of the resilience operation or event.</param>
```
martincostello, line 17: I think this needs some more explanation on what the difference is between "primary" and "action".

src/Polly.Core/Hedging/OnHedgingArguments.cs
```diff
@@ -14,36 +14,31 @@ namespace Polly.Hedging;
...
+    /// Gets the action context that will be used for the hedged action.
     /// </summary>
-    public ResilienceContext Context { get; }
+    /// <remarks>
+    /// This context is cloned from <see cref="PrimaryContext"/>.
+    /// </remarks>
+    public ResilienceContext ActionContext { get; }
 
     /// <summary>
     /// Gets the zero-based hedging attempt number.
```
martincostello, line 41: Is this the correct explanation? Looking at this fresh, if this is zero-based shouldn't we call it an index rather than a number?

src/Polly.Core/Hedging/OnHedgingArguments.cs
martintmk: Added remarks and enhanced the docs.

src/Polly.Core/Hedging/OnHedgingArguments.cs
martintmk: IF you don't mind i'll leave it as it is because we agreed on this in the API review and to me it sounds nicer that `AttemptIndex`.

src/Polly.Core/Hedging/HedgingActionGeneratorArguments.cs
```diff
@@ -7,17 +7,20 @@ namespace Polly.Hedging;
 /// </summary>
 /// <typeparam name="TResult">The type of the result.</typeparam>
 /// <remarks>
+/// The <see cref="PrimaryContext"/> represents the context that was received by the hedging strategy and used to execute the primary action.
+/// To prevent race conditions, hedging strategy then clones the primary context into <see cref="ActionContext"/> and uses it to execute the hedged action.
+/// Every hedged action gets its own context that is cloned from primary.
```
martincostello, line 12:
```suggestion
/// To prevent race conditions, the hedging strategy then clones the primary context into <see cref="ActionContext"/> and uses it to execute the hedged action.
/// Every hedged action gets its own context that is cloned from the primary.
```

src/Polly.Core/Hedging/HedgingActionGeneratorArguments.cs
```diff
@@ -7,17 +7,20 @@ namespace Polly.Hedging;
...
 {
     /// <summary>
     /// Initializes a new instance of the <see cref="HedgingActionGeneratorArguments{TResult}"/> struct.
     /// </summary>
-    /// <param name="primaryContext">The primary resilience context.</param>
-    /// <param name="actionContext">
-    /// The context that will be passed to action generated by <see cref="HedgingStrategyOptions{TResult}.ActionGenerator"/>.
-    /// .</param>
+    /// <param name="primaryContext">The primary context received by hedging strategy.</param>
+    /// <param name="actionContext">The action context. cloned from the primary context.</param>
```
martincostello, line 23:
```suggestion
    /// <param name="primaryContext">The primary context received by the hedging strategy.</param>
    /// <param name="actionContext">The action context cloned from the primary context.</param>
```

src/Polly.Core/Hedging/HedgingActionGeneratorArguments.cs
```diff
@@ -33,12 +36,12 @@ namespace Polly.Hedging;
     }
 
     /// <summary>
-    /// Gets the primary resilience context.
+    /// Gets the primary resilience context as received by hedging strategy.
```
martincostello, line 39:
```suggestion
    /// Gets the primary resilience context as received by the hedging strategy.
```

src/Polly.Core/Hedging/OnHedgingArguments.cs
```diff
@@ -14,9 +14,14 @@ namespace Polly.Hedging;
     /// <summary>
     /// Initializes a new instance of the <see cref="OnHedgingArguments{TResult}"/> struct.
     /// </summary>
-    /// <param name="primaryContext">The outcome of the resilience operation or event.</param>
-    /// <param name="actionContext">The action context.</param>
+    /// <param name="primaryContext">The primary context received by hedging strategy.</param>
+    /// <param name="actionContext">The action context. cloned from the primary context.</param>
```
martincostello, line 18:
```suggestion
    /// <param name="primaryContext">The primary context received by the hedging strategy.</param>
    /// <param name="actionContext">The action context cloned from the primary context.</param>
```

src/Polly.Core/Hedging/OnHedgingArguments.cs
```diff
@@ -14,9 +14,14 @@ namespace Polly.Hedging;
...
     /// <param name="attemptNumber">The zero-based hedging attempt number.</param>
+    /// <remarks>
+    /// The <paramref name="primaryContext"/> represents the context that was received by the hedging strategy and used to execute the primary action.
+    /// To prevent race conditions, hedging strategy then clones the primary context into <paramref name="actionContext"/> and uses it to execute the hedged action.
+    /// Every hedged action gets its own context that is cloned from primary.
```
martincostello, line 23:
```suggestion
    /// To prevent race conditions, the hedging strategy then clones the primary context into <paramref name="actionContext"/> and uses it to execute the hedged action.
    /// Every hedged action gets its own context that is cloned from the primary.
```

src/Polly.Core/Hedging/OnHedgingArguments.cs
```diff
@@ -25,7 +30,7 @@ public OnHedgingArguments(ResilienceContext primaryContext, ResilienceContext ac
     }
 
     /// <summary>
-    /// Gets the primary resilience context.
+    /// Gets the primary resilience context as received by hedging strategy.
```
martincostello, line 33:
```suggestion
    /// Gets the primary resilience context as received by the hedging strategy.
```

src/Polly.Core/Hedging/OnHedgingArguments.cs
martincostello: Ok, but I would expect the first value to be one if it's a number. I remember discussing calling it `Number` but I thought that was in the context of retries and it wasn't zero-based.

martincostello (approved):

src/Polly.Core/Hedging/OnHedgingArguments.cs
martintmk: For retries it's also called `AttemptNumber` and it's also zero-based. It's just the name we agreed on. Docs make it clear it starts from zero.

PR #1630, opened 2023-09-26
Documentation updates

martincostello:
Apply some grammar and wording changes to the documentation for v8.

I'm slowly making my way through the PRs from while I was on holiday and making edits that I would have suggested via PR review. Once I've caught up and move this out of draft for review.

### DIFF
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -1,20 +1,20 @@
 # Migration guide from v7 to v8
 
-Welcome to the migration guide for Polly's v8 version. The v8 version of Polly brings major new enhancements and supports all the same scenarios as previous versions. In the following sections, we'll detail the differences between the v7 and v8 APIs and provide steps on how to transition smoothly.
+Welcome to the migration guide for Polly's v8 release. Version 8 of Polly brings major new enhancements and supports all of the same scenarios as previous versions. In the following sections, we'll detail the differences between the v7 and v8 APIs, and provide steps on how to transition smoothly.
 
 > [!NOTE]
-> The v7 API is still available and fully supported even when using the v8 version by referencing [Polly](https://www.nuget.org/packages/Polly) package.
+> The v7 API is still available and fully supported even when using the v8 version by referencing the [Polly](https://www.nuget.org/packages/Polly) package.
 
 ## Major differences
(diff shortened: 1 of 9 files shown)


src/Snippets/Docs/Migration.Policies.cs
```diff
@@ -106,14 +106,14 @@ public static async Task Strategies()
         // Synchronous execution
         pipelineT.Execute(() =>
         {
-            // your code here
+            // Your code here
```
martintmk, line 109: I am always undecided whether we should end these comments with period or not. Maybe we can agree on some pattern an apply it everywhere?

martincostello: I don't mind either way. Typically I do `.` for multiple lines/sentences and omit for for one-liners.

docs/migration-v8.md
```diff
@@ -186,9 +186,9 @@ ResiliencePipeline pipeline = new ResiliencePipelineBuilder()
 <!-- endSnippet -->
 
 > [!IMPORTANT]
-> In v7, the policy wrap ordering is different; the policy added first was executed last. In v8, the execution order matches the order in which they were added.
+> In v7, the policy wrap ordering is different; the policy added first was executed last (FILO). In v8, the execution order matches the order in which they were added (FIFO).
 
-## Migrating retry policy
+## Migrating Retry Policies
```
martintmk, line 191: Any reason to capitalize the headings here?

In-general, only first word is capitalized.

martincostello: I changed them to Title Case as I thought headings were to be considered as titles.

martintmk: I think it's better to leave them as before, for consistency and alignment with official docs guidelines.

martincostello: Ok I'll change them back. Do you have a link to these guidelines so I can take a look?

martintmk: https://learn.microsoft.com/en-us/contribute/content/dotnet/dotnet-style-guide#headings

martintmk (approved):
Outside of that capitalization thing it looks good! Great job :)

PR #1638, opened 2023-09-27
[Docs] Add safe execution section to the migration guide

peter-csala:
- Added a new section about safe execution to the migration guide

### Confirm the following


src/Snippets/Docs/Migration.Execute.cs
```diff
@@ -0,0 +1,100 @@
...
+        ResiliencePipeline<int> pipeline = new ResiliencePipelineBuilder<int>()
+            .AddTimeout(TimeSpan.FromSeconds(1))
+            .Build();
+
+        // Synchronous execution
+        // Polly v8 does not provide an API to synchronously execute and capture the outcome of a pipeline
+
+        // Asynchronous execution
+        var context = ResilienceContextPool.Shared.Get();
+        Outcome<int> pipelineResult = await pipeline.ExecuteOutcomeAsync(
```
martintmk, line 59: @martincostello , @peter-csala 

Wondering whether it makes sense to add more `ExecuteOutcome` convenience overloads so we do not force passing `ResilienceContext` and `state`?

Or maybe wait for feedback until someone really needs these?

src/Snippets/Docs/Migration.Execute.cs
```diff
@@ -0,0 +1,100 @@
...
+                {
+                    args.Context.Properties.Set(contextKey, "context_value");
+                    return default;
+                }
+            })
+            .Build();
+
+        context = ResilienceContextPool.Shared.Get();
+        pipelineResult = await pipelineWithContext.ExecuteOutcomeAsync(
+            static async (ctx, state) => Outcome.FromResult(await MethodAsync()), context, "state");
```
martintmk, line 94:
```suggestion
            static async (ctx, state) => Outcome.FromResult(await MethodAsync(ctx.CancellationToken)), context, "state");
```

Let's pass cancellation to async method to be alligned with official guidelines.

src/Snippets/Docs/Migration.Execute.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        // Asynchronous execution
+        IAsyncPolicy<int> asyncPolicy = Policy.TimeoutAsync<int>(TimeSpan.FromSeconds(1));
+        PolicyResult<int> asyncPolicyResult = await asyncPolicy.ExecuteAndCaptureAsync(MethodAsync);
+
+        // Assess policy result
+        if (policyResult.Outcome == OutcomeType.Successful)
+        {
+            var result = policyResult.Result;
+            // process result
```
martintmk, line 24:
```suggestion
            // Process result
```

Capitalize the first letter (applies to other comments too)

src/Snippets/Docs/Migration.Execute.cs
```diff
@@ -0,0 +1,100 @@
...
+        PolicyResult<int> policyResult = syncPolicy.ExecuteAndCapture(Method);
+
+        // Asynchronous execution
+        IAsyncPolicy<int> asyncPolicy = Policy.TimeoutAsync<int>(TimeSpan.FromSeconds(1));
+        PolicyResult<int> asyncPolicyResult = await asyncPolicy.ExecuteAndCaptureAsync(MethodAsync);
+
+        // Assess policy result
+        if (policyResult.Outcome == OutcomeType.Successful)
+        {
+            var result = policyResult.Result;
```
martintmk, line 23: Maybe have explicit type here to see what is being returned?

src/Snippets/Docs/Migration.Execute.cs
martincostello: Should be easy enough for people to just use a discard if they don't need them?

src/Snippets/Docs/Migration.Execute.cs
```diff
@@ -0,0 +1,100 @@
...
+        var context = ResilienceContextPool.Shared.Get();
+        Outcome<int> pipelineResult = await pipeline.ExecuteOutcomeAsync(
+            static async (ctx, state) => Outcome.FromResult(await MethodAsync(ctx.CancellationToken)), context, "state");
+        ResilienceContextPool.Shared.Return(context);
+
+        // Assess policy result
+        if (pipelineResult.Exception is null)
+        {
+            int result = pipelineResult.Result;
+            // Process result
```
martincostello, line 67:
```suggestion
            // Process result
```

src/Snippets/Docs/Migration.Execute.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+        else
+        {
+            Exception exception = pipelineResult.Exception;
+            // Process failure
```
martincostello, line 72:
```suggestion
            // Process failure
```

src/Snippets/Docs/Migration.Execute.cs
peter-csala: IMHO it would definitely ease the usage of the `ExecuteOutcomeAsync` method.

src/Snippets/Docs/Migration.Execute.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            // If needed then you can rethrow the exception
```
martincostello, line 74:
```suggestion
            // If needed you can rethrow the exception
```

src/Snippets/Docs/Migration.Execute.cs
martintmk: One reason why there is only a single method is that I considered this advanced API that won't be used that much. But maybe I am wrong and it will be commonly used. 

In that case having some convenience overloads makes sense (don't force  people to use `ResilienceContextPool.Shared.Get()`). 

But maybe let's just have "wait and see" approach? Adding new APIs won't be a breaking change anyway.

martincostello (approved):

martintmk (approved):

martincostello:
Looks like this has picked up some merge conflicts.

martincostello (approved):

PR #1641, opened 2023-09-27
Update docs source

martincostello:
Rename exception variable.

See https://github.com/App-vNext/Polly/pull/1640#issuecomment-1737177624.

### DIFF
--- a/src/Snippets/Docs/Fallback.cs
+++ b/src/Snippets/Docs/Fallback.cs
@@ -125,9 +125,9 @@ internal static class Fallback
                 return Outcome.FromResult(result);
             }, context, "state");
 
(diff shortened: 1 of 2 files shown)


src/Snippets/Docs/Fallback.cs
```diff
@@ -125,9 +125,9 @@ public static async ValueTask<HttpResponseMessage> Action()
                 return Outcome.FromResult(result);
             }, context, "state");
 
-        if (outcome.Exception is HttpRequestException hre)
+        if (outcome.Exception is HttpRequestException httpEx)
```
martintmk, line 128:
```suggestion
        if (outcome.Exception is HttpRequestException requestException)
```
?

PR #1667, opened 2023-09-29
Add spell checker

### LINKED ISSUE #1639: Introduce linting for the docs
As more content is being added to the `/docs` we might consider adding linter to the build to verify things such as:

- Markdown syntax
- Grammar

martincostello:
Add a spell checker for Markdown files.

Resolves #1639.


martincostello:
I'll come back to this later. The process is working, but it's picking up _a lot_ of false positives, like GitHub logins and .NET method names etc.

IEvangelist (approved):

martincostello:
I've fixed (or "fixed") a bunch more errors so there's a lot fewer now.

Most of what's left are type names or GitHub user logins.

I might need to ignore a few more files rather than add everyone to the wordlist file, but ignoring files seems to be a bit of a faff: https://github.com/rojopolis/spellcheck-github-actions/issues/176

Will come back to this again later.

PR #1677, opened 2023-10-04
Enable CA1068

davidCett:
# Pull Request

Cleanup the Polly and Polly.Specs codebase #1290

Enable CA1068 warning (CancellationToken parameters must come last) and code cleanup

## Confirm the following


davidCett:
@dotnet-policy-service agree

src/Polly/AsyncPolicy.GenericImplementation.cs
```diff
@@ -13,7 +13,7 @@ public abstract partial class AsyncPolicy<TResult>
     protected abstract Task<TResult> ImplementationAsync(
         Func<Context, CancellationToken, Task<TResult>> action,
         Context context,
-        CancellationToken cancellationToken,
-        bool continueOnCapturedContext
+        bool continueOnCapturedContext,
```
martincostello, line 16: Just scan reading this PR before a proper review tomorrow, the public API surface area like this **must not** be changed to resolve these warnings. Anything this isn't internal or private must be suppressed only.

Changing the order of the parameters in the public API surface is a breaking change we cannot accept.

martincostello (changes requested):

martincostello:
As you can [see in the build log](https://github.com/App-vNext/Polly/actions/runs/6409461852/job/17403157798?pr=1677#step:5:674), the Public API analyser flags the changes to the public surface of the library and breaks the build.

For example:

```
Error: C:\Program Files\dotnet\sdk\7.0.401\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.ApiCompat.ValidatePackage.targets(39,5): error CP0002: Member 'Polly.AsyncPolicy.ExecuteAsync(System.Func<System.Threading.CancellationToken, System.Threading.Tasks.Task>, System.Threading.CancellationToken, bool)' exists on [Baseline] lib/net461/Polly.dll but not on lib/netstandard2.0/Polly.dll [D:\a\Polly\Polly\src\Polly\Polly.csproj]
```

src/Polly/PublicAPI.Shipped.txt
martincostello, line 1: This is the wrong thing to do - the file is effectively set in stone and should not be edited.

davidCett: I get it. My bad.

davidCett:
Thank you for de review. I will work on it😀

PR #1682, opened 2023-10-09
Simmy v8 feedback

vany0114:
# Pull Request
Address the remaining feedback in the initial implementation and also the feedback on the Simmy API (https://github.com/App-vNext/Polly/pull/1571).

* Splits fault strategy from outcomes strategy
* Removes `Outcome` property from options and relies only on `OutcomeGenerator` to avoid issues with disposable outcomes
* Removes enabled parameter from all convenience extensions

## Confirm the following


src/Polly.Core/Simmy/Fault/FaultPipelineBuilderExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+internal static partial class FaultPipelineBuilderExtensions
+{
+    /// <summary>
+    /// Adds a fault chaos strategy to the builder.
+    /// </summary>
+    /// <param name="builder">The builder instance.</param>
+    /// <param name="injectionRate">The injection rate for a given execution, which the value should be between [0, 1] (inclusive).</param>
+    /// <param name="fault">The exception to inject.</param>
+    /// <returns>The builder instance with the retry strategy added.</returns>
+    public static ResiliencePipelineBuilder AddChaosFault(this ResiliencePipelineBuilder builder, double injectionRate, Exception fault)
```
martintmk, line 18:
```suggestion
    public static TBuilder AddChaosFault<TBuilder>(this TBuilder builder, double injectionRate, Exception fault) where TBuilder: ResiliencePipelineBuilderBase
```

This will allow you to reduce a bunch od duplicate extensions.

src/Polly.Core/Simmy/Outcomes/OutcomeChaosStrategy.cs
```diff
@@ -8,65 +8,39 @@ internal class OutcomeChaosStrategy<T> : MonkeyStrategy<T>
...
-        Fault = options.Fault;
-        OnFaultInjected = options.OnFaultInjected;
-        FaultGenerator = options.FaultGenerator is not null ? options.FaultGenerator : (_) => new(options.Fault);
-    }
-
     public OutcomeChaosStrategy(OutcomeStrategyOptions<T> options, ResilienceStrategyTelemetry telemetry)
         : base(options)
     {
-        if (options.Outcome is null && options.OutcomeGenerator is null)
+        if (options.OutcomeGenerator is null)
```
martintmk, line 14: Make the property required and this check won't be necessary anymore.

src/Polly.Core/Simmy/Outcomes/OutcomeStrategyOptions.TResult.cs
```diff
@@ -19,18 +21,6 @@ internal class OutcomeStrategyOptions<TResult> : MonkeyStrategyOptions
...
-    /// <summary>
-    /// Gets or sets the outcome to be injected for a given execution.
-    /// </summary>
-    /// <remarks>
-    /// Defaults to <see langword="null"/>. Either <see cref="OutcomeGenerator"/> or this property is required.
-    /// When this property is <see langword="null"/> the <see cref="OutcomeGenerator"/> is used.
-    /// </remarks>
-    public Outcome<TResult>? Outcome { get; set; }
+    [Required]
+    public Func<OutcomeGeneratorArguments, ValueTask<Outcome<TResult>?>> OutcomeGenerator { get; set; }
```
martintmk, line 25:
```suggestion
    public Func<OutcomeGeneratorArguments, ValueTask<Outcome<TResult>?>>? OutcomeGenerator { get; set; }
```

it's null bu default because we cannot assign it any sensible default value.

src/Polly.Core/Simmy/Outcomes/OutcomePipelineBuilderExtensions.cs
```diff
@@ -6,76 +6,77 @@ namespace Polly.Simmy;
...
     /// <param name="builder">The builder instance.</param>
-    /// <param name="enabled">A value that indicates whether or not the chaos strategy is enabled for a given execution.</param>
     /// <param name="injectionRate">The injection rate for a given execution, which the value should be between [0, 1] (inclusive).</param>
-    /// <param name="faultGenerator">The exception generator delegate.</param>
+    /// <param name="outcomeGenerator">The outcome generator delegate.</param>
     /// <returns>The builder instance with the retry strategy added.</returns>
-    public static ResiliencePipelineBuilder AddChaosFault(
-        this ResiliencePipelineBuilder builder, bool enabled, double injectionRate, Func<Exception?> faultGenerator)
+    public static ResiliencePipelineBuilder<TResult> AddChaosResult<TResult>(
+        this ResiliencePipelineBuilder<TResult> builder, double injectionRate, Func<TResult?> outcomeGenerator)
```
martintmk, line 41:
```suggestion
        this ResiliencePipelineBuilder<TResult> builder, double injectionRate, Func<TResult?> resultGenerator)
```

src/Polly.Core/Simmy/Fault/FaultChaosStrategy.cs
```diff
@@ -0,0 +1,56 @@
+﻿using Polly.Telemetry;
+
+namespace Polly.Simmy.Fault;
```
martincostello, line 3: (and any other place VS might have created new files like this)

```suggestion
namespace Polly.Simmy.Fault;

```

src/Polly.Core/Simmy/Fault/FaultChaosStrategy.cs
```diff
@@ -0,0 +1,56 @@
...
+internal class FaultChaosStrategy<T> : MonkeyStrategy<T>
```
martincostello, line 4: If this is going to stay internal (rather than just being internal for now to hide the public API from v8), can it be `sealed` to improve the ability of the JIT to inline?

src/Polly.Core/Simmy/Fault/FaultChaosStrategy.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        _telemetry = telemetry;
+        Fault = options.Fault;
+        OnFaultInjected = options.OnFaultInjected;
+        FaultGenerator = options.FaultGenerator is not null ? options.FaultGenerator : (_) => new(options.Fault);
+    }
+
+    public Func<OnFaultInjectedArguments, ValueTask>? OnFaultInjected { get; }
+
+    public Func<FaultGeneratorArguments, ValueTask<Exception?>>? FaultGenerator { get; }
```
martincostello, line 24: This doesn't appear to need to be nullable as the constructor makes it non-null. That then means you can delete the `!` on line 34.

src/Polly.Core/Simmy/Outcomes/OutcomeChaosStrategy.cs
vany0114: Good catch, it was actually already required.

src/Polly.Core/Simmy/Outcomes/OutcomeStrategyOptions.TResult.cs
vany0114: It cannot be null since now we don't have the `Outcome` property just the `OutcomeGenerator`, see your previous comment https://github.com/App-vNext/Polly/pull/1682#discussion_r1349917952

src/Polly.Core/Simmy/Fault/FaultChaosStrategy.cs
vany0114: I think this isn't going to be hidden for a long time, so I would avoid it, however, can you expand on how sealed classes would improve things?

martintmk: I think `FaultChaosStrategy` should stay hidden though? It's implementation details of fault strategy.

src/Polly.Core/Simmy/Fault/FaultPipelineBuilderExtensions.cs
martintmk: @vany0114 , unresolving as this is not yet addressed

vany0114: @martintmk  I did try your suggestion however it does not work, am I missing something? 

image

I also took a look at the Retry extensions and it is not implementing using that approach either, I guess it's because it does not work for generic strategies.

martintmk: It's still generic, but the generic constraint is on the method instead:
https://github.com/App-vNext/Polly/blob/a3ca9d070a8921e0cec54f8f43b26647ca4ec795/src/Polly.Core/ResilienceStrategy.cs#L23

Because you are just injection exceptions, there is no need to scope the generic to a single constrained generic parameter.

vany0114: Gotcha, I'll take a look next weekend then, I tried changing the class signature and it broke the unit tests.

vany0114: Done

Issue #1687, opened 2023-10-11
Make ResilienceContextPool settable per pipeline


cmeyertons:
# Feature Request

## Is your feature request related to a specific problem? Or an existing feature?
Specific problem -- in an Orleans grain, we should by default continue on the capturing context to avoid issues with grain losing required context to operate within the framework's custom task scheduler, etc.

This has been mentioned before  #676 but received no traction, despite others mentioning Orleans specifically.

## Describe your proposed or preferred solution
~~This repo did a great job defining a `PolicyBase.DefaultContinueOnCapturedContext` internal const.  This issue should aim to publicize that variable as a static property.~~

^ the above refers to pre-v8 Polly.  This issue will aim to address this in v8 using the `ResiliencePipelines` arch by making the `ResiliencePipelineBuilder` have a configurable `ContextPool` that supplies this behavior.

## Describe any alternative options you've considered
Teams _can_ create their own overloads to accomplish this behavior -- however, this is generally unsafe and requires creating & maintaining 20+ overloads.  It also requires more contextual knowledge within the app to "know" to use the correct overload and leaves the door open to unsafe behavior by default (someone familiar w/ Polly will try to use the base Polly library)

martintmk:
I am always little hesitant exposing APIs that are static and can change the global behavior. This is that case. Have you considered the new v8 version?

Archiving this in v8 should be relatively straightforward. First define your custom pool that changes the default `ContinueOnCapturedContextValue`:

``` csharp
public class CustomResilienceContextPool : ResilienceContextPool
{
    public override ResilienceContext Get(ResilienceContextCreationArguments arguments)
    {
        if (arguments.ContinueOnCapturedContext is null)
        {
            arguments = new ResilienceContextCreationArguments(arguments.OperationKey, continueOnCapturedContext: true, arguments.CancellationToken);
        }

        return Shared.Get(arguments);
    }

    public override void Return(ResilienceContext context) => ResilienceContextPool.Shared.Return(context);
}
```

Now, assign this pool to `ResiliencePipelineBuilder`:

``` csharp
var builder = new ResiliencePipelineBuilder()
{
   ResilienceContextPool = new CustomResilienceContextPool();
}

var pipeline = builder.Build();
```

Now, whenever you execute any action using `pipeline` the new default will be respected. Please note that the `ResilienceContextPool` property is not yet exposed in the builder but it should be straightforward to introduce.

martintmk:
[...]

PR #1688, opened 2023-10-11
Publicize DefaultContinueOnCapturedContext

cmeyertons:
# Pull Request

#1687

The issue describes the PR pretty well.  This PR **does not** introduce any binary breaking changes and should be safe to release via a minor release.  

To avoid overload confusion, I put the method with the actual logic into its own `xxxInternalAsync` submethod to differentiate between the `bool?` and `bool` overloads.  This has the added benefit of making the decompiled source easier to navigate as Visual Studio tends to get confused w/ heavily overloaded methods in these kinds of files.

## Confirm the following

- [x]  I have included unit tests for the issue/feature - _I don't think any new tests are needed here?_
- [x] Still need to do the `Result` overloads

^^ Still working on tests


martintmk:
What's the reason for this change?

The main problem I see is that this exposes a static API that also affects all other policies when the  `DefaultContinueOnCapturedContext` is changed.

Issue #722, opened 2019-11-14
Return timedOut, but don't cancel task, let exception come and log exception [=TimeoutStrategy.Pessimistic]


aalia:
I am using Retry and timeout policy for my http client.
It normally takes more time for an exception to come and then if we retry it for say 3 times, it will take even more time. We want to reduce this time. So, what we want is, to set a timeout, even if exception haven't occurred, it should timeout. But, because we need the information for exception, we want to continue the call in a background thread may be, so that we can log the exception details.

Following is the code that I have done till now, we are using .NET core, so adding in Startup.cs:

```
 services.AddHttpClient<IClient,
                            Client>("Client", c => { c.DefaultRequestHeaders.Add("Accept", "application/json"); })
 
            .SetHandlerLifetime(Timeout.InfiniteTimeSpan)
            .AddPolicyHandler(GetFallbackPolicy())
            .AddPolicyHandler(GetRetryPolicy())
            .AddPolicyHandler(GetTimeOutPolicy());
			
			private IAsyncPolicy<HttpResponseMessage> GetRetryPolicy()
        {
            return HttpPolicyExtensions
                    .HandleTransientHttpError() 
                    .OrInner<WebException>(exception => new[] {WebExceptionStatus.ConnectionClosed,
                  WebExceptionStatus.Timeout,
                  WebExceptionStatus.RequestCanceled }.
            Contains(exception.Status))                   
                    .OrResult(msg => msg.StatusCode == System.Net.HttpStatusCode.NotFound)
                    .WaitAndRetryAsync(3, retryAttempt => TimeSpan.FromSeconds(Math.Pow(500, retryAttempt)));
        }

        private IAsyncPolicy<HttpResponseMessage> GetFallbackPolicy()
        {            
            return HttpPolicyExtensions
                    .HandleTransientHttpError().Or<Exception>()
                    .FallbackAsync<HttpResponseMessage>(fallbackAction: (context) =>
                    {
                        return new Task<HttpResponseMessage>(() => new HttpResponseMessage() { StatusCode = HttpStatusCode.InternalServerError });
                    }, onFallbackAsync: (exception) =>
                    {                        
                        throw exception.Exception;
                    });
        }

        private IAsyncPolicy<HttpResponseMessage> GetTimeOutPolicy()
        {
            return Policy.TimeoutAsync(30, TimeoutStrategy.Optimistic).AsAsyncPolicy<HttpResponseMessage>();           
        }
```		

[...]

PR #1697, opened 2023-10-13
Update to cancel downstream operation in TimeoutStrategy.Pessimistic

lor1mp:
This PR has changes to extend the life of the `combinedTokenSource` in the `AsyncTimeoutEngine`, so that the cancellation will get propegated downstream in the case of a `TimeoutRejectedException`.

# Pull Request

We noticed that the downstream cancellation token was never signaled in `TimeoutStrategy.Pessimistic`. The root cause seems to be the `TimeoutRejectionException` escaping the scope of the `ImplementationAsync` function on `AsyncTimeoutEngine.cs:5`. This causes the `CancellationTokenSource` to be disposed and the downstream token to never get canceled. 

This is related to this issue: https://github.com/App-vNext/Polly/issues/722

In this fix we explicitly cancel the `combinedTokenSource` as the function leaves scope, because without the exception will propagate and cause the source to be disposed before the downstream token can be canceled. This issue was demonstrated in a UT. However, that UT requires that we use "real time" as the `TimeSpecBase.cs` `SystemClock` overrides actually cancel the downstream token themselves on a timeout. 

## Confirm the following

### DIFF
--- a/src/Polly/Timeout/AsyncTimeoutEngine.cs
+++ b/src/Polly/Timeout/AsyncTimeoutEngine.cs
@@ -17,26 +17,17 @@ internal static class AsyncTimeoutEngine
         using var timeoutCancellationTokenSource = new CancellationTokenSource();
         // Do not use a using here, the exception will exit the scope before we have time to
         // notify the downstream of the cancellation.
-        var combinedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, timeoutCancellationTokenSource.Token);
+        using var combinedTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, timeoutCancellationTokenSource.Token);
 
         Task<TResult> actionTask = null;
         CancellationToken combinedToken = combinedTokenSource.Token;
 
         try
         {
-
-            var actionTaskFn = async () =>
-            {
-                var result = await action(context, combinedToken).ConfigureAwait(continueOnCapturedContext);
-                // dispose of the token source after we've waited on the task to finish.
-                combinedTokenSource.Dispose();
-                return result;
-            };
-
             if (timeoutStrategy == TimeoutStrategy.Optimistic)
             {
                 SystemClock.CancelTokenAfter(timeoutCancellationTokenSource, timeout);
-                return await actionTaskFn().ConfigureAwait(continueOnCapturedContext);
+                return await action(context, combinedToken).ConfigureAwait(continueOnCapturedContext);
             }
 
             // else: timeoutStrategy == TimeoutStrategy.Pessimistic
@@ -45,10 +36,9 @@ internal static class AsyncTimeoutEngine
 
             SystemClock.CancelTokenAfter(timeoutCancellationTokenSource, timeout);
 
-            actionTask = actionTaskFn();
+            actionTask = action(context, combinedToken);
 
             return await (await Task.WhenAny(actionTask, timeoutTask).ConfigureAwait(continueOnCapturedContext)).ConfigureAwait(continueOnCapturedContext);
-
         }
         catch (Exception ex)
         {
@@ -62,6 +52,15 @@ internal static class AsyncTimeoutEngine
 
             throw;
         }
+        finally
+        {
+            // If the timeoutCancellation was canceled & our combined token hasn't been signaled, signal it.
+            // This avoids the exception propagating before the linked token can signal the downstream to cancel.
+            if (!combinedTokenSource.IsCancellationRequested && timeoutCancellationTokenSource.IsCancellationRequested)
+            {
+                combinedTokenSource.Cancel();
+            }
+        }
     }
 
     private static Task<TResult> AsTask<TResult>(this CancellationToken cancellationToken)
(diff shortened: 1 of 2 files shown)


test/Polly.Specs/Timeout/TimeoutAsyncSpecs.cs
```diff
@@ -248,6 +250,31 @@ public async Task Should_rethrow_exception_from_inside_delegate__pessimistic()
...
+    public async Task Should_cancel_downstream_token_on_timeout__pessimistic()
+    {
+        // It seems that there's a difference in the mocked clock vs. the real time clock.
+        // This test does not fail when using the mocked timer.
+        // In the TimeoutSpecsBase we actually cancel the combined token. Which hides
+        // the fact that it doesn't actually cancel irl.
+        SystemClock.Reset();
+
+        var policy = Policy.TimeoutAsync(TimeSpan.FromMilliseconds(200), TimeoutStrategy.Pessimistic);
+        var upstreamToken = new CancellationToken();
```
martincostello, line 263: Use `CancellationToken.None`? I think the intent (i.e. it won't actually cancel) is clearer then.

test/Polly.Specs/Timeout/TimeoutAsyncSpecs.cs
```diff
@@ -248,6 +250,31 @@ public async Task Should_rethrow_exception_from_inside_delegate__pessimistic()
...
+
+        var act = () => policy.ExecuteAsync(async (combinedToken) =>
+        {
+            combinedToken.Register(() => isCancelled = true);
+            await SystemClock.SleepAsync(TimeSpan.FromMilliseconds(300), combinedToken);
+        }, upstreamToken);
+
+        await act.Should().ThrowAsync<TimeoutRejectedException>();
+
+        await SystemClock.SleepAsync(TimeSpan.FromMilliseconds(400), default);
```
martincostello, line 274: Maybe make these 300/400ms sleeps longer so there's less chance of a slow CI system causing the test to pass unexpectedly or otherwise introduce flakiness?

src/Polly/Timeout/AsyncTimeoutEngine.cs
```diff
@@ -15,17 +15,28 @@ internal static class AsyncTimeoutEngine
...
         try
         {
+
+            var actionTaskFn = async () =>
+            {
+                var result = await action(context, combinedToken).ConfigureAwait(continueOnCapturedContext);
+                // dispose of the token source after we've waited on the task to finish.
+                combinedTokenSource.Dispose();
+                return result;
+            };
```
martincostello, line 34: My concern here is that, unless I'm misunderstanding, this creates scenarios where the pessimistic timeout doesn't happen and an exception is thrown (i.e. any other type of exception) so the disposal is then left to the finalizer.

I can see [in the dotnet/runtime source code](https://github.com/dotnet/runtime/blob/4322fc9f5732fbbe0d95727cc75b439e66196906/src/libraries/System.Private.CoreLib/src/System/Threading/CancellationTokenSource.cs#L884-L973) that things won't fire if the token source is disposed of more than once, but the registrations should still file when the scope exits and it's disposed of?

martincostello: Maybe just explicitly cancelling it via the source if it wasn't already in the `finally` block would resolve it?

lor1mp: @martincostello -- I will `try` that and report back.

lor1mp: That seems to fix it! Nice one, thanks!

src/Polly/Timeout/AsyncTimeoutEngine.cs
```diff
@@ -15,6 +15,8 @@ internal static class AsyncTimeoutEngine
         TimeSpan timeout = timeoutProvider(context);
 
         using var timeoutCancellationTokenSource = new CancellationTokenSource();
+        // Do not use a using here, the exception will exit the scope before we have time to
+        // notify the downstream of the cancellation.
```
martincostello, line 19: Comment is no longer relevant since you changed it back.

```suggestion
```

src/Polly/Timeout/AsyncTimeoutEngine.cs
```diff
@@ -51,6 +52,15 @@ internal static class AsyncTimeoutEngine
 
             throw;
         }
+        finally
+        {
+            // If the timeoutCancellation was canceled & our combined token hasn't been signaled, signal it.
+            // This avoids the exception propagating before the linked token can signal the downstream to cancel.
```
martincostello, line 58:
```suggestion
            // If timeoutCancellationTokenSource was canceled & our combined token hasn't been signaled, cancel it.
            // This avoids the exception propagating before the linked token can signal the downstream to cancel.
            // See https://github.com/App-vNext/Polly/issues/722.
```

lor1mp:
@martincostello - sorry this is my first PR into Polly. I'm not sure how to go about covering this line? (This is the line it's telling me I'm missing right?) Is it saying I don't have the false condition of the branch, and it's just way off?
image

test/Polly.Specs/Timeout/TimeoutAsyncSpecs.cs
martincostello: I think having just one test that's slow because it needs to be to reproduce the issue is fine. A second should be fine as then it's 5x longer than the shortest one.

martincostello:
Don't worry about the coverage comment, it's more advisory as long as the PR status checks are green.

From looking at it, I _think_ it's impossible to hit anyway.

test/Polly.Specs/Timeout/TimeoutAsyncSpecs.cs
lor1mp: Okay I made the sleep inside 1000ms, and left the cancellation 200 ms. I removed the last wait, because it was unnecessary and could actually hide a race condition. I had it there to prove however long you waited it would never succeed. But I think now that it's fixed this way, to avoid regression we actually don't want it.

martincostello (approved):

martincostello:
No, it's just being especially flaky today and I'm retrying it until it passes 😅

PR #1698, opened 2023-10-16
[Docs] Fallback after retries

### LINKED ISSUE #1694: Pipeline order of Retry + Fallback in polly 8
# Bug Report

## Summary
I'm upgrading from polly 7 to 8 and follow the guide over at https://www.pollydocs.org/migration-v8.html

Specifically, the section about `Policy.Wrap` which says:

> IMPORTANT
> In v7, the policy wrap ordering is different; the policy added first was executed last (FILO). In v8, the execution order matches the 
> order in which they were added (FIFO).

I had a retry with fallback policy in 7.x that looked like

```
var retryPolicy = Policy<MyResult>.Handle<...>().RetryAsync(...);
var fallbackPolicy = Policy<MyResult>.Handle<...>().FallbackAsync(...);
// in 7 they are executed LIFO
var policy = Policy.Wrap(fallback, retry)
```

Running the above on v8 works fine.

My attempt to use the new `ResiliencePipelineBuilder` to achieve the same looks something like this:

```
// In v8 the are supposedly executed in FIFO order
var pipeline = new ResiliencePipelineBuilder<MyResult>()
     .AddRetry(....)
     .AddFallback(....)
    .Build();
```

Executing with the new pipeline always immediately goes to the fallback. Retry is never called

Interestingly, swapping the order in the pipeline builder does in fact produce the correct result. 

## Steps / Code to reproduce the problem

The attached sample sets up a simple retry + fallback pipeline and shows how the retry is never called unless it is the last one added to the pipeline

This is the output I get on Polly 8.0.0

```
Starting to execute pipeline 1
Fallback! Outcome: This will never work.
Result after pipeline 1: False
Starting to execute pipeline 2
OnRetry attempt 0. Outcome: This will never work
OnRetry attempt 1. Outcome: This will never work
Fallback! Outcome: This will never work.
Result after pipeline 2: False
```

```cs
public static class PollyTest
{
    public static void TestRetryFallbackAsync()
    {
        var nonWorkingPipeline = new ResiliencePipelineBuilder<bool>()
                       .AddRetry(GetRetryStrategyOptions())
                       .AddFallback(GetFallbackStrategyOptions())
                       .Build();

        Debug.WriteLine("Starting to execute pipeline 1");
        bool result = nonWorkingPipeline.Execute(() => TryButFail());
        Debug.WriteLine("Result after pipeline 1: {0}", result);
        
        
        var workingPipeline = new ResiliencePipelineBuilder<bool>()
                       .AddFallback(GetFallbackStrategyOptions())
                       .AddRetry(GetRetryStrategyOptions())
                       .Build();

        Debug.WriteLine("Starting to execute pipeline 2");
        result = workingPipeline.Execute(() => TryButFail());
        Debug.WriteLine("Result after pipeline 2: {0}", result);
    }
    
    private static FallbackStrategyOptions<bool> GetFallbackStrategyOptions()
    {
        return new FallbackStrategyOptions<bool>()
        {
            ShouldHandle = new PredicateBuilder<bool>().Handle<ApplicationException>(),
            OnFallback = args =>
            {
                Debug.WriteLine("Fallback! Outcome: {0}.", args.Outcome);
                return default;
            },
            FallbackAction = args => Outcome.FromResultAsValueTask<bool>(false),
        };
    }

    private static RetryStrategyOptions<bool> GetRetryStrategyOptions()
    {
        return new RetryStrategyOptions<bool>()
        {
            ShouldHandle = new PredicateBuilder<bool>().Handle<ApplicationException>(),
            MaxRetryAttempts = 2,
            Delay = TimeSpan.Zero,
            OnRetry = args =>
            {
                Debug.WriteLine("OnRetry attempt {0}. Outcome: {1}", args.AttemptNumber, args.Outcome);
                return default;
            },
        };
    }
    
    private static bool TryButFail()
    {
        throw new ApplicationException("This will never work");
    }
}
```

martintmk:

- Improve fallback docs with a new "fallback after retries" pattern.
- Split "Patterns and anti-patterns" into a separate sections where it makes sense.

Closes #1694

## Confirm the following


src/Snippets/Docs/Fallback.cs
```diff
@@ -232,4 +232,40 @@ private static ValueTask<HttpResponseMessage> ActionCore()
...
+                ShouldHandle = new PredicateBuilder<HttpResponseMessage>()
+                    .Handle<HttpRequestException>()
+                    .HandleResult(r => r.StatusCode == HttpStatusCode.InternalServerError),
+                FallbackAction = args =>
+                {
+                    // Try to resolve the fallback response
+                    HttpResponseMessage fallbackResponse = ResolveFallbackResponse(args.Outcome);
+
+                    return Outcome.FromResultAsValueTask(fallbackResponse);
+
```
martincostello, line 252:
```suggestion
```

docs/strategies/fallback.md
```diff
@@ -65,7 +65,50 @@ new ResiliencePipelineBuilder<UserAvatar>()
...
+        MaxRetryAttempts = 3,
+    })
+    .Build();
+
+// Execution for demonstration purposes that always returns an invalid result
+pipeline.Execute(() => new HttpResponseMessage(HttpStatusCode.InternalServerError));
+```
+<!-- endSnippet -->
+
+Here's a breakdown of the behavior when the callback produces either an `HttpStatusCode.InternalServerError` or a `HttpRequestException`:
```
martincostello, line 104:
```suggestion
Here's a breakdown of the behavior when the callback produces either an `HttpStatusCode.InternalServerError` or an `HttpRequestException`:
```

martincostello (approved):

docs/strategies/fallback.md
```diff
@@ -65,7 +65,50 @@ new ResiliencePipelineBuilder<UserAvatar>()
 | `FallbackAction` | `Null`, **Required**                                                       | Fallback action to be executed.                                                             |
 | `OnFallback`     | `null`                                                                     | Event that is raised when fallback happens.                                                 |
 
-## Patterns and anti-patterns
+## Patterns
+
+### Fallback after retries
+
+When designing resilient systems, a common pattern is to use a fallback after multiple retry attempts. This approach is especially relevant when a fallback strategy can provide a sensible default value.
```
peter-csala, line 72: > after multiple retry attempts

after multiple _failed_ retry attempts

docs/strategies/fallback.md
```diff
@@ -65,7 +65,50 @@ new ResiliencePipelineBuilder<UserAvatar>()
...
+
+<!-- snippet: fallback-after-retries -->
+```cs
+var pipeline = new ResiliencePipelineBuilder<HttpResponseMessage>()
+    .AddFallback(new()
+    {
+        ShouldHandle = new PredicateBuilder<HttpResponseMessage>()
```
peter-csala, line 79: In order to empathize that  the fallback would trigger after the retry attempts exhausted we could reuse the `PredicateBuilder` between the two strategies.

docs/strategies/retry.md
```diff
@@ -104,7 +104,44 @@ new ResiliencePipelineBuilder().AddRetry(new RetryStrategyOptions
...
+
+<!-- snippet: retry-pattern-max-delay -->
+```cs
+ResiliencePipeline pipeline = new ResiliencePipelineBuilder()
+    .AddRetry(new()
+    {
+        Delay = TimeSpan.FromSeconds(2),
+        MaxRetryAttempts = int.MaxValue,
+
+        // Initially, we aim for an exponential backoff, but after a certain number of retries, we set a maximum delay of 15 minutes.
```
peter-csala, line 123: Would it make sense to explicitly use `BackoffType = DelayBackoffType.Exponential` ?

docs/strategies/fallback.md
isak1080: I like this suggestion. It was something I was considering doing when I started migrating to v8 but didn't dare to do because I wasn't sure it was allowed (thinking maybe the builder would have some state or whatever..). Having it here would remove such doubts.

And as suggested, it makes it clear to the reader of the code that they handle the same problem.

peter-csala: The only "state" of the `PredicateBuilder` is the [list of predicates](https://github.com/App-vNext/Polly/blob/main/src/Polly.Core/PredicateBuilder.TResult.cs#L9).

martintmk: >In order to empathize that the fallback would trigger after the retry attempts exhausted we could reuse the PredicateBuilder between the two strategies.
 
Good suggestion! Done.

peter-csala (approved):

PR #1706, opened 2023-10-19
[Docs] Add sequence diagrams to hedging

peter-csala:
# Pull Request

- Add sequence diagrams to the documentation to ease the understanding of the strategies

- Latency mode
  - [Happy path](https://github.com/peter-csala/Polly/blob/add-diagrams-part-4/docs/strategies/hedging.md#latency-happy-path-sequence-diagram)
  - [Unhappy path](https://github.com/peter-csala/Polly/blob/add-diagrams-part-4/docs/strategies/hedging.md#latency-unhappy-path-sequence-diagram)
- Fallback mode
  - [Happy path](https://github.com/peter-csala/Polly/blob/add-diagrams-part-4/docs/strategies/hedging.md#fallback-happy-path-sequence-diagram)
  - [Unhappy path](https://github.com/peter-csala/Polly/blob/add-diagrams-part-4/docs/strategies/hedging.md#fallback-unhappy-path-sequence-diagram)
- Parallel mode
  - [Happy path](https://github.com/peter-csala/Polly/blob/add-diagrams-part-4/docs/strategies/hedging.md#parallel-happy-path-sequence-diagram)
  - [Unhappy path](https://github.com/peter-csala/Polly/blob/add-diagrams-part-4/docs/strategies/hedging.md#parallel-unhappy-path-sequence-diagram)
- Dynamic mode
  - [Happy path](https://github.com/peter-csala/Polly/blob/add-diagrams-part-4/docs/strategies/hedging.md#dynamic-happy-path-sequence-diagram)
  - [Unhappy path](https://github.com/peter-csala/Polly/blob/add-diagrams-part-4/docs/strategies/hedging.md#dynamic-unhappy-path-sequence-diagram)
  
## Confirm the following

### DIFF
--- a/docs/strategies/hedging.md
+++ b/docs/strategies/hedging.md
@@ -87,6 +87,68 @@ When the `Delay` property is set to a value greater than zero, the hedging strat
 - The final result is the result of fastest successful execution.
 - If all executions fail, the final result will be the first failure encountered.
 
+#### Latency: happy path sequence diagram
+
+The hedging strategy does not trigger because the response arrives faster than the threshold.
+
+```mermaid
+sequenceDiagram
+    autonumber
+    actor C as Caller
+    participant P as Pipeline
+    participant H as Hedging
+    participant HUC as HedgedUserCallback
+
+    C->>P: Calls ExecuteAsync
+    P->>H: Calls ExecuteCore
+    Note over H: Wait start
+
+    activate H
+    H->>+HUC: Invokes
+    HUC-->>HUC: Processes request
+    HUC->>-H: Returns result
+    deactivate H
+
+    Note over H: Wait end
+    H->>P: Returns result
+    P->>C: Returns result
+```
+
+#### Latency: unhappy path sequence diagram
+
+The hedging strategy triggers because the response arrives slower than the threshold.
+
+```mermaid
+sequenceDiagram
+    autonumber
+    actor C as Caller
+    participant P as Pipeline
+    participant H as Hedging
+    participant D as HedgedUserCallback
+
+    C->>P: Calls ExecuteAsync
+    P->>H: Calls ExecuteCore
+    Note over H: Wait start
+
+    activate H
+    activate D
+    H->>D: Invokes (R1)
+    D-->>D: Processes R1<br/> slowly ...
+    Note over H: Wait end
+
+    H->>D: Invokes (R2)
+    activate D
+    D-->>D: Processes R2<br/> quickly
+    D->>H: Returns result (R2)
+    deactivate D
+    H->>D: Propagates cancellation (R1)
+    deactivate H
+
+    deactivate D
+    H->>P: Returns result
+    P->>C: Returns result
+```
+
 ### Fallback mode
 
 In fallback mode, the `Delay` value should be less than `TimeSpan.Zero`. This mode allows only a single execution to proceed at a given time.
@@ -95,6 +157,64 @@ In fallback mode, the `Delay` value should be less than `TimeSpan.Zero`. This mo
 - If the initial execution fails, new one is initiated.
 - The final result will be the first successful execution.
 - If all executions fail, the final result will be the first failure encountered.
+
+#### Fallback: happy path sequence diagram
+
+The hedging strategy triggers because the first attempt fails. It succeeds because the retry attempts do not exceed the `MaxHedgedAttempts`.
+
+```mermaid
+sequenceDiagram
+    autonumber
+    actor C as Caller
+    participant P as Pipeline
+    participant H as Hedging
+    participant HUC as HedgedUserCallback
+
+    C->>P: Calls ExecuteAsync
+    P->>H: Calls ExecuteCore
+    activate H
+
+    H->>+HUC: Invokes (R1)
+    HUC-->>HUC: Processes R1
+    HUC->>-H: Fails (R1)
+
+    H->>+HUC: Invokes (R2)
+    HUC-->>HUC: Processes R2
+    HUC->>-H: Returns result (R2)
+
+    deactivate H
+    H->>P: Returns result (R2)
+    P->>C: Returns result (R2)
+```
+
+#### Fallback: unhappy path sequence diagram
+
+The hedging strategy triggers because the first attempt fails. It fails because all retry attempts failed as well.
+
+```mermaid
+sequenceDiagram
+    autonumber
+    actor C as Caller
+    participant P as Pipeline
+    participant H as Hedging
+    participant HUC as HedgedUserCallback
+
+    C->>P: Calls ExecuteAsync
+    P->>H: Calls ExecuteCore
+    activate H
+
+    H->>+HUC: Invokes (R1)
+    HUC-->>HUC: Processes R1
+    HUC->>-H: Fails (R1)
+
+    H->>+HUC: Invokes (R2)
+    HUC-->>HUC: Processes R2
+    HUC->>-H: Fails (R2)
+
+    deactivate H
+    H->>P: Propagates failure (R1)
+    P->>C: Propagates failure
+```
 
 ### Parallel mode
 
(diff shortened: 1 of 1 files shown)


peter-csala:
@martincostello , @martintmk 

Several questions:
- What do you guys think, would it make sense to add sequence diagram to each concurrency mode to visualize the differences?
- Are the current diagrams too detailed? I mean what do you think adding `LoadBalancer`, `ServiceReplica1`, and `ServiceRelica2` ease the understanding or complicates the diagram unnecessarily?
- In case of unhappy path I'm unsure about the following: will the hedging receive the slow response as well? Or it won't as I depicted on the diagram.

martintmk:
> @martincostello , @martintmk
> 
> Several questions:
> 
> * What do you guys think, would it make sense to add sequence diagram to each concurrency mode to visualize the differences?
> * Are the current diagrams too detailed? I mean what do you think adding `LoadBalancer`, `ServiceReplica1`, and `ServiceRelica2` eases the understanding or complicates the diagram unnecessarily?
> * In case of unhappy path I'm unsure about the following: will the hedging receive the slow response as well? Or it won't as I depicted on the diagram.
> * And last but not least: would it make sense to use here the `autonumber` feature (like in the [circuit breaker cases](https://www.pollydocs.org/strategies/circuit-breaker.html#simple-happy-path-sequence-diagram))?

Thanks for these! 

> What do you guys think, would it make sense to add sequence diagram to each concurrency mode to visualize the differences?

I would say yes, each concurrency mode changes behavior of hedging and it would be better to explain these. 

> * Are the current diagrams too detailed? I mean what do you think adding `LoadBalancer`, `ServiceReplica1`, and `ServiceRelica2` eases the understanding or complicates the diagram unnecessarily?

My initial thinking is that those are tied to HTTP scenarios, maybe we can get away of just having `DecoratedUserCallback` and `HedgedUserCallback`?

Esentially, merge those 3 under `HedgedUserCallback`?

> In case of unhappy path I'm unsure about the following: will the hedging receive the slow response as well? Or it won't as I depicted on the diagram.

Not really, once the acceptable result is received, the hedging cancels all pending requests, wait until the cancellation is finished and then returns the accepted response.

> > * And last but not least: would it make sense to use here the `autonumber` feature (like in the [circuit breaker cases](https://www.pollydocs.org/strategies/circuit-breaker.html#simple-happy-path-sequence-diagram))?

Sure, looks nice to me.

martintmk:
Maybe also include the 'ActionGenerator' as an additional actor that produces 'HedgedActionCallback'?

peter-csala:
> > * Are the current diagrams too detailed? I mean what do you think adding `LoadBalancer`, `ServiceReplica1`, and `ServiceRelica2` eases the understanding or complicates the diagram unnecessarily?
> 
> My initial thinking is that those are tied to HTTP scenarios, maybe we can get away of just having `DecoratedUserCallback` and `HedgedUserCallback`?
> 
> Esentially, merge those 3 under `HedgedUserCallback`?
> 

The [The Tail at Scale](https://www.barroso.org/publications/TheTailAtScale.pdf)) paper introduced "hedged requests" like this:
> Hedged requests. A simple way to curb latency variability is to issue the same request to multiple replicas and
use the results from whichever replica responds first. We term such requests “hedged requests” because a client first sends one request to the replica believed to be the most appropriate, but then falls back on sending a secondary request after some brief delay. The client cancels remaining outstanding requests once the first result is received.

Accordingly to my understanding it is tightly coupled to HTTP and tail-tolerance. In other implementations, what I have seen, you can define the wait dynamically via percentiles (like 95p). So, I think here the primary usage involves HTTP. (but maybe I'm wrong with this assumption).

docs/strategies/hedging.md
```diff
@@ -73,6 +73,80 @@ You can use the following special values for `Delay` or in `DelayGenerator`:
...
+    S2->>S2: Processes request
+    S1-->>S1: Still processing ...
+    S2->>-D: Returns result (R2)
+    D->>H: Returns result (R2)
+    H->>S1: Propagates Cancellation
+    deactivate S1
+    deactivate H
+
+    H->>P: Returns result
+    P->>C: Returns result
```
martincostello, line 147: For clarity:

```suggestion
    H->>P: Returns result (R2)
    P->>C: Returns result (R2)
```

martincostello (commented):
I think these HTTP-specific examples are fine, but that should be pointed out in the preamble to the diagrams that the illustrations are based on that kind of scenario.

peter-csala:
@martincostello , @martintmk 

To be able to compare the two versions (with HTTP and without HTTP) I've created alternative diagrams:
- [Fast response scenario](https://github.com/peter-csala/Polly/blob/add-diagrams-part-4/docs/strategies/hedging.md#alternative-happy-path-sequence-diagram)
- [Slow response scenario](https://github.com/peter-csala/Polly/blob/add-diagrams-part-4/docs/strategies/hedging.md#alternative-unhappy-path-sequence-diagram)

martintmk:
> @martincostello , @martintmk
> 
> To be able to compare the two versions (with HTTP and without HTTP) I've created alternative diagrams:
> 
> * [Fast response scenario](https://github.com/peter-csala/Polly/blob/add-diagrams-part-4/docs/strategies/hedging.md#alternative-happy-path-sequence-diagram)
> * [Slow response scenario](https://github.com/peter-csala/Polly/blob/add-diagrams-part-4/docs/strategies/hedging.md#alternative-unhappy-path-sequence-diagram)

I also prefer the second one, mainly because it explains the universal nature of hedging strategy. You are right that it is primary for HTTP scenarios, but it can be applied to anything where latency is priority.

docs/strategies/hedging.md
```diff
@@ -73,6 +73,133 @@ You can use the following special values for `Delay` or in `DelayGenerator`:
...
+
+### Alternative: Unhappy path sequence diagram
+
+```mermaid
+sequenceDiagram
+    autonumber
+    actor C as Caller
+    participant P as Pipeline
+    participant H as Hedging
+    participant D as HedgedUserCallback
```
martintmk, line 181: participant C as UserCallback
participant G as ActionGenerator

- The first request executes UserCallback directly
- Second and all other create `HedgedUserCallback` from `ActionGenerator`

peter-csala: I think omitting the `ActionGenerator` from the majority of the diagrams eases the understanding.

We could add a single diagram to the *Action Generator* which shows the usage of the generator.

Can you help me by creating that diagram?

docs/strategies/hedging.md
```diff
@@ -96,6 +158,64 @@ In fallback mode, the `Delay` value should be less than `TimeSpan.Zero`. This mo
...
+    HUC-->>HUC: Processes R1
+    HUC->>-H: Fails (R1)
+
+    H->>+HUC: Invokes (R2)
+    HUC-->>HUC: Processes R2
+    HUC->>-H: Fails (R2)
+
+    deactivate H
+    H->>P: Propagates failure (R1)
+    P->>C: Propagates failure
```
martincostello, line 216:
```suggestion
    P->>C: Propagates failure (R1)
```

docs/strategies/hedging.md
```diff
@@ -87,6 +87,68 @@ When the `Delay` property is set to a value greater than zero, the hedging strat
...
+    activate D
+    D-->>D: Processes R2<br/> quickly
+    D->>H: Returns result (R2)
+    deactivate D
+    H->>D: Propagates cancellation (R1)
+    deactivate H
+
+    deactivate D
+    H->>P: Returns result
+    P->>C: Returns result
```
martincostello, line 149:
```suggestion
    H->>P: Returns result (R2)
    P->>C: Returns result (R2)
```

docs/strategies/hedging.md
```diff
@@ -147,6 +337,102 @@ With this configuration, the hedging strategy:
...
+    participant DG as DelayGenerator
+    participant HUC as HedgedUserCallback
+
+    C->>P: Calls ExecuteAsync
+    P->>H: Calls ExecuteCore
+    activate H
+
+    Note over H: Parallel mode
+    par
+    H -> DG: Gets delay
```
martincostello, line 358: We should have all these "Gets delay" lines be like the others with an arrow each way and with the delay value being returned, rather than just a single horizontal line.

docs/strategies/hedging.md
```diff
@@ -147,6 +337,110 @@ With this configuration, the hedging strategy:
 - Initiates a maximum of `4` executions. This includes initial action and an additional 3 attempts.
 - Allows the first two executions to proceed in parallel, while the third and fourth executions follow the fallback mode.
```
peter-csala, line 338: @martintmk At the _fallback mode_ section you have stated:
> In fallback mode, the `Delay` value should be less than `TimeSpan.Zero`.

And here we are using `InfiniteTimeSpan`. I'm not sure that at glance it is obvious why `ystem.Threading.Timeout.InfiniteTimeSpan < TimeSpan.Zero` is true.

I think it might make sense to add some explanation here. Do you agree?

martincostello (approved):

docs/strategies/hedging.md
martintmk: Agree, the behavior will be the same whenever you provide `System.Threading.Timeout.InfiniteTimeSpan` or any value lesser than 0.

Maybe System.Threading.Timeout.InfiniteTimeSpan wasn't the good choice as it suggests infinite waiting, i.e. fallback.

docs/strategies/hedging.md
peter-csala: Okay, I will add that diagram into the next PR.

docs/strategies/hedging.md
peter-csala: In the next PR I will replace `System.Threading.Timeout.InfiniteTimeSpan` to `TimeSpan.FromSeconds(-1)`

martintmk (approved):

martincostello:
@peter-csala You might want to update your local git config for the polly repo so your Git identity matches your GitHub identity - otherwise your commits all come in as this white circle that appears to be a different person:

image

PR #1707, opened 2023-10-19
Add link to retries blog

martincostello:
Add a link to a really good blog post that interactively demonstrates how retries work with exponential back-off and jitter.


martincostello:
> Can you also add an section to a retry strategy with this info?

That's how I started, but there's no real prose explaining what timeouts are in that document so it looked out of place. I don't have the time to do that today, so I can do that in a follow-up?

martintmk (approved):

PR #1728, opened 2023-10-25
[Docs] Fix the policywrap sample

peter-csala:
# Pull Request

- #1718 

- V7's sample code created a policy chain where the timeout was the outer and the retry was the inner
- Fixed the V7's sample code to have the same behavior as V8's sample
- Also extended the comment to further clarify which one is the outer and which one is the inner

## Confirm the following

### DIFF
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -157,9 +157,9 @@ IAsyncPolicy retryPolicy = Policy.Handle<Exception>().WaitAndRetryAsync(3, _ => 
 IAsyncPolicy timeoutPolicy = Policy.TimeoutAsync(TimeSpan.FromSeconds(3));
 
 // Wrap the policies. The policies are executed in the following order (i.e. Last-In-First-Out):
-// 1. Retry
-// 2. Timeout
-IAsyncPolicy wrappedPolicy = Policy.WrapAsync(timeoutPolicy, retryPolicy);
+// 1. Retry << outer
+// 2. Timeout << inner
+IAsyncPolicy wrappedPolicy = Policy.WrapAsync(retryPolicy, timeoutPolicy);
 ```
 <!-- endSnippet -->
 
@@ -170,8 +170,8 @@ In v8, there's no need to use policy wrap explicitly. Instead, policy wrapping i
 <!-- snippet: migration-policy-wrap-v8 -->
 ```cs
 // The "PolicyWrap" is integrated directly. Strategies are executed in the same order as they were added (i.e. First-In-First-Out):
-// 1. Retry
-// 2. Timeout
+// 1. Retry << outer
+// 2. Timeout << outer
 ResiliencePipeline pipeline = new ResiliencePipelineBuilder()
     .AddRetry(new()
     {
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -157,8 +157,8 @@ IAsyncPolicy retryPolicy = Policy.Handle<Exception>().WaitAndRetryAsync(3, _ => 
 IAsyncPolicy timeoutPolicy = Policy.TimeoutAsync(TimeSpan.FromSeconds(3));
 
 // Wrap the policies. The policies are executed in the following order (i.e. Last-In-First-Out):
-// 1. Retry << outer
-// 2. Timeout << inner
+// 1. Retry <== outer
+// 2. Timeout <== inner
 IAsyncPolicy wrappedPolicy = Policy.WrapAsync(retryPolicy, timeoutPolicy);
 ```
 <!-- endSnippet -->
@@ -170,8 +170,8 @@ In v8, there's no need to use policy wrap explicitly. Instead, policy wrapping i
 <!-- snippet: migration-policy-wrap-v8 -->
 ```cs
 // The "PolicyWrap" is integrated directly. Strategies are executed in the same order as they were added (i.e. First-In-First-Out):
-// 1. Retry << outer
-// 2. Timeout << outer
+// 1. Retry <== outer
+// 2. Timeout <== outer
 ResiliencePipeline pipeline = new ResiliencePipelineBuilder()
     .AddRetry(new()
     {
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -156,7 +156,7 @@ IAsyncPolicy retryPolicy = Policy.Handle<Exception>().WaitAndRetryAsync(3, _ => 
 
 IAsyncPolicy timeoutPolicy = Policy.TimeoutAsync(TimeSpan.FromSeconds(3));
 
-// Wrap the policies. The policies are executed in the following order (i.e. Last-In-First-Out):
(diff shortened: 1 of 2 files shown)


docs/migration-v8.md
```diff
@@ -157,9 +157,9 @@ IAsyncPolicy retryPolicy = Policy.Handle<Exception>().WaitAndRetryAsync(3, _ =>
 IAsyncPolicy timeoutPolicy = Policy.TimeoutAsync(TimeSpan.FromSeconds(3));
 
 // Wrap the policies. The policies are executed in the following order (i.e. Last-In-First-Out):
-// 1. Retry
-// 2. Timeout
-IAsyncPolicy wrappedPolicy = Policy.WrapAsync(timeoutPolicy, retryPolicy);
+// 1. Retry << outer
+// 2. Timeout << inner
```
martincostello, line 161: It might be misunderstood as "much less than", so maybe this is better because (at least to me) it looks like an arrow.

```suggestion
// 1. Retry <== outer
// 2. Timeout <== inner
```

docs/migration-v8.md
```diff
@@ -170,8 +170,8 @@ In v8, there's no need to use policy wrap explicitly. Instead, policy wrapping i
 <!-- snippet: migration-policy-wrap-v8 -->
 ```cs
 // The "PolicyWrap" is integrated directly. Strategies are executed in the same order as they were added (i.e. First-In-First-Out):
-// 1. Retry
-// 2. Timeout
+// 1. Retry << outer
+// 2. Timeout << outer
```
martincostello, line 174:
```suggestion
// 1. Retry <== outer
// 2. Timeout <== outer
```

martincostello (approved):

peter-csala:
@martintmk In case of V8 the FIFO statement is incorrect IMHO. The first added strategy will be executed last. 

So, first you define the outer most and last the inner most.

Which is the same order as in case of V7 (first parameter is the outer most, last parameter is the inner most).

src/Snippets/Docs/Migration.Wrap.cs
```diff
@@ -11,9 +11,9 @@ public static void PolicyWrap_V7()
         IAsyncPolicy timeoutPolicy = Policy.TimeoutAsync(TimeSpan.FromSeconds(3));
 
         // Wrap the policies. The policies are executed in the following order (i.e. Last-In-First-Out):
-        // 1. Retry
-        // 2. Timeout
-        IAsyncPolicy wrappedPolicy = Policy.WrapAsync(timeoutPolicy, retryPolicy);
+        // 1. Retry <== outer
+        // 2. Timeout <== inner
+        IAsyncPolicy wrappedPolicy = Policy.WrapAsync(retryPolicy, timeoutPolicy);
```
martintmk, line 16: Is this true though? Isn't timeout policy executed first here?

peter-csala: https://dotnetfiddle.net/znmcPA

peter-csala: You can see the order in case of V7 and V8.

martintmk: true, i got it wrong it my head then. I think we should also update/drop this note then:

peter-csala: I will update the PR later today or on Monday

martintmk (approved):

martintmk:
> @martintmk In case of V8 the FIFO statement is **incorrect** IMHO. The firstly added strategy will be executed last.
> 
> So, first you define the outer most and last the inner most.
> 
> Which is the same order as in case of V7 (first parameter is the outer most, last parameter is the inner most).
> 
> If you agree then I will update this section to reflect the current behavior.
> 
> Dotnet fiddle: https://dotnetfiddle.net/znmcPA

They problem is relying on the callback events, what's important is the when retry strategy starts executing and when it ends.  Yes, it's true that the timeout event is reported first, but at that point the retry strategy is already executing.

If you put custom strategy that just reports events before and after the callback is executed, it would become clear.

peter-csala:
> They problem is relying on the callback events, what's important is the when retry strategy starts executing and when it ends. Yes, it's true that the timeout event is reported first, but at that point the retry strategy is already executing.
> 
> If you put custom strategy that just reports events before and after the callback is executed, it would become clear.

Sorry @martintmk , but I'm a bit lost. Could you please rephrase your statement? (I'm not sure I understand it 100%).

martintmk:
> > They problem is relying on the callback events, what's important is the when retry strategy starts executing and when it ends. Yes, it's true that the timeout event is reported first, but at that point the retry strategy is already executing.
> > If you put custom strategy that just reports events before and after the callback is executed, it would become clear.
> 
> Sorry @martintmk , but I'm a bit lost. Could you please rephrase your statement? (I'm not sure I understand it 100%).

I believe this test conveys what I am trying to say. It adds multiple strategies to the pipeline and asserts the order of execution:

https://github.com/App-vNext/Polly/blob/41d4e8dbd7947717970822e5e3095c70d80a2d23/test/Polly.Core.Tests/ResiliencePipelineBuilderTests.cs#L64

peter-csala:
> ResiliencePipelineBuilderTests.cs

I understand that 
- in case of `Before` the order is **from outer to inner** and 
- in case of `After` the order is _from outer to inner_. 

([sequence diagram](https://mermaid.live/edit#pako:eNptkk1rxCAQQP-KzHlz6Dd4ENq0uRUCMTcvUzO7K5uY1BhoWfa_V-tKoKkHYeY9dYbxDHrsCDjM9LmQ1fRq8OBwUJaFNaHzRpsJrWc1w5nVZqLeWNriKuLKuNlvWRNZQ3q03RbKCOXRuH9YW0bYzuRK7PsP1Cdlk1UXQlScpaAqUvRC-9ERZzc5LUSTnaZIUXZuc1oImR0ZHLkqd_kxGaS25OztizRn9ymZ9rb8e0GMnveeHGcP6_FNHVflcS1j085VecplxH7qKCkLOxjIDWi6MLpzxAr8kQZSEATo0J0UKHsJHi5-bL6tBu7dQjtYpg59HjPwPfZzyFJn_Oje01_4_RKXH1pzn50))
___
I think I get it what do you want to describe. 

Forgot the `onRetry` and `onTimeout` delegates and let's focus on this example:
```cs
var retryPolicy = Policy.Handle<Exception>().WaitAndRetryAsync(2, _ => TimeSpan.FromSeconds(0.5));
var timeoutPolicy = Policy.TimeoutAsync(TimeSpan.FromSeconds(1));

var wrappedPolicy = Policy.WrapAsync(retryPolicy, timeoutPolicy);

Console.WriteLine("Outer: " + wrappedPolicy.Outer.GetType().Name);
Console.WriteLine("Inner: " + wrappedPolicy.Inner.GetType().Name);
```
and the output:
```none
Outer: AsyncRetryPolicy
Inner: AsyncTimeoutPolicy
```
are you convinced ?

Issue #1732, opened 2023-10-27
[Feature request]: AOT compatibility


davidfowl:
### Is your feature request related to a specific problem? Or an existing feature?

Polly is at the core of [Microsoft.Extensions.Http.Resilience](https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.Http.Resilience) and we'd like the libraries that depend on this package and polly itself to work in AOT environments.

### Describe the solution you'd like

Add `<IsAotCompatible>true</IsAotCompatible>` to the project files and work through the AOT issues. I've run a project that has a dependency and saw this:

```
  ILC: Method '[Polly.Core]Polly.Utils.Pipeline.CompositeComponent+DelegatingComponent.ExecuteCore<__Canon,ValueTuple`3<__Canon,__Canon,ValueTuple`3<__Canon,__Canon,ValueTuple`3<__Canon,__Canon,ValueTuple`3<__Canon,__Canon,ValueTuple`2<__Canon,__Canon>>>>>>(Func`3<Resi
  lienceContext,ValueTuple`3<__Canon,__Canon,ValueTuple`3<__Canon,__Canon,ValueTuple`3<__Canon,__Canon,ValueTuple`3<__Canon,__Canon,ValueTuple`2<__Canon,__Canon>>>>>,ValueTask`1<Outcome`1<__Canon>>>,ResilienceContext,ValueTuple`3<__Canon,__Canon,ValueTuple`3<__Canon,__
  Canon,ValueTuple`3<__Canon,__Canon,ValueTuple`3<__Canon,__Canon,ValueTuple`2<__Canon,__Canon>>>>>)' will always throw because: Failed to load type 'Polly.Utils.Pipeline.PipelineComponent' from assembly 'Polly.Core, Version=8.0.0.0, Culture=neutral, PublicKeyToken=c8a
  3ffc3f8f825cc'
ILC : AOT analysis warning IL3054: Polly.Utils.Pipeline.CompositeComponent.DelegatingComponent.ExecuteCore<__Canon,ValueTuple`3<__Canon,__Canon,ValueTuple`3<__Canon,__Canon,ValueTuple`3<__Canon,__Canon,ValueTuple`3<__Canon,__Canon,ValueTuple`2<__Canon,__Canon>>>>>>(Fun
[...]

PR #1737, opened 2023-10-27
Resolve AOT compilation issues

martincostello:
A _potential_ fix for #1732 using `RuntimeFeature.IsDynamicCodeSupported` to guard against an allocation regression while still avoiding the infinite generic recursion the current code in Polly 8.0.0 causes.

This is just a starting point for discussion on what the best fix and/or compromises are to support using Polly with AoT compiled applications.

I'm very much out of my comfort zone when it comes to AoT, so this was just something I came up with that resolves the issue at hand. It does trade-off against either increased code size by using the guard, or without the guard it introduces an allocation regression of 24 bytes per invocation ([original thread](https://github.com/martincostello/Polly/commit/393b2dbf422825db6c1ee0dd24c4ddc42d31f05c#r131079631)).

💯% open to suggestions for a better solution.

/cc @davidfowl @MichalStrehovsky @eerhardt @martintmk

### DIFF
--- a/src/Polly.Core/Utils/Pipeline/CompositeComponent.cs
+++ b/src/Polly.Core/Utils/Pipeline/CompositeComponent.cs
@@ -1,4 +1,6 @@
-﻿using Polly.Telemetry;
+﻿using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using Polly.Telemetry;
 
 namespace Polly.Utils.Pipeline;
 
@@ -120,7 +122,7 @@ internal sealed class CompositeComponent : PipelineComponent
     /// <summary>
     /// A component that delegates the execution to the next component in the chain.
     /// </summary>
-    private sealed class DelegatingComponent : PipelineComponent
+    internal sealed class DelegatingComponent : PipelineComponent
     {
         private readonly PipelineComponent _component;
 
@@ -130,7 +132,49 @@ internal sealed class CompositeComponent : PipelineComponent
 
         public override ValueTask DisposeAsync() => default;
 
+        [ExcludeFromCodeCoverage]
         internal override ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState>(
+            Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
+            ResilienceContext context,
+            TState state)
+        {
+#if NET6_0_OR_GREATER
+            return RuntimeFeature.IsDynamicCodeSupported ? ExecuteComponent(callback, context, state) : ExecuteComponentAot(callback, context, state);
+#else
+            return ExecuteComponent(callback, context, state);
+#endif
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static ValueTask<Outcome<TResult>> ExecuteNext<TResult, TState>(
+            PipelineComponent next,
+            Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
+            ResilienceContext context,
+            TState state)
+        {
+            if (context.CancellationToken.IsCancellationRequested)
+            {
+                return Outcome.FromExceptionAsValueTask<TResult>(new OperationCanceledException(context.CancellationToken).TrySetStackTrace());
+            }
+
+            return next.ExecuteCore(callback, context, state);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal ValueTask<Outcome<TResult>> ExecuteComponent<TResult, TState>(
+            Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
+            ResilienceContext context,
+            TState state)
+        {
+            return _component.ExecuteCore(
+                static (context, state) => ExecuteNext(state.Next!, state.callback, context, state.state),
+                context,
+                (Next, callback, state));
+        }
+
+#if NET6_0_OR_GREATER
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal ValueTask<Outcome<TResult>> ExecuteComponentAot<TResult, TState>(
             Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
             ResilienceContext context,
             TState state)
@@ -143,12 +187,7 @@ internal sealed class CompositeComponent : PipelineComponent
                 {
                     var callback = (Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>>)wrapper.Callback;
                     var state = (TState)wrapper.State;
-                    if (context.CancellationToken.IsCancellationRequested)
-                    {
-                        return Outcome.FromExceptionAsValueTask<TResult>(new OperationCanceledException(context.CancellationToken).TrySetStackTrace());
-                    }
-
-                    return wrapper.Next.ExecuteCore(callback, context, state);
+                    return ExecuteNext(wrapper.Next, callback, context, state);
                 },
                 context,
                 new StateWrapper(Next!, callback, state!));
@@ -167,5 +206,6 @@ internal sealed class CompositeComponent : PipelineComponent
             public object Callback;
             public object State;
         }
+#endif
     }
 }
--- a/src/Polly.Core/Utils/Pipeline/CompositeComponent.cs
+++ b/src/Polly.Core/Utils/Pipeline/CompositeComponent.cs
@@ -1,6 +1,4 @@
-﻿using System.Diagnostics.CodeAnalysis;
-using System.Runtime.CompilerServices;
-using Polly.Telemetry;
+﻿using Polly.Telemetry;
 
 namespace Polly.Utils.Pipeline;
 
@@ -117,95 +115,5 @@ internal sealed class CompositeComponent : PipelineComponent
             new PipelineExecutedArguments(_timeProvider.GetElapsedTime(timeStamp)));
 
         return outcome;
-    }
-
-    /// <summary>
-    /// A component that delegates the execution to the next component in the chain.
-    /// </summary>
-    internal sealed class DelegatingComponent : PipelineComponent
-    {
-        private readonly PipelineComponent _component;
-
-        public DelegatingComponent(PipelineComponent component) => _component = component;
-
-        public PipelineComponent? Next { get; set; }
-
-        public override ValueTask DisposeAsync() => default;
-
-        [ExcludeFromCodeCoverage]
-        internal override ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState>(
-            Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
-            ResilienceContext context,
-            TState state)
-        {
-#if NET6_0_OR_GREATER
-            return RuntimeFeature.IsDynamicCodeSupported ? ExecuteComponent(callback, context, state) : ExecuteComponentAot(callback, context, state);
-#else
-            return ExecuteComponent(callback, context, state);
-#endif
-        }
-
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private static ValueTask<Outcome<TResult>> ExecuteNext<TResult, TState>(
-            PipelineComponent next,
-            Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
-            ResilienceContext context,
-            TState state)
-        {
-            if (context.CancellationToken.IsCancellationRequested)
-            {
-                return Outcome.FromExceptionAsValueTask<TResult>(new OperationCanceledException(context.CancellationToken).TrySetStackTrace());
-            }
-
-            return next.ExecuteCore(callback, context, state);
-        }
-
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal ValueTask<Outcome<TResult>> ExecuteComponent<TResult, TState>(
-            Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
-            ResilienceContext context,
-            TState state)
-        {
-            return _component.ExecuteCore(
-                static (context, state) => ExecuteNext(state.Next!, state.callback, context, state.state),
-                context,
-                (Next, callback, state));
-        }
-
-#if NET6_0_OR_GREATER
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal ValueTask<Outcome<TResult>> ExecuteComponentAot<TResult, TState>(
-            Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
-            ResilienceContext context,
-            TState state)
-        {
-            // Custom state object is used to cast the callback and state to prevent infinite
-            // generic type recursion warning IL3054 when referenced in a native AoT application.
-            // See https://github.com/App-vNext/Polly/issues/1732 for further context.
-            return _component.ExecuteCore(
-                static (context, wrapper) =>
-                {
-                    var callback = (Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>>)wrapper.Callback;
-                    var state = (TState)wrapper.State;
-                    return ExecuteNext(wrapper.Next, callback, context, state);
-                },
-                context,
-                new StateWrapper(Next!, callback, state!));
-        }
-
-        private struct StateWrapper
-        {
-            public StateWrapper(PipelineComponent next, object callback, object state)
-            {
-                Next = next;
-                Callback = callback;
-                State = state;
-            }
-
-            public PipelineComponent Next;
-            public object Callback;
-            public object State;
-        }
-#endif
     }
 }
--- a/src/Polly.Core/Utils/Pipeline/CompositeComponent.cs
+++ b/src/Polly.Core/Utils/Pipeline/CompositeComponent.cs
@@ -1,4 +1,5 @@
-﻿using Polly.Telemetry;
+﻿using System.Runtime.CompilerServices;
+using Polly.Telemetry;
 
 namespace Polly.Utils.Pipeline;
 
@@ -128,7 +129,22 @@ internal sealed class CompositeComponent : PipelineComponent
 
         public PipelineComponent? Next { get; set; }
 
+        public override ValueTask DisposeAsync() => default;
+
         internal override ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState>(
+            Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
+            ResilienceContext context,
+            TState state)
+        {
+#if NET6_0_OR_GREATER
+            return RuntimeFeature.IsDynamicCodeSupported ? ExecuteCoreImpl(callback, context, state) : ExecuteCoreAot(callback, context, state);
+#else
+            return ExecuteCoreImpl(callback, context, state);
+#endif
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private ValueTask<Outcome<TResult>> ExecuteCoreImpl<TResult, TState>(
             Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
             ResilienceContext context,
             TState state)
@@ -147,6 +163,45 @@ internal sealed class CompositeComponent : PipelineComponent
                 (Next, callback, state));
         }
 
-        public override ValueTask DisposeAsync() => default;
+#if NET6_0_OR_GREATER
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private ValueTask<Outcome<TResult>> ExecuteCoreAot<TResult, TState>(
+            Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
+            ResilienceContext context,
+            TState state)
+        {
+            // Custom state object is used to cast the callback and state to prevent infinite
+            // generic type recursion warning IL3054 when referenced in a native AoT application.
+            // See https://github.com/App-vNext/Polly/issues/1732 for further context.
+            return _component.ExecuteCore(
+                static (context, wrapper) =>
+                {
+                    var callback = (Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>>)wrapper.Callback;
+                    var state = (TState)wrapper.State;
+                    if (context.CancellationToken.IsCancellationRequested)
+                    {
+                        return Outcome.FromExceptionAsValueTask<TResult>(new OperationCanceledException(context.CancellationToken).TrySetStackTrace());
+                    }
+
+                    return wrapper.Next.ExecuteCore(callback, context, state);
+                },
+                context,
+                new StateWrapper(Next!, callback, state!));
+        }
+
+        private struct StateWrapper
+        {
+            public StateWrapper(PipelineComponent next, object callback, object state)
+            {
+                Next = next;
+                Callback = callback;
+                State = state;
+            }
+
+            public PipelineComponent Next;
+            public object Callback;
+            public object State;
+        }
+#endif
     }
 }
--- a/src/Polly.Core/Utils/Pipeline/CompositeComponent.cs
+++ b/src/Polly.Core/Utils/Pipeline/CompositeComponent.cs
@@ -1,5 +1,4 @@
-﻿using System.Runtime.CompilerServices;
-using Polly.Telemetry;
+﻿using Polly.Telemetry;
 
 namespace Polly.Utils.Pipeline;
 
@@ -136,40 +135,6 @@ internal sealed class CompositeComponent : PipelineComponent
             ResilienceContext context,
             TState state)
         {
-#if NET6_0_OR_GREATER
-            return RuntimeFeature.IsDynamicCodeSupported ? ExecuteCoreImpl(callback, context, state) : ExecuteCoreAot(callback, context, state);
-#else
-            return ExecuteCoreImpl(callback, context, state);
-#endif
-        }
-
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private ValueTask<Outcome<TResult>> ExecuteCoreImpl<TResult, TState>(
-            Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
-            ResilienceContext context,
-            TState state)
-        {
-            return _component.ExecuteCore(
-                static (context, state) =>
-                {
-                    if (context.CancellationToken.IsCancellationRequested)
-                    {
-                        return Outcome.FromExceptionAsValueTask<TResult>(new OperationCanceledException(context.CancellationToken).TrySetStackTrace());
-                    }
-
-                    return state.Next!.ExecuteCore(state.callback, context, state.state);
-                },
-                context,
-                (Next, callback, state));
-        }
-
-#if NET6_0_OR_GREATER
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private ValueTask<Outcome<TResult>> ExecuteCoreAot<TResult, TState>(
-            Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
-            ResilienceContext context,
-            TState state)
-        {
             // Custom state object is used to cast the callback and state to prevent infinite
             // generic type recursion warning IL3054 when referenced in a native AoT application.
             // See https://github.com/App-vNext/Polly/issues/1732 for further context.
@@ -202,6 +167,5 @@ internal sealed class CompositeComponent : PipelineComponent
             public object Callback;
             public object State;
         }
-#endif
     }
 }
--- a/src/Polly.Core/Utils/Pipeline/DelegatingComponent.cs
+++ b/src/Polly.Core/Utils/Pipeline/DelegatingComponent.cs
@@ -1,0 +1,94 @@
+﻿using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+
+namespace Polly.Utils.Pipeline;
+
+/// <summary>
+/// A component that delegates the execution to the next component in the chain.
+/// </summary>
+internal sealed class DelegatingComponent : PipelineComponent
+{
+    private readonly PipelineComponent _component;
+
+    public DelegatingComponent(PipelineComponent component) => _component = component;
+
+    public PipelineComponent? Next { get; set; }
+
+    public override ValueTask DisposeAsync() => default;
+
+    [ExcludeFromCodeCoverage]
+    internal override ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState>(
+        Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
+        ResilienceContext context,
+        TState state)
+    {
+#if NET6_0_OR_GREATER
+        return RuntimeFeature.IsDynamicCodeSupported ? ExecuteComponent(callback, context, state) : ExecuteComponentAot(callback, context, state);
+#else
+        return ExecuteComponent(callback, context, state);
+#endif
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private static ValueTask<Outcome<TResult>> ExecuteNext<TResult, TState>(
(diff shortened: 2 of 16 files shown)


MichalStrehovsky:
I'll just re-state what I wrote in https://github.com/martincostello/Polly/commit/393b2dbf422825db6c1ee0dd24c4ddc42d31f05c#r131086685. The allocation either matters or it doesn't.

If it doesn't matter, this looks very similar to the problem that Rx had that got simply fixed like this: https://github.com/dotnet/reactive/pull/1116. This could be a similar fix (just drop the `static` keyword from the lambda and pick up whatever you need from higher-up scope; let the C# compiler create the closure class for you).

If the allocation matters, it's going to matter for PublishAot too. The profiling/etc. workflows are different between native and non-native deployments. Having drastically different (assuming the allocation matters here) perf characteristics is not great. This will be hard to troubleshoot. Getting rid of the allocation would likely require shaking up more than just `DelegatingComponent`.

My gut feel is that the allocation doesn't matter. Whenever `DelegatingComponent` is used, the CPU time is going to be dominated by dispatching 3 generic virtual methods. It is the slowest possible method call in .NET (by orders of magnitude). The cost of the allocation is dwarfed by that. It adds a bit of GC pressure (might need more gen0 collections), but the work GC will have to do is likely still going to be smaller than what we just burned on the generic virtual method calls. (I.e. if this is perf sensitive, there shouldn't be any generic virtual method calls, and definitely not 3 of them).

martincostello:
Thanks for the information.

A core part of the original design was to make things allocation free, which is why I was trying to keep it to zero. However it sounds like achieving that while resolving the AOT issue isn't going to be possible without significant refactoring.

It might be the case that we just accept that this code path needs to regress slightly (and we do that simple thing of dropping the `static`) in order to support AOT, but I'll defer to @martintmk for an opinion on that.

I might have a bit of a further play around today to see if there's a different approach to the higher-level of what the code is doing that would indirectly solve the issue.

As I mentioned in the original issue, do you have a sample project you can point me to for testing AoT compatibility here going forwards?

I could pretty much only repro the issue by putting the **exact** code into an AoT app. We could do with something that's more realistic and representative of user code, but still uncovers the original problem.

davidfowl:
cc @agocke

martincostello:
Some results for comparison on my laptop based on the benchmark in #1743.

The Mean value seems quite noisy, so I'd just focus on the Allocated column.

```
BenchmarkDotNet v0.13.9+228a464e8be6c580ad9408e98f18813f6407fb5a, Windows 11 (10.0.22621.2428/22H2/2022Update/SunValley2)
12th Gen Intel Core i7-1270P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 7.0.403
  [Host] : .NET 7.0.13 (7.0.1323.51816), X64 RyuJIT AVX2
Job=MediumRun  Toolchain=InProcessEmitToolchain  IterationCount=15  
LaunchCount=2  WarmupCount=10  
```

### Polly 8.0.0

| Method                         | Mean     | Error    | StdDev   | Allocated |
|------------------------------- |---------:|---------:|---------:|----------:|
| CompositeComponent_ExecuteCore | 44.37 ns | 1.994 ns | 2.923 ns |         - |

### With boxing via wrapper

| Method                         | Mean     | Error    | StdDev   | Gen0   | Allocated |
|------------------------------- |---------:|---------:|---------:|-------:|----------:|
| CompositeComponent_ExecuteCore | 37.87 ns | 0.650 ns | 0.972 ns | 0.0025 |      24 B |

```csharp
internal override ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState>(
    Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
    ResilienceContext context,
    TState state)
{
    return _component.ExecuteCore(
        static (context, wrapper) =>
        {
            var callback = (Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>>)wrapper.Callback;
            var state = (TState)wrapper.State;
            if (context.CancellationToken.IsCancellationRequested)
            {
                return Outcome.FromExceptionAsValueTask<TResult>(new OperationCanceledException(context.CancellationToken).TrySetStackTrace());
            }

            return wrapper.Next.ExecuteCore(callback, context, state);
        },
        context,
        new StateWrapper(Next!, callback, state!));
}

public override ValueTask DisposeAsync() => default;

private readonly struct StateWrapper
{
    public StateWrapper(PipelineComponent next, object callback, object state)
    {
        Next = next;
        Callback = callback;
        State = state;
    }

    public readonly PipelineComponent Next;
    public readonly object Callback;
    public readonly object State;
}
```

### Removing `static`

| Method                         | Mean     | Error    | StdDev   | Gen0   | Allocated |
|------------------------------- |---------:|---------:|---------:|-------:|----------:|
| CompositeComponent_ExecuteCore | 54.23 ns | 1.147 ns | 1.570 ns | 0.0101 |      96 B |

```csharp
internal override ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState>(
    Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
    ResilienceContext context,
    TState state)
{
    return _component.ExecuteCore(
        (context, state) =>
        {
            if (context.CancellationToken.IsCancellationRequested)
            {
                return Outcome.FromExceptionAsValueTask<TResult>(new OperationCanceledException(context.CancellationToken).TrySetStackTrace());
            }

            return Next!.ExecuteCore(callback, context, state);
        },
        context,
        state);
}
```

MichalStrehovsky:
> The Mean value seems quite noisy, so I'd just focus on the Allocated column.

Is there a real-world application that could be tested with? Microbenchmarks are a good supportive evidence, but they fail to capture other aspects e.g. what proportion is the 24 bytes compared to allocations in callbacks that do actual/real work? How does this change the time we spent JITting and loading new types (this reduces AOT compilation, but also less JITting outside AOT)? It's possible this change is simply a noise level regression in allocations and noise level improvement in startup time/working set with JIT. And an improvement for AOT.

> As I mentioned in the original issue, do you have a sample project you can point me to for testing AoT compatibility here going forwards?

Testing for AOT compatibility is done by the analyzers. The fact they don't find generic cycles is an analyzer bug. I filed it as https://github.com/dotnet/runtime/issues/94131. I thought we had a bug but I couldn't find it. Generic cycles are so rare that this hasn't been a huge priority in the past (the compiler even used to crash on them - we knew of the 3-4 libraries that had problems with it in the past, but not much else got ever reported). So Polly is just lucky.

One shouldn't really need to do anything else besides enabling analyzers and ensuring all external dependencies run with analyzers.

The low tech way to test AOT compatibility outside the analyzer (that one really shouldn't need because the analyzers are the testing story) is:

1. Create a dummy console project (`dotnet new console --aot -o dummy`). It doesn't matter what's in Main. It can be hello world.
2. Add a ProjectReference to your library
3. Add `<ItemGroup><TrimmerRootAssembly Include="NameOfYourLibrary" /></ItemGroup>`
4. Enable warnings as errors
5. `dotnet publish` the project. If the publish success, everything should be fine.

Step 3 ensures the compiler compiles everything in `NameOfYourLibrary` assembly. If it sees an issue in it, it will report a warning that gets upgraded to an error and fails the build.

martincostello:
> Is there a real-world application that could be tested with?

@davidfowl Are you (or someone you were reporting on behalf of) able to run this change through whatever app you found the issue with originally to see if we can get some better data on the net gain or loss with this change if it resolves the AOT issue?

> Testing for AOT compatibility is done by the analyzers. The fact they don't find generic cycles is an analyzer bug. I filed it as https://github.com/dotnet/runtime/issues/94131.

@MichalStrehovsky OK cool, I didn't realise that was an issue (or that it was a very rare type of problem) - I misunderstood it to be "by design" that it wasn't detected by the analysers, so that's why I thought we needed a proper full repro project of our own to detect the issue.

I'll put together a console app like you've described to check that it detects the original issue - if it does, then we can make that part of our CI until the analysis bug is fixed.

agocke:
The analyzers would ideally catch everything, but there are a few cases where it's very hard for them to see through all the possible situations. In those cases, running the AOT compiler is the best way to catch the problems. Luckily, as Michal detailed, it's pretty easy to create a simple console app for this. The instructions

> 1. Create a dummy console project (dotnet new console --aot -o dummy). It doesn't matter what's in Main. It can be hello world.
> 2. Add a ProjectReference to your library
> 3. Add <ItemGroup><TrimmerRootAssembly Include="NameOfYourLibrary" /></ItemGroup>
> 4. Enable warnings as errors
> 5. dotnet publish the project. If the publish success, everything should be fine.

Sum it up pretty well. This is worth a docs page though. https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?pivots=dotnet-8-0 is our page for trimming, but we should have one for Native AOT too.

davidfowl:
@martincostello 

> @davidfowl Are you (or someone you were reporting on behalf of) able to run this change through whatever app you found the issue with originally to see if we can get some better data on the net gain or loss with this change if it resolves the AOT issue?

This was found using the new resilience libraries in a branch of eShop. I was trying to AOT one of the services and several warnings showed up. Do you want me to verify the breakage or performance (I think this scenario causes breakage)?

@agocke 

>  This is worth a docs page though. https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?pivots=dotnet-8-0 is our page for trimming, but we should have one for Native AOT too.

Are you going to file a docs issue?

martincostello:
@davidfowl The perf impact of this fix if you could please. I opened #1745 earlier which repros the bug you found - if I do some rebasing later and incorporate it into this PR, it should pass to resolve the AoT issue.

martincostello:
(I also need to re-run our own benchmarks)

martintmk:
> Some results for comparison on my laptop based on the benchmark in #1743.
> 
> The Mean value seems quite noisy, so I'd just focus on the Allocated column.
> 
> ```
> BenchmarkDotNet v0.13.9+228a464e8be6c580ad9408e98f18813f6407fb5a, Windows 11 (10.0.22621.2428/22H2/2022Update/SunValley2)
> 12th Gen Intel Core i7-1270P, 1 CPU, 16 logical and 12 physical cores
> .NET SDK 7.0.403
>   [Host] : .NET 7.0.13 (7.0.1323.51816), X64 RyuJIT AVX2
> Job=MediumRun  Toolchain=InProcessEmitToolchain  IterationCount=15  
> LaunchCount=2  WarmupCount=10  
> ```
> 
> ### Polly 8.0.0
> Method	Mean	Error	StdDev	Allocated
> CompositeComponent_ExecuteCore	44.37 ns	1.994 ns	2.923 ns	-
> ### With boxing via wrapper
> Method	Mean	Error	StdDev	Gen0	Allocated
> CompositeComponent_ExecuteCore	37.87 ns	0.650 ns	0.972 ns	0.0025	24 B
> ```cs
> internal override ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState>(
>     Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
>     ResilienceContext context,
>     TState state)
> {
>     return _component.ExecuteCore(
>         static (context, wrapper) =>
>         {
>             var callback = (Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>>)wrapper.Callback;
>             var state = (TState)wrapper.State;
>             if (context.CancellationToken.IsCancellationRequested)
>             {
>                 return Outcome.FromExceptionAsValueTask<TResult>(new OperationCanceledException(context.CancellationToken).TrySetStackTrace());
>             }
> 
>             return wrapper.Next.ExecuteCore(callback, context, state);
>         },
>         context,
>         new StateWrapper(Next!, callback, state!));
> }
> 
> public override ValueTask DisposeAsync() => default;
> 
> private readonly struct StateWrapper
> {
>     public StateWrapper(PipelineComponent next, object callback, object state)
>     {
>         Next = next;
>         Callback = callback;
>         State = state;
>     }
> 
>     public readonly PipelineComponent Next;
>     public readonly object Callback;
>     public readonly object State;
> }
> ```
> 
> ### Removing `static`
> Method	Mean	Error	StdDev	Gen0	Allocated
> CompositeComponent_ExecuteCore	54.23 ns	1.147 ns	1.570 ns	0.0101	96 B
> ```cs
> internal override ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState>(
>     Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
>     ResilienceContext context,
>     TState state)
> {
>     return _component.ExecuteCore(
>         (context, state) =>
>         {
>             if (context.CancellationToken.IsCancellationRequested)
>             {
>                 return Outcome.FromExceptionAsValueTask<TResult>(new OperationCanceledException(context.CancellationToken).TrySetStackTrace());
>             }
> 
>             return Next!.ExecuteCore(callback, context, state);
>         },
>         context,
>         state);
> }
> ```

Based on these benchmarks there is a clear regression for both execution time and allocations when static is removed from lambda. Additionally, if we went ahead with this, we no longer can claim to have zero-allocation API. 

I would say, let's keep both branches and just have an extra allocation with AOT. In the future we can try to remove it too. Don't know how though 🤨.

src/Polly.Core/Utils/Pipeline/CompositeComponent.cs
```diff
@@ -128,25 +128,44 @@ private sealed class DelegatingComponent : PipelineComponent
...
                     {
                         return Outcome.FromExceptionAsValueTask<TResult>(new OperationCanceledException(context.CancellationToken).TrySetStackTrace());
                     }
 
-                    return state.Next!.ExecuteCore(state.callback, context, state.state);
+                    return wrapper.Next.ExecuteCore(callback, context, state);
                 },
                 context,
-                (Next, callback, state));
+                new StateWrapper(Next!, callback, state!));
```
martintmk, line 154: Note: this allocation happens for every strategy in that pipeline.

So 5 strategies --> around 100 bytes allocated per execution.

martincostello:
Excluding a flash of inspiration or a significant refactor, the options so far appear to be:

1. Accept the closure penalty of 96 bytes per invocation for the simplest code and a consistent experience regardless of whether AoT is used or not.
1. Accept the boxing penalty of 24 bytes per invocation for a consistent experience regardless of whether AoT is used at the expense of some additional cycles to pass through the data we need and unbox it.
1. Use `RuntimeFeature.IsDynamicCodeSupported` to mitigate the impact of approach 2 to only when AoT is being used at the expense of additional code complexity.

Personally I'm erring to option 3.

I'm going to revert the code back to dd2fd54fdee06f682a0f61cea3eace9b905aaee4 locally and see what the benchmarks give me.

martintmk:
> Excluding a flash of inspiration or a significant refactor, the options so far appear to be:
> 
> 1. Accept the closure penalty of 96 bytes per invocation for the simplest code and a consistent experience regardless of whether AoT is used or not.
> 2. Accept the boxing penalty of 24 bytes per invocation for a consistent experience regardless of whether AoT is used at the expense of some additional cycles to pass through the data we need and unbox it.
> 3. Use `RuntimeFeature.IsDynamicCodeSupported` to mitigate the impact of approach 2 to only when AoT is being used at the expense of additional code complexity.
> 
> Personally I'm erring to option 3.
> 
> I'm going to revert the code back to [dd2fd54](https://github.com/App-vNext/Polly/commit/dd2fd54fdee06f682a0f61cea3eace9b905aaee4) locally and see what the benchmarks give me.

I prefer the option 3 too. While we introduce extra branch for AOT this change is limited to a very small % of code-base.

Note, you can do that check even when building the pipeline and create an instance and chain `AotDelegatingStrategy` instances instead.

martincostello:
I'm about to push up some changes I've made to re-introduce the branch (just waiting for the mutation tests), after that I'll try the approach with having two different delegating strategy types, as that should be a lot simpler 👍.

martincostello:
I've tried to split the code by type (rather than with a method) in three different approaches, but it seems to still hit the original issue as the AoT compiler tries to include the version with the generic recursion and then fails to compile.

martintmk:
> I've tried to split the code by type (rather than with a method) in three different approaches, but it seems to still hit the original issue as the AoT compiler tries to include the version with the generic recursion and then fails to compile.

Can't we just say "we know better" and suppress the warning then?

martincostello:
Unsure - it looked like an error rather than a warning coming from a single line, and to be honest all three of the things I tried looked more complicated than this or had more duplication.

I would suggest we leave it like this for now if this is the approach we want (branch, with allocation regression for AoT) and we can try and refactor it more later after shipping a patch release that fixes things for AoT.

agocke:
> 
> > This is worth a docs page though. https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming?pivots=dotnet-8-0 is our page for trimming, but we should have one for Native AOT too.
> 
> Are you going to file a docs issue?

https://github.com/dotnet/docs/issues/37788

MichalStrehovsky:
> Additionally, if we went ahead with this, we no longer can claim to have zero-allocation API.

If this API is documented as zero-allocation, it should be called out that it will not deliver on such promise with AOT.

> I would say, let's keep both branches and just have an extra allocation with AOT. In the future we can try to remove it too. Don't know how though

For this, I'll just go back to what I said about generic virtual methods earlier. If ExecuteCore wasn't a generic virtual method, but simply took an object and boxed/unboxed things as needed, the benchmark results would look different. The generic virtual method approach avoids the boxing/unboxing, but then burns all the saved time trying to figure out where to dispatch the method calls. What this PR does to the AOT version is "worst of both worlds" - we still do a generic virtual method call, but also box. Here's how "best of both worlds" looks like (either no boxing, but generic virtual, or boxing and no generic virtual):

| Method  | Mean     | Error    | StdDev   | Gen0   | Allocated |
|-------- |---------:|---------:|---------:|-------:|----------:|
| Generic | 21.17 ns | 0.066 ns | 0.059 ns |      - |         - |
| Object  | 18.64 ns | 0.253 ns | 0.212 ns | 0.0035 |      88 B |

The version that boxes comes ahead by 13% solely because we didn't burn CPU time dispatching generic virtual methods.

(Notice the standard deviation is higher for the Object version, likely due to counting the time in GCs, but the boxing version still comes ahead with plenty of buffer.)

This might not be applicable to this problem, or there are other constraints, but wanted to call this out because sometimes people pay too much attention to small short lived allocations and miss the whole picture. BDN doesn't do a good job of showing where the costs shifted (with generic virtual methods, the costs shift to dispatch time and extra JIT time (we need to JIT a new native method for each unique T) - this translates to more working set, and worse startup time). Small short-lived allocations are cheap to clean up.

<details>
<summary>The whole benchmark backing the table, in case I made a mistake somewhere (click to expand). But the gist is that a generic virtual method call has a comparable CPU cost to an object allocation on average</summary>

```csharp
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Jobs;
using BenchmarkDotNet.Running;

BenchmarkRunner.Run<TestPipelines>();

[MemoryDiagnoser]
[SimpleJob(RuntimeMoniker.Net80)]
public class TestPipelines
{
    PipelineComponent p = new DelegatingComponent(new IdentityComponent(), new IdentityComponent());

    [Benchmark]
    public int Generic() => p.ExecuteCore<int, int>(state => state + 1, 0);

    [Benchmark]
    public int Object() => (int)p.ExecuteCore(state => (object)((int)state + 1), (object)0);
}

abstract class PipelineComponent
{
    public abstract TResult ExecuteCore<TResult, TState>(Func<TState, TResult> callback, TState state);

    public abstract object ExecuteCore(Func<object, object> callback, object state);
}

class IdentityComponent : PipelineComponent
{
    public override TResult ExecuteCore<TResult, TState>(Func<TState, TResult> callback, TState state)
        => callback(state);

    public override object ExecuteCore(Func<object, object> callback, object state)
        => callback(state);
}

class DelegatingComponent : PipelineComponent
{
    PipelineComponent _component;
    PipelineComponent _next;

    public DelegatingComponent(PipelineComponent component, PipelineComponent next)
    {
        _component = component;
        _next = next;
    }

    public override TResult ExecuteCore<TResult, TState>(Func<TState, TResult> callback, TState state)
    {
        return _component.ExecuteCore(static (state) =>
        {
            return state._next.ExecuteCore(state.callback, state.state);
        },
        (_next, callback, state));
    }

    public override object ExecuteCore(Func<object, object> callback, object state)
        => _component.ExecuteCore(static (object state) =>
        {
            (PipelineComponent next, Func<object, object> callback, object state_) =
                ((PipelineComponent, Func<object, object>, object))state;

            return next.ExecuteCore(callback, state_);
        },
        (_next, callback, state));
}
```
</details>

martincostello:
The options I can see here so far are:

1. Accept Polly v8 doesn't currently work with AoT even though it was always the intention it would (i.e. do nothing);
2. Degrade _something_ to unblock AoT support in the short term (i.e. an 8.0.1 release this week) with a view to future improvement;
3. Abandon this change completely and pursue a larger internal refactor to try and do something else that resolves the remaining AoT issue and improves performance further by removing the virtual dispatch. I don't know what the timelines for this would be.

I think it's better to unblock the scenario for AoT (which is effectively a bug because we always wanted to support it: #1110) for .NET 8, which is option 2.

It's also worth noting that the performance of Polly v8 with the new API surface in general is much better than Polly v7, so any users moving to .NET 8 and the new resilience features from dotnet/extensions are going to see improvements with or without AoT, even if AoT performance for Polly specifically is slightly worse than without. In other words, a user adopting the new Polly v8 APIs and .NET 8 won't miss something they didn't yet have.

It's not ideal, but coming up with a perfect solution (if possible) isn't going to be quick, so if we want to enable Polly with AoT for .NET 8 it seems like we're going to have to make some sort of compromise.

> If this API is documented as zero-allocation, it should be called out that it will not deliver on such promise with AOT.

Searching the code for [_"zero"_](https://github.com/search?q=repo%3AApp-vNext%2FPolly%20zero&type=code) and [_"allocation"_](https://github.com/search?q=repo%3AApp-vNext%2FPolly%20allocation&type=code), all I can find is references to use trying to avoid/minimise/reduce them. I think the only explicit "documentation" of zero are [the benchmark results](https://github.com/App-vNext/Polly/tree/main/bench/BenchmarkDotNet.Artifacts/results) checked into the repo that happen to be zero. I don't think we ever had a cast-iron guarantee that the new Polly v8 API was allocation free in 100% of scenarios.

martintmk:
While supporting AOT is important we should also realize that majority of .NET users simply won't be using it in near feature. I am all for supporting AOT even if it comes with some extra allocation. We can do this while still supporting non-AOT scenarios with no perf or allocation degradation. 

Regarding virtual generic methods, at this point IMHO I would stick with current design. If I read the numbers correctly it would give us around ~3 ns improvements per invocation. This improvement will be dwarfed by the overhead and ceremony required by each resilience strategy (see the [perf of standard pipeline](https://github.com/App-vNext/Polly/blob/main/bench/BenchmarkDotNet.Artifacts/results/Polly.Core.Benchmarks.MultipleStrategiesBenchmark-report-github.md) of 5 strategies which is around ~2000ns per execution).

On top of that Polly will be used in the context of async IO operations and we can multiple all the numbers by 10x (at minimum).

So my vote is to merge this PR, resolve AOT issues for 8.0.0, and keep an eye for future improvements.

martintmk:
Anyway, thanks @MichalStrehovsky for all the context. I didn't realize that generic virtual call have such overhead.

src/Polly.Core/Utils/Pipeline/CompositeComponent.cs
```diff
@@ -120,33 +122,90 @@ public override async ValueTask DisposeAsync()
     /// <summary>
     /// A component that delegates the execution to the next component in the chain.
     /// </summary>
-    private sealed class DelegatingComponent : PipelineComponent
+    internal sealed class DelegatingComponent : PipelineComponent
```
martintmk, line 125: Extract this component to it's own file as its not so straightforward anymore?

src/Polly.Core/Utils/Pipeline/DelegatingComponent.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                var callback = (Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>>)wrapper.Callback;
+                var state = (TState)wrapper.State;
+                return ExecuteNext(wrapper.Next, callback, context, state);
+            },
+            context,
+            new StateWrapper(Next!, callback, state!));
+    }
+
+    private struct StateWrapper
```
martintmk, line 80:
```suggestion
    private readonly record struct StateWrapper(...);
```

nit

martincostello: I'll re-benchmark again, but I did have this locally and one point and it seemed to make it slower than without.

martintmk: No need then, it's just nitpicking.

martincostello: I still think there's noise in the benchmarks, but it came out faster anyway, so I've adopted it for that and also the terser code. These were the numbers I got:

### Before

```
BenchmarkDotNet v0.13.9+228a464e8be6c580ad9408e98f18813f6407fb5a, Windows 11 (10.0.22621.2428/22H2/2022Update/SunValley2)
12th Gen Intel Core i7-1270P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 7.0.403
  [Host] : .NET 7.0.13 (7.0.1323.51816), X64 RyuJIT AVX2

Job=MediumRun  Toolchain=InProcessEmitToolchain  IterationCount=15
LaunchCount=2  WarmupCount=10
```

| Method                         | Mean     | Error    | StdDev   | Allocated |
|------------------------------- |---------:|---------:|---------:|----------:|
| CompositeComponent_ExecuteCore | 47.53 ns | 1.541 ns | 2.210 ns |         - |

### With `readonly`

```
BenchmarkDotNet v0.13.9+228a464e8be6c580ad9408e98f18813f6407fb5a, Windows 11 (10.0.22621.2428/22H2/2022Update/SunValley2)
12th Gen Intel Core i7-1270P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 7.0.403
  [Host] : .NET 7.0.13 (7.0.1323.51816), X64 RyuJIT AVX2

Job=MediumRun  Toolchain=InProcessEmitToolchain  IterationCount=15
LaunchCount=2  WarmupCount=10
```

| Method                         | Mean     | Error    | StdDev   | Allocated |
|------------------------------- |---------:|---------:|---------:|----------:|
| CompositeComponent_ExecuteCore | 43.40 ns | 0.736 ns | 1.032 ns |         - |

### With `readonly record`

```
BenchmarkDotNet v0.13.9+228a464e8be6c580ad9408e98f18813f6407fb5a, Windows 11 (10.0.22621.2428/22H2/2022Update/SunValley2)
12th Gen Intel Core i7-1270P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 7.0.403
  [Host] : .NET 7.0.13 (7.0.1323.51816), X64 RyuJIT AVX2

Job=MediumRun  Toolchain=InProcessEmitToolchain  IterationCount=15
LaunchCount=2  WarmupCount=10
```

| Method                         | Mean     | Error    | StdDev   | Allocated |
|------------------------------- |---------:|---------:|---------:|----------:|
| CompositeComponent_ExecuteCore | 37.44 ns | 0.713 ns | 0.952 ns |         - |

martincostello: Just realised it must be noise as BDN shouldn't be executing that code path anyway...

src/Polly.Core/Utils/Pipeline/DelegatingComponent.cs
```diff
@@ -0,0 +1,82 @@
...
+namespace Polly.Utils.Pipeline;
+
+/// <summary>
+/// A component that delegates the execution to the next component in the chain.
+/// </summary>
+internal sealed class DelegatingComponent : PipelineComponent
+{
+    private readonly PipelineComponent _component;
+
+    public DelegatingComponent(PipelineComponent component) => _component = component;
```
martintmk, line 13:
```suggestion
    public DelegatingComponent(PipelineComponent component, bool aotSupported) => _component = component;
```

And pass it from outside? This would allow testing both paths.

src/Polly.Core/Utils/Pipeline/DelegatingComponent.cs
martincostello: AoT vs. not for the three ways:

### Before

```
BenchmarkDotNet v0.13.9+228a464e8be6c580ad9408e98f18813f6407fb5a, Windows 11 (10.0.22621.2428/22H2/2022Update/SunValley2)
12th Gen Intel Core i7-1270P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 7.0.403
  [Host] : .NET 7.0.13 (7.0.1323.51816), X64 RyuJIT AVX2

Job=MediumRun  Toolchain=InProcessEmitToolchain  IterationCount=15
LaunchCount=2  WarmupCount=10
```

| Method                              | Mean     | Error    | StdDev   | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|------------------------------------ |---------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
| DelegatingComponent_ExecuteCore_Jit | 33.14 ns | 2.851 ns | 3.996 ns |  1.00 |    0.00 |      - |         - |          NA |
| DelegatingComponent_ExecuteCore_Aot | 35.31 ns | 2.358 ns | 3.456 ns |  1.08 |    0.13 | 0.0025 |      24 B |          NA |

### With `readonly`

```
BenchmarkDotNet v0.13.9+228a464e8be6c580ad9408e98f18813f6407fb5a, Windows 11 (10.0.22621.2428/22H2/2022Update/SunValley2)
12th Gen Intel Core i7-1270P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 7.0.403
  [Host] : .NET 7.0.13 (7.0.1323.51816), X64 RyuJIT AVX2

Job=MediumRun  Toolchain=InProcessEmitToolchain  IterationCount=15
LaunchCount=2  WarmupCount=10
```

| Method                              | Mean     | Error    | StdDev   | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|------------------------------------ |---------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
| DelegatingComponent_ExecuteCore_Jit | 32.06 ns | 1.823 ns | 2.672 ns |  1.00 |    0.00 |      - |         - |          NA |
| DelegatingComponent_ExecuteCore_Aot | 35.52 ns | 0.482 ns | 0.721 ns |  1.12 |    0.09 | 0.0025 |      24 B |          NA |

### With `readonly record`

```
BenchmarkDotNet v0.13.9+228a464e8be6c580ad9408e98f18813f6407fb5a, Windows 11 (10.0.22621.2428/22H2/2022Update/SunValley2)
12th Gen Intel Core i7-1270P, 1 CPU, 16 logical and 12 physical cores
.NET SDK 7.0.403
  [Host] : .NET 7.0.13 (7.0.1323.51816), X64 RyuJIT AVX2

Job=MediumRun  Toolchain=InProcessEmitToolchain  IterationCount=15
LaunchCount=2  WarmupCount=10
```

| Method                              | Mean     | Error    | StdDev   | Ratio | RatioSD | Gen0   | Allocated | Alloc Ratio |
|------------------------------------ |---------:|---------:|---------:|------:|--------:|-------:|----------:|------------:|
| DelegatingComponent_ExecuteCore_Jit | 31.50 ns | 1.256 ns | 1.719 ns |  1.00 |    0.00 |      - |         - |          NA |
| DelegatingComponent_ExecuteCore_Aot | 35.22 ns | 1.678 ns | 2.511 ns |  1.14 |    0.09 | 0.0025 |      24 B |          NA |

src/Polly.Core/Utils/Pipeline/DelegatingComponent.cs
martincostello: I don't think it's worth it to remove the need to add `[ExcludeFromCodeCoverage]` for one branch of one line of code and then need to try and suppress the IL3054 warning.

martintmk: Ok, haven't realized this. Let's leave it as it is.

martintmk (approved):

PR #1738, opened 2023-10-27
Update to .NET 8 SDK

martincostello:
- Build with the .NET 8 SDK.
- Use new [artifacts output](https://devblogs.microsoft.com/dotnet/announcing-dotnet-8-preview-3/#simplified-output-path).
- Use C# 12 collection initializers where relevant. Suppress false positives when using `TheoryData`.
- Remove some redundant code analysis suppressions.

Separates the .NET 8 upgrade in #1144 into using .NET 8 (this PR) and adding `net8.0` support.


src/Polly/Policy.ExecuteOverloads.cs
```diff
@@ -10,7 +10,7 @@ public abstract partial class Policy : ISyncPolicy
     /// <param name="action">The action to perform.</param>
     [DebuggerStepThrough]
     public void Execute(Action action) =>
-        Execute((_, _) => action(), new Context(), DefaultCancellationToken);
+        Execute((_, _) => action(), [], DefaultCancellationToken);
```
martintmk, line 13: Is this right?  Why [] instead of `new Context()`?

martincostello: The SDK suggests using collection expressions, and in these cases it still works. I can double check that the underlying generated code is the same and it's not a false positive (like the TestData suppressions) but if it's semantically equivalent it's better than fighting the analyzer with suppressions.

martintmk: Just in case, doesn't this creates a new dictionary? I.e. change of behavior and call to a different overload.

martincostello: I'm not sure. I assumed it was some compiler smarts. I'll look at it in ILSpy next time I'm at a computer and see what it's actually producing.

martincostello: Looking at this in Visual Studio shows that it's calling this method:

https://github.com/App-vNext/Polly/blob/a98198f7fc9717babd2ca2f9bb58e65fcfa8c493/src/Polly/Policy.ExecuteOverloads.cs#L60

image

`Context` implements a number of dictionary interfaces, so the compiler is effectively using the collection expression syntax to create an empty `Context`:

https://github.com/App-vNext/Polly/blob/a98198f7fc9717babd2ca2f9bb58e65fcfa8c493/src/Polly/Context.Dictionary.cs#L7

Which agrees with what I can see in ILSpy:

image

martincostello:
Reported docfx failures as https://github.com/dotnet/sdk/issues/36539.

src/Polly/Policy.ExecuteOverloads.cs
martintmk: Nice, looks good then!

martintmk (approved):

martincostello:
This should give a green build now, so we should be good to merge this once the `8.0.100` SDK is released.

PR #1745, opened 2023-10-28
Validate AoT compatibility

martincostello:
Add native AoT project to repro issues from #1732 and #1744.

~~Builds on top of #1743, so needs rebasing once that is merged.~~

### DIFF
--- a/test/Polly.AotTest/Program.cs
+++ b/test/Polly.AotTest/Program.cs
@@ -1,8 +1,1 @@
-using Polly.Utils.Pipeline;
-
-// See https://github.com/App-vNext/Polly/issues/1732#issuecomment-1782466692.
-// This code is needed as a workaround until https://github.com/dotnet/runtime/issues/94131 is resolved.
-var pipeline = CompositeComponent.Create(new[] { PipelineComponent.Empty, PipelineComponent.Empty }, null!, null!);
-await pipeline.ExecuteCore<int, int>((state, context) => default, default!, default);
-
 Console.WriteLine("Hello Polly!");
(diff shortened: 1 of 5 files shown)


test/Polly.AotTest/Program.cs
```diff
@@ -0,0 +1,8 @@
+using Polly.Utils.Pipeline;
+
+// See https://github.com/App-vNext/Polly/issues/1732#issuecomment-1782466692.
+// This code is needed as a workaround until https://github.com/dotnet/runtime/issues/94131 is resolved.
+var pipeline = CompositeComponent.Create(new[] { PipelineComponent.Empty, PipelineComponent.Empty }, null!, null!);
```
martintmk, line 5: Can we create a more elaborate pipeline (possibly of all strategies we have).

Would cover much more code that way.

martincostello: As I understood it, this is only needed to detect the generic cycles that are missed in the analyzer due to dotnet/runtime#94131. Otherwise, all the code is inspected by the analyser through the presence of the `<TrimmerRootAssembly>` items, so we don't actually need to exercise all the code paths in the actual program except for this specific case.

vitek-karas: The `TrimmerRootAssembly` items in the project will "root" all code in the assemblies it specified, regardless if it's used by the `Main` here. The compiler will also analyze all of that code, and report warnings if there are any.

I would think you should not need any code in `Main`, but generic cycle detection is tricky, so it might need to be "Triggered" like this. You should try to remove this though and see if it still reports the cycle (I hope it will).

martincostello: Looks like you're right, the hint is redundant when compiling.

martincostello:
This PR demonstrates the repro finds the issue; the changes have been integrated into #1737.

PR #1749, opened 2023-10-29
[Docs] Hedging context

martintmk:
# Pull Request

## The issue or feature being addressed

Closes #1727 

## Confirm the following

- [x]  I started this PR by branching from the head of the default branch
- [x]  I have targeted the PR to merge into the default branch
- [x]  I have included unit tests for the issue/feature
- [x]  I have successfully run a local build

PR #1751, opened 2023-10-30
[Docs] Add diagram about hedging's context and callbacks

peter-csala:
# Pull Request

- #1749

- As it was discussed in the review comments: it makes sense to visualize the sequence of actions related to context and callbacks
- @martintmk Can you please double check whether the diagram depicts everything and in the right order?
- [Preview](https://github.com/peter-csala/Polly/blob/add-sequence-diagram-about-hedgings-contexts-and-callbacks/docs/strategies/hedging.md#sequence-diagram-about-contexts-and-callbacks)

## Confirm the following

### DIFF
--- a/docs/strategies/hedging.md
+++ b/docs/strategies/hedging.md
@@ -454,7 +454,6 @@ Here's the flow:
 - After the strategy has an accepted result from a hedged action, the resilience context from the action is merged back into the primary context.
 - All ongoing hedged actions are cancelled and discarded. The hedging strategy awaits the propagation of cancellation.
 
-
 ### Merging action context into the primary context
 
 After merging, the primary context contains:
--- a/docs/strategies/hedging.md
+++ b/docs/strategies/hedging.md
@@ -454,6 +454,7 @@ Here's the flow:
 - After the strategy has an accepted result from a hedged action, the resilience context from the action is merged back into the primary context.
 - All ongoing hedged actions are cancelled and discarded. The hedging strategy awaits the propagation of cancellation.
 
+
 ### Merging action context into the primary context
 
 After merging, the primary context contains:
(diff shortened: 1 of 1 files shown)


docs/strategies/hedging.md
```diff
@@ -471,6 +471,45 @@ Some insights about the hedging callback behavior in relation to hedging context
...
+    P->>H: Calls ExecuteCore <br/>with Primary Context
+    H-->>H: Deep clones <br/>Primary Context
+
+    H->>+UC: Invokes <br/>with Primary Context
+    UC-->>UC: Processes <br/>+ Modifies Context
+    UC->>-H: Fails
+
+    H-->>H: Deep clones <br/>Primary Context
+    H->>+AG: Invokes <br/>with both Contexts
+    AG-->>AG: Modifies Primary Context
```
martintmk, line 497:
```suggestion
    AG-->>AG: Modifies Primary or Action Context 
```

you can also modify action context here.

docs/strategies/hedging.md
```diff
@@ -471,6 +471,45 @@ Some insights about the hedging callback behavior in relation to hedging context
...
+    participant H as Hedging
+    participant AG as ActionGenerator
+    participant UC as UserCallback
+    participant HUC as HedgedUserCallback
+    participant OH as OnHedging
+
+    P->>H: Calls ExecuteCore <br/>with Primary Context
+    H-->>H: Deep clones <br/>Primary Context
+
+    H->>+UC: Invokes <br/>with Primary Context
```
martintmk, line 491: note: in this call the Primary = Action context, they are filled but these are the same instances.

But I am thinking, is this OK? 
Shall we have the same behavior even for primary action? Just clone the primary context and use it for primary callback too. wdyt?

docs/strategies/hedging.md
peter-csala: I've updated the `OnHedging` part also.

docs/strategies/hedging.md
peter-csala: I think that approach would be cleaner as the original request won't be treated in a special way.

martintmk: True, I have prepared the PR with the changes: #1754 

Can you also align the diagram with it?

martintmk:
Some real-case scenario where this is used:

https://github.com/dotnet/extensions/blob/3038d565967654ffb92cdd5edf434930c9249e91/src/Libraries/Microsoft.Extensions.Http.Resilience/Hedging/ResilienceHttpClientBuilderExtensions.Hedging.cs#L84

docs/strategies/hedging.md
peter-csala: Done

martintmk (approved):

PR #1752, opened 2023-10-30
Run mutation tests on Windows instead of Linux

martincostello:
As discovered in #1738, there's an issue with Stryker with .NET 8 that breaks it on Linux and macOS, so run on Windows until that's resolved.

Doing this as a separate PR to reduce the scope of the changes needed ahead of our actual support for .NET 8.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,6 +25,10 @@ jobs:
     name: ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
 
+    env:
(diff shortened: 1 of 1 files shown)


martintmk:
The downside is that this will slow-down our builds :/

martintmk (approved):

martincostello:
> The downside is that this will slow-down our builds :/

- This build: [50m 18s](https://github.com/App-vNext/Polly/actions/runs/6691268039/job/18178173753?pr=1752)
- This other build: [47m 0s](https://github.com/App-vNext/Polly/actions/runs/6691287395/job/18178229803?pr=1753)

I think for 3 minutes extra it's fine 😅

PR #1763, opened 2023-10-31
[Docs] Add minor tweaks to migration guide 1/3

peter-csala:
# Pull Request

## The issue or feature being addressed

 - Part 1: Review from _Migrating execution policies_ till _Migrating timeout policies_
 - Part 2: Add **missing** _Migrating circuit breaker policies_
 - Part 3: Review from _Migrating other policies_ till _Interoperability between policies and resilience pipelines_

## Details on the issue fix or feature implementation
- _Migrating execution policies_
   - Reordered the interface list to align with code samples
   - Added line breaks to the V7's non-generic examples to have the same look and feel as the generic ones
   - Renamed the `Execute{Async}`'s `CancellationToken` parameter to `token` to avoid ambiguity 
   - Renamed lambda function's `r` parameter to `result` to align with V8's sample
   - In case of V8 used static anonymous function whenever it is possible
   - Added important section at the end as a TL;DR + link for further info
- _Migrating policy wrap_ 
  -  Added important section at the end as a TL;DR + link for further info
- _Migrating retry policies_
  - Fixed V7's code examples indentations
  - Moved and fixed the V7's `Retry forever` example
  -  In case of V8 used static anonymous function whenever it is possible
  - Added important section at the end as a TL;DR + link for further info 
- _Migrating rate limit policies_ 
   - Added important section at the end as a TL;DR + link for further info  
- _Migrating bulkhead policies_ 
   - Fixed V7's code examples indentations to align with rate limiter examples
   - Added important section at the end as a TL;DR + link for further info   
- _Migrating timeout policies_ 
   - Added important section at the end as a TL;DR + link for further info

[Preview](https://github.com/peter-csala/Polly/blob/add-minor-tweaks-to-migration-guide/docs/migration-v8.md)
  
## Confirm the following

- [x]  I started this PR by branching from the head of the default branch
- [x]  I have targeted the PR to merge into the default branch
- [ ]  I have included unit tests for the issue/feature
- [x]  I have successfully run a local build

PR #1764, opened 2023-10-31
[Docs] Add circuit breaker to the migration guide

peter-csala:
# Pull Request

- #1763  
  - As I've pointed in the previous PR the migration guide did not have a circuit breaker section 

- Created a `Migration.CircuitBreaker.cs` file with V7 and V8 samples
- Added a _Migrating circuit breaker policies_ section to the `migration-v8.md` files

## Open questions
- Does it make sense to have a "standard"/"classic" section under the V7 paragraph?
  - Or shall I delete it since it is not supported in V8?
  - Personally I would vote to keep it to make it explicit that the "classic" version support is discontinued

**Answer: keep it under the V7 paragraph**

-  In the github wiki under the Circuit breaker we have a "suggest" which does not work for V8
   - [Wiki link](https://github.com/App-vNext/Polly/wiki/Circuit-Breaker#reducing-thrown-exceptions-when-the-circuit-is-broken)
   - Several weeks ago [I've detailed the problem](https://stackoverflow.com/questions/77203937/polly-v8-circuitbreaker-behavior-is-different-the-circuit-gets-stuck-open/77214137#77214137) from V8 perspective on StackOverflow
   - What do want to do with this?
      - Shall I add a new anti-pattern under the Circuit Breaker page?
      - Shall I add a note at the end of this circuit breaker migration paragraph?
      - Other? 

**Answer: add it to the anti-patterns section and add reference to that from the migration guide**

## Confirm the following

### DIFF
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -142,7 +142,7 @@ await pipelineT.ExecuteAsync(static async token =>
 ```
 <!-- endSnippet -->
 
-> [!IMPORTANT]
+> [!TIP]
 >
 > Things to remember:
 >
@@ -195,7 +195,7 @@ ResiliencePipeline pipeline = new ResiliencePipelineBuilder()
 
 See [fallback after retries](strategies/fallback.md#fallback-after-retries) for an example on how the strategies are executed.
 
-> [!IMPORTANT]
+> [!TIP]
 >
 > Things to remember:
 >
@@ -404,7 +404,7 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(new RetryStrategyO
 ```
 <!-- endSnippet -->
 
-> [!IMPORTANT]
+> [!TIP]
 >
 > Things to remember:
 >
@@ -475,7 +475,7 @@ ResiliencePipeline<HttpResponseMessage> pipelineT = new ResiliencePipelineBuilde
 ```
 <!-- endSnippet -->
 
-> [!IMPORTANT]
+> [!TIP]
 >
 > Things to remember:
 >
@@ -538,7 +538,7 @@ ResiliencePipeline<HttpResponseMessage> pipelineT = new ResiliencePipelineBuilde
 ```
 <!-- endSnippet -->
 
-> [!IMPORTANT]
+> [!TIP]
 >
 > Things to remember:
 >
@@ -588,7 +588,7 @@ ResiliencePipeline<HttpResponseMessage> pipelineT = new ResiliencePipelineBuilde
 ```
 <!-- endSnippet -->
 
-> [!IMPORTANT]
+> [!TIP]
 >
 > Things to remember:
 >
@@ -758,7 +758,7 @@ await manualControl.CloseAsync(); // transitions into Closed state
 
 ____
 
-> [!IMPORTANT]
+> [!TIP]
 >
 > Things to remember:
 >
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -682,8 +682,8 @@ ICircuitBreakerPolicy cbPolicy = (ICircuitBreakerPolicy)asyncPolicy;
 bool isOpen = cbPolicy.CircuitState == CircuitState.Open || cbPolicy.CircuitState == CircuitState.Isolated;
 
 // Manually control state
-cbPolicy.Isolate(); // transitions into Isolated state
-cbPolicy.Reset(); // transitions into Closed state
+cbPolicy.Isolate(); // Transitions into the Isolated state
+cbPolicy.Reset(); // Transitions into the Closed state
 ```
 <!-- endSnippet -->
 
@@ -711,7 +711,7 @@ ResiliencePipeline pipeline = new ResiliencePipelineBuilder()
     .Build();
 
 // Create a generic pipeline with circuit breaker. Because ResiliencePipeline<T> supports both sync and async
-// callbacks, there is no need to define it twice.
+// callbacks, there is also no need to define it twice.
 ResiliencePipeline<HttpResponseMessage> pipelineT = new ResiliencePipelineBuilder<HttpResponseMessage>()
     .AddCircuitBreaker(new CircuitBreakerStrategyOptions<HttpResponseMessage>
     {
@@ -745,8 +745,8 @@ ResiliencePipeline pipelineState = new ResiliencePipelineBuilder()
 bool isOpen = stateProvider.CircuitState == CircuitState.Open || stateProvider.CircuitState == CircuitState.Isolated;
 
 // Manually control state
-await manualControl.IsolateAsync(); // transitions into Isolated state
-await manualControl.CloseAsync(); // transitions into Closed state
+await manualControl.IsolateAsync(); // Transitions into the Isolated state
+await manualControl.CloseAsync(); // Transitions into the Closed state
 ```
 <!-- endSnippet -->
 
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -750,6 +750,14 @@ await manualControl.CloseAsync(); // transitions into Closed state
 ```
 <!-- endSnippet -->
 
+> [!NOTE]
+>
+> On the [V7's wiki page](https://github.com/App-vNext/Polly/wiki/Circuit-Breaker#reducing-thrown-exceptions-when-the-circuit-is-broken) you can find a tip how to reduce thrown exceptions. The described technique does **not** work with V8.
+>
+> Under the [circuit breaker's anti-patterns](strategies/circuit-breaker.md#4---reducing-thrown-exceptions) you can find the suggested way for V8.
+
+____
+
 > [!IMPORTANT]
 >
 > Things to remember:
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -752,7 +752,9 @@ await manualControl.CloseAsync(); // Transitions into the Closed state
 
 > [!NOTE]
 >
-> On the [V7's wiki page](https://github.com/App-vNext/Polly/wiki/Circuit-Breaker#reducing-thrown-exceptions-when-the-circuit-is-broken) you can find a tip how to reduce thrown exceptions. The described technique does **not** work with V8.
+> In case of V7 you could do an optimization to reduce the thrown exceptions.
+>
+> You could guard the `Execute{Async}` call with a condition that the circuit is not broken. This technique does **not** work with V8.
 >
 > Under the [circuit breaker's anti-patterns](strategies/circuit-breaker.md#4---reducing-thrown-exceptions) you can find the suggested way for V8.
 
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -603,7 +603,7 @@ This section describes how to migrate v7 circuit breaker policies to V8 circuit 
 
 ### Circuit breaker in v7
 
-V7's "Standard" Circuit Breaker policy could be defined like bellow:
+V7's "Standard" Circuit Breaker policy could be defined like below:
 
 <!-- snippet: migration-circuit-breaker-v7 -->
 ```cs
@@ -637,7 +637,7 @@ IAsyncPolicy<HttpResponseMessage> asyncPolicyT = Policy<HttpResponseMessage>
 ```
 <!-- endSnippet -->
 
-V7's Advanced Circuit Breaker policy could be defined like bellow:
+V7's Advanced Circuit Breaker policy could be defined like below:
 
 <!-- snippet: migration-advanced-circuit-breaker-v7 -->
 ```cs
@@ -691,7 +691,7 @@ cbPolicy.Reset(); // transitions into Closed state
 
 > [!IMPORTANT]
 >
-> Polly V8 does not support the standard ("classic") circuit breaker with consecutive failure counting.
+> Polly V8 does not support the standard (_"classic"_) circuit breaker with consecutive failure counting.
 >
 > In case of V8 you can define a Circuit Breaker strategy which works like the advanced circuit breaker in V7.
 
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -687,8 +687,7 @@ cbPolicy.Reset(); // transitions into Closed state
 ```
 <!-- endSnippet -->
 
-
-### Timeout in v8
+### Circuit breaker in v8
 
 > [!IMPORTANT]
 >
@@ -759,7 +758,6 @@ await manualControl.CloseAsync(); // transitions into Closed state
 > - Use the `CircuitBreakerStrategyOptions{<TResult>}` to customize your circuit breaker behavior to meet your requirements
 >
 > For further information please check out the [Circuit Breaker resilience strategy documentation](strategies/circuit-breaker.md).
-
 
 ## Migrating other policies
 
--- a/docs/migration-v8.md
+++ b/docs/migration-v8.md
@@ -142,7 +142,7 @@ await pipelineT.ExecuteAsync(static async token =>
 ```
 <!-- endSnippet -->
 
-> [!TIP]
+> [!IMPORTANT]
 >
 > Things to remember:
 >
@@ -195,7 +195,7 @@ ResiliencePipeline pipeline = new ResiliencePipelineBuilder()
 
 See [fallback after retries](strategies/fallback.md#fallback-after-retries) for an example on how the strategies are executed.
 
-> [!TIP]
+> [!IMPORTANT]
 >
 > Things to remember:
 >
@@ -404,7 +404,7 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(new RetryStrategyO
 ```
 <!-- endSnippet -->
 
-> [!TIP]
+> [!IMPORTANT]
 >
 > Things to remember:
 >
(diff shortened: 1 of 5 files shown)


martintmk:
> Does it make sense to have a "standard"/"classic" section under the V7 paragraph?

I think we should keep it for the reasons you mentioned - note that v8 does not have it anymore.

> Shall I add a note at the end of this circuit breaker migration paragraph?

I think this is fine, another alternative is to use the low-allocation `ExecuteOutcomeAsync` API to avoid throwing exception. In that case caller can just execute the callback that will fail very quickly with minimum overhead.

src/Snippets/Docs/Migration.CircuitBreaker.cs
```diff
@@ -0,0 +1,154 @@
+using System.Net.Http;
+using System.Runtime.Versioning;
+using Polly.CircuitBreaker;
+using Snippets.Docs.Utils;
+
+namespace Snippets.Docs;
+
+internal static partial class Migration
+{
+    [RequiresPreviewFeatures]
```
martintmk, line 10: Just curious, why this attribute?

martintmk (approved):

peter-csala:
> I think this is fine, another alternative is to use the low-allocation `ExecuteOutcomeAsync` API to avoid throwing exception. In that case caller can just execute the callback that will fail very quickly with minimum overhead.

I think it might make sense to mention this case on the migration guide and provide a link to the related anti-pattern section where we detail the solution. Do you agree?

src/Snippets/Docs/Migration.CircuitBreaker.cs
martintmk: Yes, please, it's a mistake.

peter-csala:
> I think this is fine, another alternative is to use the low-allocation `ExecuteOutcomeAsync` API to avoid throwing exception. In that case caller can just execute the callback that will fail very quickly with minimum overhead.

I try to put together the suggested way but it feels a bit clumsy for me.

The anti-pattern
```cs
var stateProvider = new CircuitBreakerStateProvider();
var circuitBreaker = new ResiliencePipelineBuilder()
    .AddCircuitBreaker(new()
    {
        ShouldHandle = new PredicateBuilder().Handle<HttpRequestException>(),
        BreakDuration = TimeSpan.FromSeconds(0.5),
        StateProvider = stateProvider
    })
    .Build();

if (stateProvider.CircuitState
    is not CircuitState.Open
    and not CircuitState.Isolated)
{
    await circuitBreaker.ExecuteAsync(static ct =>
    {
        // Your code goes here
        return default;
    }, CancellationToken.None);
}
```
The pattern
```cs
var context = ResilienceContextPool.Shared.Get();
var circuitBreaker = new ResiliencePipelineBuilder()
    .AddCircuitBreaker(new()
    {
        ShouldHandle = new PredicateBuilder().Handle<HttpRequestException>(),
        BreakDuration = TimeSpan.FromSeconds(0.5),
    })
    .Build();

Outcome<bool> result = await circuitBreaker.ExecuteOutcomeAsync(static (ctx, state) =>
{
    // Your code goes here

    return Outcome.FromResultAsValueTask(true);
}, context, "state");

ResilienceContextPool.Shared.Return(context);
```
Using `bool` as a `TResult` feels a bit odd. Even if I replace that to `object?` still does not feel good
```cs
Outcome<object?> result = await circuitBreaker.ExecuteOutcomeAsync(static (ctx, state) =>
{
    // Your code goes here

    return Outcome.FromResultAsValueTask<object?>(null);
}, context, "state");
```

@martintmk Do you have any other alternative which feels more natural?

martintmk:
@peter-csala 

The main idea is to just execute the callback as one normally would and afterwards one can check the result. For example:

``` csharp
Outcome<bool> result = await circuitBreaker.ExecuteOutcomeAsync(static (ctx, state) => ... );

if (result.Exception is BrokenCircuitException)
{
   // the execution was stopped by CB
}
else
{
   // the callback was executed and passed through CB
}
```

peter-csala:
@martintmk Yeah I do understand that part. Let me try to rephrase my problem.

There is no non-generic `Outcome` data structure. That's why using `ExecuteOutcomeAsync` does not neatly work well with async methods.

If there were a non-generic `Outcome` then you would write something like this 
```
Outcome result = await circuitBreaker.ExecuteOutcomeAsync(static (ctx, state) =>
{
    // Your code goes here

    return ValueTask.FromResult(new Outcome());
}, context, "state");
```
Anyway maybe overthinking and it is the same problem as the `TaskCompletionSource`

martintmk:
> @martintmk Yeah I do understand that part. Let me try to rephrase my problem.
> 
> There is no non-generic `Outcome` data structure. That's why using `ExecuteOutcomeAsync` does not neatly work well with async methods.
> 
> If there were a non-generic `Outcome` then you would write something like this
> 
> ```cs
> Outcome result = await circuitBreaker.ExecuteOutcomeAsync(static (ctx, state) =>
> {
>     // Your code goes here
> 
>     return ValueTask.FromResult(new Outcome());
> }, context, "state");
> ```
> 
> Anyway maybe overthinking and it is the same problem as with the `TaskCompletionSource<TResult>`.

You are right, there is currently no way to return "void outcome", you need to use workaround you mentioned above. Still, this is for more advanced scenarios so I would wait for feedback until someone really complains :)

peter-csala:
> You are right, there is currently no way to return "void outcome", you need to use workaround you mentioned above. Still, this is for more advanced scenarios so I would wait for feedback until someone really complains :)

In this case I will use `HttpResponseMessage` as `TResult` to make the example smooth.

docs/migration-v8.md
```diff
@@ -597,6 +597,176 @@ ResiliencePipeline<HttpResponseMessage> pipelineT = new ResiliencePipelineBuilde
 >
 > For further information please check out the [Timeout resilience strategy documentation](strategies/timeout.md).
 
+## Migrating circuit breaker policies
+
+This section describes how to migrate v7 circuit breaker policies to V8 circuit breaker strategies.
+
+### Circuit breaker in v7
+
+V7's "Standard" Circuit Breaker policy could be defined like bellow:
```
martincostello, line 606:
```suggestion
V7's "Standard" Circuit Breaker policy could be defined like below:
```

docs/migration-v8.md
```diff
@@ -597,6 +597,176 @@ ResiliencePipeline<HttpResponseMessage> pipelineT = new ResiliencePipelineBuilde
...
+// Create generic async circuit breaker
+IAsyncPolicy<HttpResponseMessage> asyncPolicyT = Policy<HttpResponseMessage>
+    .Handle<SomeExceptionType>()
+    .CircuitBreakerAsync(
+        handledEventsAllowedBeforeBreaking: 2,
+        durationOfBreak: TimeSpan.FromSeconds(1));
+```
+<!-- endSnippet -->
+
+V7's Advanced Circuit Breaker policy could be defined like bellow:
```
martincostello, line 640:
```suggestion
V7's Advanced Circuit Breaker policy could be defined like below:
```

docs/migration-v8.md
```diff
@@ -597,6 +597,176 @@ ResiliencePipeline<HttpResponseMessage> pipelineT = new ResiliencePipelineBuilde
...
+cbPolicy.Isolate(); // transitions into Isolated state
+cbPolicy.Reset(); // transitions into Closed state
+```
+<!-- endSnippet -->
+
+### Circuit breaker in v8
+
+> [!IMPORTANT]
+>
+> Polly V8 does not support the standard ("classic") circuit breaker with consecutive failure counting.
```
martincostello, line 694:
```suggestion
> Polly V8 does not support the standard (_"classic"_) circuit breaker with consecutive failure counting.
```

docs/migration-v8.md
```diff
@@ -597,6 +597,176 @@ ResiliencePipeline<HttpResponseMessage> pipelineT = new ResiliencePipelineBuilde
...
+
+// Manually control state
+await manualControl.IsolateAsync(); // transitions into Isolated state
+await manualControl.CloseAsync(); // transitions into Closed state
+```
+<!-- endSnippet -->
+
+> [!NOTE]
+>
+> On the [V7's wiki page](https://github.com/App-vNext/Polly/wiki/Circuit-Breaker#reducing-thrown-exceptions-when-the-circuit-is-broken) you can find a tip how to reduce thrown exceptions. The described technique does **not** work with V8.
```
martincostello, line 755: I'm not sure we should point this out - for each page in the wiki where the content is "v8-ified" in the docs, I've been aiming to go through them and just essentially blank them out and provide a link to the new location in pollydocs.

docs/strategies/circuit-breaker.md
```diff
@@ -535,3 +535,84 @@ public Downstream1Client(
 > The above sample code used the `AsAsyncPolicy<HttpResponseMessage>()` method to convert the `ResiliencePipeline<HttpResponseMessage>` to `IAsyncPolicy<HttpResponseMessage>`.
 > It is required because the `AddPolicyHandler()` method anticipates an `IAsyncPolicy<HttpResponse>` parameter.
 > Please be aware that, later an `AddResilienceHandler()` will be introduced in the `Microsoft.Extensions.Http.Resilience` package which is the successor of the `Microsoft.Extensions.Http.Polly`.
+
+### 4 - Reducing thrown exceptions
+
+In case of Circuit Breaker when it is either in `Open` or in `Isolated` state new requests are rejected immediately.
```
martincostello, line 541:
```suggestion
In case of Circuit Breaker when it is either in the `Open` or `Isolated` state new requests are rejected immediately.
```

docs/strategies/circuit-breaker.md
```diff
@@ -535,3 +535,84 @@ public Downstream1Client(
...
+    }, CancellationToken.None);
+
+    // Your code goes here to process response
+}
+```
+<!-- endSnippet -->
+
+**Reasoning**:
+
+- The problem with this approach is that the circuit breaker will never transition into `HalfOpen` state.
```
martincostello, line 577:
```suggestion
- The problem with this approach is that the circuit breaker will never transition into the `HalfOpen` state.
```

docs/strategies/circuit-breaker.md
```diff
@@ -535,3 +535,84 @@ public Downstream1Client(
...
+- The circuit breaker does not act as an active object. In other words the state transition does not happen automatically in the background.
+- The circuit transition into `HalfOpen` state when the `Execute{Async}` is called and the `BreakDuration` elapsed.
```
martincostello, line 579:
```suggestion
- The circuit transition into the `HalfOpen` state when the `Execute{Async}` method is called and the `BreakDuration` elapsed.
```

docs/strategies/circuit-breaker.md
```diff
@@ -535,3 +535,84 @@ public Downstream1Client(
...
+else
+{
+    // Your code goes here to process response (outcome.Result)
+}
+```
+<!-- endSnippet -->
+
+**Reasoning**:
+
+- The `ExecuteOutcomeAsync` is a low-allocation API which does not throw exception rather captures it inside the `Exception` property of the `Outcome` data structure.
```
martincostello, line 617:
```suggestion
- The `ExecuteOutcomeAsync` is a low-allocation API which does not throw exceptions; rather it captures it inside the `Exception` property of the `Outcome` data structure.
```

docs/strategies/circuit-breaker.md
```diff
@@ -535,3 +535,84 @@ public Downstream1Client(
...
+- Since your are calling one of the `Execute` methods that's why the circuit breaker can transition into `HalfOpen` state.
```
martincostello, line 618:
```suggestion
- Since you are calling one of the `Execute` methods, that's why the circuit breaker can transition into the `HalfOpen` state.
```

peter-csala:
@martincostello , @martintmk 

I've just realized that the `[!IMPORTANT]` blocks are [rendered](https://dotnet.github.io/docfx/docs/markdown.html?tabs=linux%2Cdotnet#alerts) as red boxes via docfx:

DOCFX
Screenshot 2023-11-02 at 10 51 13

GITHUB
Screenshot 2023-11-02 at 10 51 27

It feels a bit "aggressive" to me for a tip. When I changed to `[!TIP]` then github rendering failed :(

DOCFX
Screenshot 2023-11-02 at 10 49 46

GITHUB
Screenshot 2023-11-02 at 10 49 39

What do you guys think what should we use on the migration V8 page?

I've found this bit odd [alternative](https://stackoverflow.com/a/76515821/13268855) to render tip as a single column table.

src/Snippets/Docs/CircuitBreaker.cs
```diff
@@ -214,4 +214,68 @@ [new Uri("https://downstreamN.com")] = GetCircuitBreaker()
...
+        {
+            var response = await IssueRequest();
+            return Outcome.FromResult(response);
+        }, context, "state");
+
+        ResilienceContextPool.Shared.Return(context);
+
+        if (outcome.Exception is BrokenCircuitException)
+        {
+            // the execution was stopped by CB
```
martincostello, line 272:
```suggestion
            // The execution was stopped by the circuit breaker
```

src/Snippets/Docs/CircuitBreaker.cs
```diff
@@ -214,4 +214,68 @@ [new Uri("https://downstreamN.com")] = GetCircuitBreaker()
...
+        }
+        else
+        {
+            // Your code goes here to process response (outcome.Result)
```
martincostello, line 276:
```suggestion
            // Your code goes here to process the response (outcome.Result)
```

src/Snippets/Docs/Migration.CircuitBreaker.cs
```diff
@@ -0,0 +1,152 @@
...
+                minimumThroughput: 2,
+                durationOfBreak: TimeSpan.FromSeconds(1));
+
+        // Check circuit state
+        ICircuitBreakerPolicy cbPolicy = (ICircuitBreakerPolicy)asyncPolicy;
+        bool isOpen = cbPolicy.CircuitState == CircuitState.Open || cbPolicy.CircuitState == CircuitState.Isolated;
+
+        // Manually control state
+        cbPolicy.Isolate(); // transitions into Isolated state
+        cbPolicy.Reset(); // transitions into Closed state
```
martincostello, line 90:
```suggestion
        cbPolicy.Isolate(); // Transitions into the Isolated state
        cbPolicy.Reset(); // Transitions into the Closed state
```

src/Snippets/Docs/Migration.CircuitBreaker.cs
```diff
@@ -0,0 +1,152 @@
...
+                ShouldHandle = new PredicateBuilder().Handle<SomeExceptionType>(),
+                FailureRatio = 0.5d,
+                SamplingDuration = TimeSpan.FromSeconds(5),
+                MinimumThroughput = 2,
+                BreakDuration = TimeSpan.FromSeconds(1)
+            })
+            .Build();
+
+        // Create a generic pipeline with circuit breaker. Because ResiliencePipeline<T> supports both sync and async
+        // callbacks, there is no need to define it twice.
```
martincostello, line 113:
```suggestion
        // callbacks, there is also no need to define it twice.
```

src/Snippets/Docs/Migration.CircuitBreaker.cs
```diff
@@ -0,0 +1,152 @@
...
+                ManualControl = manualControl
+            })
+            .Build();
+
+        // Check circuit state
+        bool isOpen = stateProvider.CircuitState == CircuitState.Open || stateProvider.CircuitState == CircuitState.Isolated;
+
+        // Manually control state
+        await manualControl.IsolateAsync(); // transitions into Isolated state
+        await manualControl.CloseAsync(); // transitions into Closed state
```
martincostello, line 148:
```suggestion
        await manualControl.IsolateAsync(); // Transitions into the Isolated state
        await manualControl.CloseAsync(); // Transitions into the Closed state
```

docs/strategies/circuit-breaker.md
peter-csala: > rather it captures it

```
- The `ExecuteOutcomeAsync` is a low-allocation API which does not throw exceptions; rather it captures them inside the `Outcome` data structure.
```
Are you okay with this?

docs/migration-v8.md
peter-csala: Are you suggestion to delete the entire note block or just remove the wiki link or something else?

docs/strategies/circuit-breaker.md
martincostello: Sure.

docs/migration-v8.md
martincostello: I was referring to calling out anything in the old documentation as the plan is to remove it, rather than leave it lying around.

peter-csala: I was referring to that section because one might have applied that optimization for their V7 code. If we did not emphasize in the migration guide that do not use the same optimization for V8 because it breaks the circuit breaker normal behavior then it won't generate compile-time error rather give hard to find application bug.

martintmk (approved):

docs/migration-v8.md
martincostello: My concern is more pointing to documentation that won't exist anymore.

It would be better to phrase it based on an approach, not by explicitly referring to code using that approach that may have come from the old documentation.

martincostello (approved):

PR #1800, opened 2023-11-15
Publish AoT for test on macOS

martincostello:
This should work in .NET 8, so simplify by removing the condition.


martincostello:
Something foul is afoot...

martincostello:
Logged https://github.com/dotnet/sdk/issues/36960.

PR #1838, opened 2023-12-06
[Docs] Add retry and circuit breaker cheat sheets

peter-csala:
# Pull Request

Screenshot 2023-12-06 at 11 10 06
Screenshot 2023-12-06 at 11 10 21

## Confirm the following


docs/media/cheat-sheet-circuit-breaker.pdf
martincostello, line 1: - `underlying resource detected as unhealthy` => `underlying resource is detected as unhealthy`
- `In Closed state` => `In the Closed state`
- `allows traffic` => `allows invocations` (or another word, traffic is too "HTTP"-y to me) - there's other uses of the word "traffic" in the document that should similarly use a different term.
- `In Open state` => `In the Open state`
- `In HalfOpen state` => `In the HalfOpen state`
- `if it was in Open state` => `if it was in the Open state`
- `if it was in Isolated state` => `if it was in the Isolated state`

-

docs/media/cheat-sheet-retry.pdf
martincostello, line 1: - `the strategy will return with the last failure` - should this be "the strategy will throw the exception from the last failure" or similar?
- `on RetryAfter header` => `on HTTP Retry-After header`

peter-csala: > should this be "the strategy will throw the exception from the last failure" or similar

You can define retry based on the result object as well:
```cs
var retry = new ResiliencePipelineBuilder<int>()
    .AddRetry(new ()
    {
        ShouldHandle = new PredicateBuilder<int>().HandleResult(r => r < 5)
    }).Build();

int i = 0;

var final = retry.Execute(() => i++);
System.Console.WriteLine(final); // prints 3
```
So, I would keep the current wording. Can you accept it?

martincostello: Maybe reword it to be result or exception instead then? It's on the left-hand side separate to a specific example of code.

martincostello, line 1: Tiniest change - `RetryAfter` => [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After)

martincostello (approved):

PR #1863, opened 2023-12-21
[Docs] Revise reactive strategy sample

peter-csala:
# Pull Request

- `Arguments` and `PredicateArguments` were defined as `struct` not as `readonly struct`
- Inside the `samples/Extensibility/Program.cs` there were some dead code
- The comments on the reactive sample were a bit different than on the proactive sample

- Defined `Arguments` and `PredicateArguments` as `readonly struct`
- Deleted dead code Inside the `samples/Extensibility/Program.cs`
- Simplified one of the switch case's pattern matching
- Synchronized the comments between the reactive and the proactive samples
- Change the target of `samples/Extensibility/Extensibility.csproj` from .NET 7 to .NET 8

## Open question
- In order to enhance clarity shouldn't we use `TResult` instead of `T` for the `ResultReportingResilienceStrategy` class?

## Confirm the following

### DIFF
--- a/docs/extensibility/reactive-strategy.md
+++ b/docs/extensibility/reactive-strategy.md
@@ -4,7 +4,7 @@ This document describes how to set up a **Result reporting resilience strategy**
 
 ## Implementation
 
-Reactive resilience strategies inherit from the [`ResilienceStrategy<T>`](xref:Polly.ResilienceStrategy`1) base class. The implementation for this specific strategy is as follows:
+Reactive resilience strategies inherit from the [`ResilienceStrategy<T>`](xref:Polly.ResilienceStrategy`1) base class. The implementation for this specific strategy:
 
 <!-- snippet: ext-reactive-strategy -->
 ```cs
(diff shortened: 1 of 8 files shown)


martincostello:
> Would it make sense to target net8.0 for the samples/Extensibility/Extensibility.csproj?

Yes.

peter-csala:
> > Would it make sense to target net8.0 for the samples/Extensibility/Extensibility.csproj?
> 
> Yes.

Also, do we want to change most of the supporting classes, structs to positional records? 

Example #1
```cs
// Structs for arguments encapsulate details about specific events within the resilience strategy.
// Relevant properties to the event can be exposed. In this event, the actual execution time and the exceeded threshold are included.
// As per convention, all arguments should provide a "Context" property.
public readonly record struct OnThresholdExceededArguments(ResilienceContext Context, TimeSpan Threshold, TimeSpan Duration);
```

Example #2
```cs
// Apply validation attributes to guarantee the options' validity.
// The pipeline will handle validation automatically during its construction.
// Provide the delegate to be called when the threshold is surpassed.
// Ideally, arguments should share the delegate's name, but with an "Arguments" suffix.
public record TimingStrategyOptions(
    [property: Range(typeof(TimeSpan), "00:00:00", "1.00:00:00")]
    [property: Required] TimeSpan? Threshold, Func<OnThresholdExceededArguments, ValueTask>? OnThresholdExceeded)
    : ResilienceStrategyOptions
{

    // Assign a default name to the options for more detailed telemetry insights.
    public TimingStrategyOptions() : this(null, null) =>
        Name = "Timing";
}
```

martincostello:
We can do that separately if you like. The only reason it isn't targeting `net8.0` is probably because the project got created after I opened the initial PR to migrate to .NET 8, so it didn't exist when I did my original "replace 7.0 with 8.0" commit.

peter-csala:
@martincostello Is there anything that I need to tackle?

martincostello (approved):

Issue #1866, opened 2023-12-22
[Bug]: Retry Startegy stops retrying unexpectedly


Chr15P13t:
### Describe the bug

Using `DelayBackoffType.Exponential` retry startegy in combination with `MaxDelay` eventually leads to "infinite" wait. 

Delay computing method ignores `MaxDelay` in case of OverflowException and that leads to wait `TimeSpan.MaxValue` instead of `MaxDelay`

### Expected behavior

When retry delay computing leads to `TimeSpan` `OverflowException` and a `MaxDelay` is defined, `MaxDelay` should be used.

### Actual behavior

When retry delay computing leads to `TimeSpan` `OverflowException`, the applied Delay is `TimeSpan.MaxValue` regardless a `MaxDelay` is defined or not. 

## Consider following retry strategy setup

```C#
new RetryStrategyOptions<bool>()
{
    OnRetry = args =>
    {
        _logger.LogWarning("(Key: {OperationKey}) attempt {AttemptCount} failed after {AttemptDuration}ms, next attempt will take place in {RetryDelay}s",
            args.Context.OperationKey,
            args.AttemptNumber,
            args.Duration.TotalMilliseconds,
            args.RetryDelay.TotalSeconds);

        return new ValueTask();
    },
    Delay = TimeSpan.FromSeconds(2),
    BackoffType = DelayBackoffType.Exponential,
    MaxDelay = TimeSpan.FromMinutes(5),
    MaxRetryAttempts = int.MaxValue
};
```

## Observation in logs

Executing a pipeline base on this startegy leads to inconsistent behavior after the attempt 39, as I've observed in my logs:

```LOGS
(Key: 159e42ca-e781-48a8-927d-418b4ce8a811) attempt 0 failed after 1061.6031ms, next attempt will take place in 2s
(Key: 159e42ca-e781-48a8-927d-418b4ce8a811) attempt 1 failed after 248.2402ms, next attempt will take place in 4s
...
(Key: 159e42ca-e781-48a8-927d-418b4ce8a811) attempt 7 failed after 320.1541ms, next attempt will take place in 256s
(Key: 159e42ca-e781-48a8-927d-418b4ce8a811) attempt 8 failed after 950.9286ms, next attempt will take place in 300s
...
(Key: 159e42ca-e781-48a8-927d-418b4ce8a811) attempt 38 failed after 294.6921ms, next attempt will take place in 300s
[...]

PR #1868, opened 2023-12-26
Fix retry overflow with max delay

Chr15P13t:
# Pull Request

Fix issue #1866 Retry Startegy stops retrying unexpectedly

In case computing retry delay throws `OverflowException`, `RetryHelper.GetRetryDelay(...)` returns `maxDelay.value` if not null, `TimeSpan.MaxValue` otherwise.

## Confirm the following

### DIFF
--- a/src/Polly.Core/Retry/RetryHelper.cs
+++ b/src/Polly.Core/Retry/RetryHelper.cs
@@ -31,9 +31,9 @@ internal static class RetryHelper
         }
         catch (OverflowException)
         {
-            if (maxDelay is not null)
+            if (maxDelay is { } value)
             {
-                return maxDelay.Value;
+                return value;
             }
 
             return TimeSpan.MaxValue;
--- a/src/Polly.Core/Retry/RetryHelper.cs
+++ b/src/Polly.Core/Retry/RetryHelper.cs
@@ -31,6 +31,11 @@ internal static class RetryHelper
         }
         catch (OverflowException)
         {
+            if (maxDelay is not null)
+            {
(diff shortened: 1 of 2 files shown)


Chr15P13t:
@dotnet-policy-service agree

src/Polly.Core/Retry/RetryHelper.cs
```diff
@@ -31,6 +31,11 @@ internal static class RetryHelper
         }
         catch (OverflowException)
         {
+            if (maxDelay is not null)
+            {
+                return maxDelay.Value;
```
martincostello, line 36:
```suggestion
            if (maxDelay is {} value)
            {
                return value;
```

martincostello (approved):
Just one minor suggestion.

src/Polly.Core/Retry/RetryHelper.cs
Chr15P13t: Just for my understanding, why do you prefer `is {} value` to `is not null`?

martincostello: Value will re-check for null internally (so it could throw if it wasn't), where-as doing it this way means it only checks once and we have access to the value when it isn't.

ChristophePietquin: I've commited your suggestion, thanks for the explanation.

Chr15P13t: I've commited your suggestion, thanks for the explanation.

PR #1888, opened 2024-01-10
Decrease the minimum allowed timeout

### LINKED ISSUE #1884: [Feature request]: Relax validation of TimeoutStrategyOptions.Timeout

Currently, there is a validation attribute on `TimeoutStrategyOptions.Timeout` that requires a timeout to be in the range of [00:00:01, 1.00:00:00]. This is probably reasonable for most operations, but for some a min timeout of 1s might be to long.

I have tests that are rather slow because they need to use the min timeout of 1s.

It would be nice if the lower limit could be set to 0 or 1ms so that users are free to choose a timeout that is appropriate for their use case.

### Additional context

_No response_

martintmk:
# Pull Request

Closes #1884 

Relaxing the min allowed timeout to 10ms. (from 1 second)

## Confirm the following

### DIFF
--- a/src/Polly.Core/Timeout/TimeoutStrategyOptions.cs
+++ b/src/Polly.Core/Timeout/TimeoutStrategyOptions.cs
@@ -19,7 +19,7 @@ public class TimeoutStrategyOptions : ResilienceStrategyOptions
     /// <value>
     /// This value must be greater than 1 second and less than 24 hours. The default value is 30 seconds.
     /// </value>
-    [Range(typeof(TimeSpan), "00:00:01", "1.00:00:00")]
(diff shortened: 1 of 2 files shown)


martincostello (approved):

silkfire:
@martintmk I noticed that the documentation of the property wasn't updated in this PR, it still states that the minimum is (above) 1 second.

On another note, is there a reason why the maximum value is 24 hours?

martincostello:
When we picked an upper limit, a day seemed a reasonable upper limit considering that is effectively millennia to a computer.

Having a thread trying to perform a single operation through Polly for up to an entire day seems on the extreme end for a thing for code to try and do, and we need to draw a line somewhere.

Do you have a use case where you need a single method invocation to run for more than 24 hours, but still need to eventually timeout?

Issue #1901, opened 2024-01-18
Simmy API Review and Feedback


martintmk:
I was playing with the new chaos strategies and I think we can simplify the setup of common faults by introducing something similar to `PredicateBuilder`, but this time for faults.

Instead of:

``` csharp
new ResiliencePipelineBuilder<string>()
    .AddChaosFault(new()
    {
        FaultGenerator = _ =>
        {
            Exception? ex = Random.Shared.Next(380) switch
            {
                < 100 => new InvalidOperationException(),
                < 180 => new HttpRequestException(),
                < 280 => new TimeoutException(),
                < 380 => new TimeoutException(),
                _ => null,
            };

            return ValueTask.FromResult(ex);
        }
    })
    .AddChaosResult(new()
    {
        OutcomeGenerator = _ =>
        {
            Outcome<string>? result = Random.Shared.Next(420) switch
            {
                < 100 => Outcome.FromException<string>(new InvalidOperationException()),
                < 200 => Outcome.FromException<string>(new HttpRequestException()),
                < 300 => Outcome.FromException<string>(new TimeoutException()),
                < 320 => Outcome.FromException<string>(new SocketException()),
                < 420 => Outcome.FromResult("Hello World"),
                _ => null,
            };

            return ValueTask.FromResult(result);
        }
    });
```

We can allow this simplified API for common scenarios:

``` csharp
new ResiliencePipelineBuilder<string>()
    .AddChaosFault(new()
    {
        FaultGenerator = new FaultGenerator()
            .AddFault<InvalidOperationException>() // default weight is 100
            .AddFault<HttpRequestException>(weight: 80)
            .AddFault<TimeoutException>()
            .AddFault<SocketException>()
    })
    .AddChaosResult(new()
    {
        OutcomeGenerator = new OutcomeGenerator<string>()
            .AddException<InvalidOperationException>() // default weight is 100
            .AddException<HttpRequestException>()
            .AddException<TimeoutException>()
            .AddException<SocketException>(weight: 20)
            .AddResult(() => "Hello World!")
    });
```

We can also use this issue for additional API feedback.

cc @martincostello , @vany0114

peter-csala:
@martintmk  Don't we want to covert this issue to something like Simmy API review? 

I have a couple of observations as well. Just to name a few:
- `Enabled`'s default value is `false`. I can guarantee that many people will 🤦🏻 when they realize that the missing piece is `Enabled = true`
[...]

PR #1909, opened 2024-01-19
Simmy API review Part 1

peter-csala:
# Pull Request

#1901 

### Latency
- Renamed `OnLatency` to `OnLatencyInjected`
   - **Justification**: Every other chaos strategy defines its success hook as `OnXYZInjected`

### Fault
- Removed obsolete documentation comment from the `FaultGenerator` of the `FaultStrategyOptions`
   - **Justification**: #1900 has removed `Fault` from the `FaultStrategyOptions`

### All chaos strategies
- Unified `OnXYZInjectedEvent` constant's value
   - **Justification**: All of them start with `Chaos` so, it will be easier to spot these events in the telemetry logs
- Set the `Name` property for each options to a default value
   - **Justification**: All resilience strategy specifies a default `Name`. Chaos strategies should not be an exception. 

### Misc
- Fixed copy-paste issue by renaming `optionsOnBehaviorInjected` variable to `optionsOnLatencyInjected` in `Chaos.Latency`

## Confirm the following


src/Polly.Core/Simmy/Fault/FaultConstants.cs
```diff
@@ -2,5 +2,5 @@
 
 internal static class FaultConstants
 {
-    public const string OnFaultInjectedEvent = "OnFaultInjectedEvent";
+    public const string OnFaultInjectedEvent = "OnFaultInjected";
```
martintmk, line 5: How about? 

```suggestion
    public const string OnFaultInjectedEvent = "Chaos.OnFault"; 
```

Similar for others (`Chaos.OnLatency`, `Chaos.OnBehavior`).

The idea is to immediately know this is chaos event.

src/Polly.Core/Simmy/Latency/LatencyStrategyOptions.cs
```diff
@@ -13,7 +13,7 @@ public class LatencyStrategyOptions : MonkeyStrategyOptions
     /// <remarks>
```
martintmk, line 13: We should also initialize the `Name` property for all chaos options. (same as for other strategies)

For example:

- `Chaos.Latency`
- `Chaos.Behavior`
- `Chaos.Outcome`
- `Chaos.Fault`

src/Polly.Core/Simmy/Fault/FaultConstants.cs
peter-csala: Good idea, I will apply to all chaos strategies.

src/Polly.Core/Simmy/Latency/LatencyStrategyOptions.cs
peter-csala: Done

src/Polly.Core/Simmy/Fault/FaultConstants.cs
peter-csala: Done

martintmk:
Couple of ideas:

How about we rename the following options:

- `BehaviorStrategyOptions` -> `ChaosBehaviorStrategyOptions`
- `FaultStrategyOptions` -> `ChaosFaultStrategyOptions`
- `LatencyStrategyOptions` -> `ChaosLatencyStrategyOptions`

The reasoning is to be consistent with other strategies and extensions:

i.e. :
- `AddRetry` -> `RetryStrategyOptions`
- `AddChaosXYZ` -> `ChaosXYZStrategyOptions`

I am also thinking whether we should also rename:

- `MonkeyStrategyOptions` -> `ChaosStrategyOptions`
- `MonkeyStrategyOptions<T>` -> `ChaosStrategyOptions<T>`

The reason is also consistency, we are adding `Chaos` strategies, not `Monkey` strategies.

peter-csala:
> Couple of ideas:
> 
> How about we rename the following options:
> 
>     * `BehaviorStrategyOptions` -> `ChaosBehaviorStrategyOptions`
> 
>     * `FaultStrategyOptions` -> `ChaosFaultStrategyOptions`
> 
>     * `LatencyStrategyOptions` -> `ChaosLatencyStrategyOptions`
> 
> 
> The reasoning is to be consistent with other strategies and extensions:
> 
> i.e. :
> 
>     * `AddRetry` -> `RetryStrategyOptions`
> 
>     * `AddChaosXYZ` -> `ChaosXYZStrategyOptions`
> 
> 
> I am also thinking whether we should also rename:
> 
>     * `MonkeyStrategyOptions` -> `ChaosStrategyOptions`
> 
>     * `MonkeyStrategyOptions<T>` -> `ChaosStrategyOptions<T>`
> 
> 
> The reason is also consistency, we are adding `Chaos` strategies, not `Monkey` strategies.
> 
> Finally, let's rename `AddChaosResult` to `AddChaosOutcome` (because we do allow injecting both results and exceptions)

I think what we should decide here whether we want to keep the `Monkey` anywhere or prefer something else (like `Chaos`) everywhere.  If we chose the latter one then I would prefer to have a separate PR just for that renaming.

I went through several chaos engineering glossaries (like [1](https://www.gremlin.com/docs/resources/glossary/), [2](https://developer.harness.io/docs/chaos-engineering/)) and `Chaos Fault` or simply just `Fault` seems to be an accepted term. But sometimes they used the `Experiment` as an overloaded term: refers to the process, refers to the injected fault. In short I did not find an industry standard naming for this.

cc: @martintmk , @martincostello, @vany0114  
What do you guys think?

martintmk:
My preference would be to keep the `Simmy` namespace (for branding and "googlability" for anyone familiar with older Simmy), but to use `Chaos` instead of `Monkey`. For anyone new to chaos engineering, seeing `Chaos` gives more clarity compared to `Monkey`.

Just to compare:

- `AddMonkeyFault`
- `AddMonkeyBehavior`
- `AddMonkeyLatency`
 
vs

- `AddChaosFault`
- `AddChaosBehavior`
- `AddChaosLatency`

Either way, consistency should be the key and would be nice to use too many terms:

So either:

- Simmy, Chaos
- Simmy, Monkey

martincostello:
Agreed - that was one of the comments in the docs PR. We should be consistently in terminology. I would also prefer Chaos, as that has a defined meaning rather than being a personification of it (a rogue monkey).

peter-csala:
Okay, I'll do the renaming in a new PR. 

In order to minimize merge conflicts, I'll not extend this PR's scope further.

martintmk (approved):

src/Polly.Core/Simmy/Latency/LatencyStrategyOptions.cs
```diff
@@ -7,13 +7,18 @@
 /// </summary>
 public class LatencyStrategyOptions : MonkeyStrategyOptions
```
martincostello, line 8: Could/should we mark any of these as `sealed`?

martincostello (approved):

src/Polly.Core/Simmy/Latency/LatencyStrategyOptions.cs
martintmk: Let's keep them open to be aligned with other options. For example `Microsoft.Extensions.Http.Resilience` derives from our options and changes some defaults. We want the same for new chaos options.

peter-csala: In case of Microsoft.Extensions.Http.Resilience:
https://github.com/dotnet/extensions/blob/62c78eee58287a6030c13ce3e4f21190506d7291/src/Libraries/Microsoft.Extensions.Http.Resilience/Polly/HttpTimeoutStrategyOptions.cs#L11

The `HttpTimeoutStrategyOptions` derives from the `TimeoutStrategyOptions`.

I think we should keep them "open" for the same purpose.

PR #1916, opened 2024-01-23
Simmy API review Part 3 - Set enabled to true by default

peter-csala:
# Pull Request

#1901 

- Changed the default value of `Enabled` from `false` to `true`
- Added clarifications about property precedence

## Confirm the following

### DIFF
--- a/docs/chaos/index.md
+++ b/docs/chaos/index.md
@@ -73,7 +73,7 @@ Chaos strategies (formerly known as Monkey strategies) are in essence a [Resilie
 All the strategies' options implement the [`ChaosStrategyOptions`](xref:Polly.Simmy.ChaosStrategyOptions) class as it contains the basic configuration for every chaos strategy.
 
 > [!IMPORTANT]
-> Please bear in mind that with the V8 API the chaos strategies are enabled by default. So, you can opt-out them one-by-one either via the `Enabled` or via the `EnabledGenerator` property.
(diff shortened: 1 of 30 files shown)


martincostello (approved):
This is much terser and intuitive.

I wonder if we should put a note in the documentation somewhere that people should take care with the default being on, as if this is being used in production you'd be getting chaos as an opt-out rather than an opt-in as a change in behaviour to previous releases(?) of Simmy, so any configuration that might be tied to turning chaos on or not would be inverted.

martintmk (approved):

peter-csala:
> I wonder if we should put a note in the documentation somewhere that people should take care with the default being on, as if this is being used in production you'd be getting chaos as an opt-out rather than an opt-in as a change in behaviour to previous releases(?) of Simmy, so any configuration that might be tied to turning chaos on or not would be inverted.

Done

docs/chaos/index.md
```diff
@@ -72,6 +72,11 @@ Chaos strategies (formerly known as Monkey strategies) are in essence a [Resilie
 
 All the strategies' options implement the [`ChaosStrategyOptions`](xref:Polly.Simmy.ChaosStrategyOptions) class as it contains the basic configuration for every chaos strategy.
 
+> [!IMPORTANT]
+> Please bear in mind that with the V8 API the chaos strategies are enabled by default. So, you can opt-out them one-by-one either via the `Enabled` or via the `EnabledGenerator` property.
```
martincostello, line 76:
```suggestion
> Please bear in mind that with the V8 API the chaos strategies are enabled by default. So, you can opt-out of them one-by-one either via the `Enabled` or via the `EnabledGenerator` property.
```

PR #1918, opened 2024-01-23
[Docs] Add clarification about property precedence

peter-csala:
# Pull Request

#1916 

- I've added some notes about property precedence but not everywhere 

- Added clarification for Timeout about `Timeout` and `TimeoutGenerator` 
- Added clarification for CircuitBreaker about `BreakDuration` and `BreakDurationGenerator` 
- Added clarification for Hedging about `Delay` and `DelayGenerator` 
- Added clarification for Chaos about `InjectionRate` and `InjectionRateGenerator` 

## Extra
- Added clarification and examples about next delay calculation for retry logic

## Confirm the following

### DIFF
--- a/.github/wordlist.txt
+++ b/.github/wordlist.txt
@@ -13,6 +13,7 @@ deserialization
 dotnet
 dotnetrocks
 durations
+enum
 eshoponcontainers
 extensibility
 flurl
--- a/.github/wordlist.txt
+++ b/.github/wordlist.txt
@@ -13,7 +13,6 @@ deserialization
 dotnet
 dotnetrocks
 durations
-enum
 eshoponcontainers
 extensibility
 flurl
--- a/docs/strategies/retry.md
+++ b/docs/strategies/retry.md
@@ -150,6 +150,19 @@ Step 3: Using the generator if supplied
 
 ### Linear
 
+This algorithm increases the delays for every attempt in a linear fashion if no jitter is used.
+
+Step 1: Calculating the base delay:
+
+- If `UseJitter` is set to `false` and `Delay` is specified then `Delay` multiplied by the actual attempt number will be used.
+- If `UseJitter` is set to `true` and `Delay` is specified then a random value is added to the `Delay` multiplied by the actual attempt number.
+  - The random value is between -25% of the newly calculated `Delay` and +25% of the newly calculated `Delay`.
+
+> [!NOTE]
+> Because the jitter's offset is based on the newly calculated delay that's why the new delay could be smaller than the previous one.
+
+Step 2 and 3 are the same as for the Constant algorithm.
+
 ### Exponential
 
 > [!TIP]
--- a/docs/strategies/retry.md
+++ b/docs/strategies/retry.md
@@ -123,7 +123,7 @@ There are many properties that may contribute to this calculation:
 > [!IMPORTANT]
 > The summarized description below is an implementation detail. It may change in the future without notice.
 
-The `BackoffType` property's data type is the [`DelayBackOffType`](xref:Polly.DelayBackOffType) enumeration. This primarily controls how the calculation is done.
+The `BackoffType` property's data type is the [`DelayBackoffType`](xref:Polly.DelayBackoffType) enumeration. This primarily controls how the calculation is done.
 
 ### Constant
 
--- a/docs/strategies/retry.md
+++ b/docs/strategies/retry.md
@@ -127,13 +127,13 @@ The `BackoffType` property's data type is the [`DelayBackOffType`](xref:Polly.De
 
 ### Constant
 
-One might think that here we only put into the account the `Delay` property. In reality all above listed properties are used.
+Even though the `Constant` name could imply that only the `Delay` property is used, in reality all the above listed properties are used.
 
 Step 1: Calculating the base delay:
 
 - If `UseJitter` is set to `false` and `Delay` is specified then `Delay` will be used.
 - If `UseJitter` is set to `true` and `Delay` is specified then a random value is added to the `Delay`.
-  - The random value is between [-25% of `Delay`, +25% of `Delay`].
+  - The random value is between -25% of `Delay` and +25% of `Delay`.
 
 Step 2: Capping the delay if needed:
 
--- a/docs/strategies/retry.md
+++ b/docs/strategies/retry.md
@@ -115,7 +115,7 @@ If the `ShouldHandle` predicate returns `true` and the next attempt number is no
 There are many properties that may contribute to this calculation:
 
 - `BackoffType`: Specifies which calculation algorithm should run.
-- `Delay`: If only this property is specified then it will be used as-is. If others are also specified then this will be used as a _base delay_.
+- `Delay`: If only this property is specified then it will be used as-is. If others are also specified then this will be used as a *base delay*.
 - `DelayGenerator`: If specified, will override other property-based calculations, **except** if it returns a negative `TimeSpan`, in which case the other property-based calculations are used.
 - `MaxDelay`: If specified, caps the delay if the calculated delay is greater than this value, **except** if `DelayGenerator` is used, where no capping is applied.
 - `UseJitter`: If enabled, then adds a random value between between -25% of `Delay` and +25% of `Delay`, **except** if `BackoffType` is `Exponential`, where a bit more complex jitter calculation is used.
--- a/docs/strategies/retry.md
+++ b/docs/strategies/retry.md
@@ -188,7 +188,7 @@ The delays column contains an example series of five values to depict the patter
 This algorithm increases the delays for every attempt in an exponential fashion if no jitter is used.
 
 - If `UseJitter` is set to `false` and `Delay` is specified then squaring actual attempt number multiplied by the `Delay` will be used (*`attempt^2 * delay`*).
-- If `UseJitter` is set to `true` and the `Delay` is specified then a `DecorrelatedJitterBackoffV2` formula (based on the [Polly.Contrib.WaitAndRetry](https://github.com/Polly-Contrib/Polly.Contrib.WaitAndRetry)) will be used.
+- If `UseJitter` is set to `true` and the `Delay` is specified then a `DecorrelatedJitterBackoffV2` formula (based on [Polly.Contrib.WaitAndRetry](https://github.com/Polly-Contrib/Polly.Contrib.WaitAndRetry)) will be used.
 
 > [!NOTE]
 > Because the jitter calculation is based on the newly calculated delay, the new delay could be less than the previous value.
--- a/docs/strategies/retry.md
+++ b/docs/strategies/retry.md
@@ -112,7 +112,7 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(optionsExtractDela
 
 If the `ShouldHandle` predicate returns `true` and the next attempt number is not greater than `MaxRetryAttempts` then the retry strategy calculates the next delay.
 
-There are many properties that are somehow contribute to this calculation:
+There are many properties that may contribute to this calculation:
 
 - `Delay`
 - `DelayGenerator`
@@ -121,11 +121,9 @@ There are many properties that are somehow contribute to this calculation:
 - `BackoffType`
 
 > [!IMPORTANT]
-> The below described algorithm is an implementation detail. It might change in the future without further notice.
->
-> Also please bear in mind that some details are not mentioned here to keep the algorithm description short and terse.
+> The summarized description below is an implementation detail. It may change in the future without notice.
 
-The `BackoffType` property's data type is a [`DelayBackOffType`](xref:Polly.DelayBackOffType) enum. This controls primarily how the calculation is done.
+The `BackoffType` property's data type is the [`DelayBackOffType`](xref:Polly.DelayBackOffType) enumeration. This primarily controls how the calculation is done.
 
 ### Constant
 
--- a/docs/strategies/retry.md
+++ b/docs/strategies/retry.md
@@ -133,7 +133,7 @@ Step 1: Calculating the base delay:
 
 - If `UseJitter` is set to `false` and `Delay` is specified then `Delay` will be used.
 - If `UseJitter` is set to `true` and `Delay` is specified then a random value is added to the `Delay`.
-  - The random value is between -25% of `Delay` and +25% of `Delay`.
+  - The random value is between -25% and +25% of `Delay`.
 
 Step 2: Capping the delay if needed:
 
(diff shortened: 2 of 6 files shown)


peter-csala:
I think it might make sense to add a little description to the retry page about how the following properties contribute in the calculation of the next delay: `Delay`, `DelayGenerator`, `MaxDelay`, `UseJitter`, and `BackoffType`

@martintmk Do you agree?

martintmk:
> I think it might make sense to add a little description to the retry page about how the following properties contribute in the calculation of the next delay: `Delay`, `DelayGenerator`, `MaxDelay`, `UseJitter`, and `BackoffType`
> 
> @martintmk, @martincostello Do you agree?

Sure, sounds like a good idea.

martincostello (approved):

martintmk:
Maybe just give description to each property and how it's affects the calculation and then do a sample calculation for one of the backoff types?

docs/strategies/retry.md
```diff
@@ -108,6 +108,55 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(optionsExtractDela
 | `OnRetry`          | `null`                                                                     | Action executed when retry occurs.                                                       |
 | `MaxDelay`         | `null`                                                                     | Caps the calculated retry delay to a specified maximum duration.                         |
 
+## Calculation of the next delay
+
+If the `ShouldHandle` predicate returns `true` and the next attempt number is not greater than `MaxRetryAttempts` then the retry strategy calculates the next delay.
+
+There are many properties that are somehow contribute to this calculation:
```
martincostello, line 115:
```suggestion
There are many properties that may contribute to this calculation:
```

docs/strategies/retry.md
```diff
@@ -108,6 +108,55 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(optionsExtractDela
...
+
+- `Delay`
+- `DelayGenerator`
+- `MaxDelay`
+- `UseJitter`
+- `BackoffType`
+
+> [!IMPORTANT]
+> The below described algorithm is an implementation detail. It might change in the future without further notice.
```
martincostello, line 124:
```suggestion
> The summarized description below is an implementation detail. It may change in the future without notice.
```

docs/strategies/retry.md
```diff
@@ -108,6 +108,55 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(optionsExtractDela
...
+>
+> Also please bear in mind that some details are not mentioned here to keep the algorithm description short and terse.
```
martincostello, line 126: I don't think we need this, I think it's implied by the above.
```suggestion
```

docs/strategies/retry.md
```diff
@@ -108,6 +108,55 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(optionsExtractDela
...
+
+The `BackoffType` property's data type is a [`DelayBackOffType`](xref:Polly.DelayBackOffType) enum. This controls primarily how the calculation is done.
```
martincostello, line 128:
```suggestion
The `BackoffType` property's data type is the [`DelayBackOffType`](xref:Polly.DelayBackOffType) enumeration. This primarily controls how the calculation is done.
```

.github/wordlist.txt
```diff
@@ -13,6 +13,7 @@ deserialization
 dotnet
 dotnetrocks
 durations
+enum
```
martincostello, line 16: If this was only used in the one place I suggested using the full word, then remove this.

```suggestion
```

peter-csala:
> Maybe just give description to each property and how it's affects the calculation and then do a sample calculation for one of the backoff types?

I thought about the same but I soon bumped into the problem that the general description is true only under specific circumstances.

Like: 
- `UseJitter`: If enabled then it adds a random value between between [-25% of `Delay`, +25% of `Delay`]. Except if `Backoff` is set to `Exponential` where jitter is always used.
- `MaxDelay`: If enabled then it caps the delay if the calculated one is greater than this value. Except if you use `DelayGenerator` where no capping is applied.
- etc.

Or shall I create a table with three columns: `Property`, `General use`, `Exception(s)`?

docs/strategies/retry.md
```diff
@@ -108,6 +108,55 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(optionsExtractDela
...
+
+### Constant
+
+One might think that here we only put into the account the `Delay` property. In reality all above listed properties are used.
```
martincostello, line 132: I'm not a fan of the style of wording here - it reads weird here to be making assertions about what the reader might think, rather than just explaining it.

docs/strategies/retry.md
```diff
@@ -108,6 +108,55 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(optionsExtractDela
...
+
+Step 1: Calculating the base delay:
+
+- If `UseJitter` is set to `false` and `Delay` is specified then `Delay` will be used.
+- If `UseJitter` is set to `true` and `Delay` is specified then a random value is added to the `Delay`.
+  - The random value is between [-25% of `Delay`, +25% of `Delay`].
```
martincostello, line 138: Looks weird to me to use the word between but then use range notation in the middle of prose.

docs/strategies/retry.md
peter-csala:
```suggestion
Even though the `Constant` name could imply that only the `Delay` property is used, in reality all the above listed properties are used.
```

Is it better?

docs/strategies/retry.md
peter-csala:
```suggestion
  - The random value is between -25% of `Delay` and +25% of `Delay`.
```
Is it better?

docs/strategies/retry.md
```diff
@@ -114,11 +114,11 @@ If the `ShouldHandle` predicate returns `true` and the next attempt number is no
...
-- `Delay`
-- `DelayGenerator`
-- `MaxDelay`
-- `UseJitter`
-- `BackoffType`
+- `BackoffType`: It specifies which calculation algorithm should run.
+- `Delay`: If only this one is specified then it will be used as is. If others are defined as well then this will be used as a _base delay_.
+- `DelayGenerator`: If specified then it will overwrite other properties based calculation. **Except** if it returns a negative `TimeSpan` where the other properties based calculation is used.
+- `MaxDelay`: If specified then it caps the delay if the calculated one is greater than this value. **Except** if `DelayGenerator` is used where no capping is applied.
+- `UseJitter`: If enabled then it adds a random value between between -25% of `Delay` and +25% of `Delay`. **Except** if `BackoffType` is set to `Exponential` where jitter is always used.
```
martincostello:
```suggestion
- `BackoffType`: Specifies which calculation algorithm should run.
- `Delay`: If only this property is specified then it will be used as-is. If others are also specified then this will be used as a _base delay_.
- `DelayGenerator`: If specified, will override other property-based calculations, **except** if it returns a negative `TimeSpan`, in which case the other property-based calculations are used.
- `MaxDelay`: If specified, caps the delay if the calculated delay is greater than this value, **except** if `DelayGenerator` is used, where no capping is applied.
- `UseJitter`: If enabled, then adds a random value between between -25% of `Delay` and +25% of `Delay`, **except** if `BackoffType` is `Exponential`, where jitter is always used.
```

docs/strategies/retry.md
```diff
@@ -108,6 +108,54 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(optionsExtractDela
...
+
+If the `ShouldHandle` predicate returns `true` and the next attempt number is not greater than `MaxRetryAttempts` then the retry strategy calculates the next delay.
+
+There are many properties that may contribute to this calculation:
+
+- `BackoffType`: It specifies which calculation algorithm should run.
+- `Delay`: If only this one is specified then it will be used as is. If others are defined as well then this will be used as a *base delay*.
+- `DelayGenerator`: If specified then it will overwrite other properties based calculation. **Except** if it returns a negative `TimeSpan` where the other properties based calculation is used.
+- `MaxDelay`: If specified then it caps the delay if the calculated one is greater than this value. **Except** if `DelayGenerator` is used where no capping is applied.
+- `UseJitter`: If enabled then it adds a random value between between -25% of `Delay` and +25% of `Delay`. **Except** if `BackoffType` is set to `Exponential` where jitter is always used.
```
martintmk, line 121:
```suggestion
- `UseJitter`: If enabled then it adds a random value between between -25% of `Delay` and +25% of `Delay`. 
```

Are you sure this is true? Because the exponential delays are different if `UseJitter` is used. Without it the output is not randomized.

https://github.com/App-vNext/Polly/blob/f40fa295a872465c1724abcc52c6ae1f77a6d0a2/src/Polly.Core/Retry/RetryHelper.cs#L69

docs/strategies/retry.md
```diff
@@ -108,6 +108,54 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(optionsExtractDela
...
+## Calculation of the next delay
+
+If the `ShouldHandle` predicate returns `true` and the next attempt number is not greater than `MaxRetryAttempts` then the retry strategy calculates the next delay.
+
+There are many properties that may contribute to this calculation:
+
+- `BackoffType`: It specifies which calculation algorithm should run.
+- `Delay`: If only this one is specified then it will be used as is. If others are defined as well then this will be used as a *base delay*.
+- `DelayGenerator`: If specified then it will overwrite other properties based calculation. **Except** if it returns a negative `TimeSpan` where the other properties based calculation is used.
+- `MaxDelay`: If specified then it caps the delay if the calculated one is greater than this value. **Except** if `DelayGenerator` is used where no capping is applied.
```
martintmk, line 120:
```suggestion
- `MaxDelay`: If specified then it caps the delay if the calculated one is greater than this value. **Except** if `DelayGenerator` is used where no capping is applied.
```

Just wondering, shall we apply the capping to values produced by generator too?

docs/strategies/retry.md
peter-csala: Sorry, may bad. You are right.

docs/strategies/retry.md
peter-csala: I would say yes. That was my intuition before I checked the implementation. But let's deal with that in a separate PR if you don't mind.

docs/strategies/retry.md
```diff
@@ -108,6 +108,54 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(optionsExtractDela
...
+- `DelayGenerator`: If specified, will override other property-based calculations, **except** if it returns a negative `TimeSpan`, in which case the other property-based calculations are used.
+- `MaxDelay`: If specified, caps the delay if the calculated delay is greater than this value, **except** if `DelayGenerator` is used, where no capping is applied.
+- `UseJitter`: If enabled, then adds a random value between between -25% of `Delay` and +25% of `Delay`, **except** if `BackoffType` is `Exponential`, where a bit more complex jitter calculation is used.
+
+> [!IMPORTANT]
+> The summarized description below is an implementation detail. It may change in the future without notice.
+
+The `BackoffType` property's data type is the [`DelayBackOffType`](xref:Polly.DelayBackOffType) enumeration. This primarily controls how the calculation is done.
+
+### Constant
```
martintmk, line 128:
```suggestion
### Example calculation

Bellow is the example calculation for `Constant` delay type. It showcases how the delay is calculated, the calculation is almost identical for the other (`Linear`, `Exponential`) backoff types.
```

Or something in that sense.

docs/strategies/retry.md
martintmk: My intuition says yes too, the less special cases, the better.

docs/strategies/retry.md
```diff
@@ -108,6 +108,54 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(optionsExtractDela
...
 
+## Calculation of the next delay
+
+If the `ShouldHandle` predicate returns `true` and the next attempt number is not greater than `MaxRetryAttempts` then the retry strategy calculates the next delay.
+
+There are many properties that may contribute to this calculation:
+
+- `BackoffType`: It specifies which calculation algorithm should run.
+- `Delay`: If only this one is specified then it will be used as is. If others are defined as well then this will be used as a *base delay*.
+- `DelayGenerator`: If specified then it will overwrite other properties based calculation. **Except** if it returns a negative `TimeSpan` where the other properties based calculation is used.
```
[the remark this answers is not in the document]
martintmk: Also, `null` can be also returned. In that case the calculated delay is used too:

https://github.com/dotnet/extensions/blob/9f25d9d74674483298cd167cf8ff6e9d9026bf20/src/Libraries/Microsoft.Extensions.Http.Resilience/Polly/HttpRetryStrategyOptions.cs#L58

docs/strategies/retry.md
martincostello: I would agree we should apply it - `Retry-After` isn't necessarily authoritative and the user may wish to try earlier by overriding. If we do that we should make it clear in the documentation.

docs/strategies/retry.md
```diff
@@ -150,6 +150,19 @@ Step 3: Using the generator if supplied
 
 ### Linear
 
+This algorithm increases the delays for every attempt in a linear fashion if no jitter is used.
+
+Step 1: Calculating the base delay:
+
+- If `UseJitter` is set to `false` and `Delay` is specified then `Delay` multiplied by the actual attempt number will be used.
+- If `UseJitter` is set to `true` and `Delay` is specified then a random value is added to the `Delay` multiplied by the actual attempt number.
+  - The random value is between -25% of the newly calculated `Delay` and +25% of the newly calculated `Delay`.
```
martincostello, line 159:
```suggestion
  - The random value is between -25% and +25% of the newly calculated `Delay`.
```

docs/strategies/retry.md
```diff
@@ -150,6 +150,19 @@ Step 3: Using the generator if supplied
...
+
+> [!NOTE]
+> Because the jitter calculation is based on the newly calculated delay that's why the new delay could be smaller than the previous one.
```
martincostello, line 162:
```suggestion
> Because the jitter calculation is based on the newly calculated delay, the new delay could be less than the previous value.
```

docs/strategies/retry.md
```diff
@@ -156,15 +156,25 @@ Step 1: Calculating the base delay:
...
-> Because the jitter calculation is based on the newly calculated delay that's why the new delay could be smaller than the previous one.
+> Because the jitter calculation is based on the newly calculated delay, the new delay could be less than the previous value.
 
 Step 2 and 3 are the same as for the Constant algorithm.
 
 ### Exponential
 
+This algorithm increases the delays for every attempt in an exponential fashion if no jitter is used.
+
+- If `UseJitter` is set to `false` and `Delay` is specified then squaring actual attempt number multiplied by the `Delay` will be used.
```
martincostello, line 170: Maybe we should add something like `attempt^2 * delay` to these bits of text to make the meaning a bit clearer in addition to the text?

docs/strategies/retry.md
```diff
@@ -156,15 +156,25 @@ Step 1: Calculating the base delay:
...
+- If `UseJitter` is set to `true` and the `Delay` is specified then a `DecorrelatedJitterBackoffV2` formula (based on the [Polly.Contrib.WaitAndRetry](https://github.com/Polly-Contrib/Polly.Contrib.WaitAndRetry)) will be used.
```
martincostello, line 171:
```suggestion
- If `UseJitter` is set to `true` and the `Delay` is specified then a `DecorrelatedJitterBackoffV2` formula (based on [Polly.Contrib.WaitAndRetry](https://github.com/Polly-Contrib/Polly.Contrib.WaitAndRetry)) will be used.
```

docs/strategies/retry.md
```diff
@@ -108,6 +108,109 @@ new ResiliencePipelineBuilder<HttpResponseMessage>().AddRetry(optionsExtractDela
...
+- `BackoffType`: Specifies which calculation algorithm should run.
+- `Delay`: If only this property is specified then it will be used as-is. If others are also specified then this will be used as a *base delay*.
+- `DelayGenerator`: If specified, will override other property-based calculations, **except** if it returns a negative `TimeSpan`, in which case the other property-based calculations are used.
+- `MaxDelay`: If specified, caps the delay if the calculated delay is greater than this value, **except** if `DelayGenerator` is used, where no capping is applied.
+- `UseJitter`: If enabled, then adds a random value between between -25% of `Delay` and +25% of `Delay`, **except** if `BackoffType` is `Exponential`, where a bit more complex jitter calculation is used.
+
+> [!IMPORTANT]
+> The summarized description below is an implementation detail. It may change in the future without notice.
+
+The `BackoffType` property's data type is the [`DelayBackOffType`](xref:Polly.DelayBackOffType) enumeration. This primarily controls how the calculation is done.
```
peter-csala, line 126: @martincostello , @martintmk 

Do you have any idea why this link is broken?
`[`DelayBackOffType`](xref:Polly.DelayBackOffType)`

I got this warning when I run docfx locally:
> Polly/docs/strategies/retry.md: warning UidNotFound: 1 invalid cross reference(s) "Polly.DelayBackOffType".

martintmk: Should be `Polly.DelayBackoffType` (lower-case o)

martintmk (approved):

docs/strategies/retry.md
peter-csala: It is a concision decision by the strategy creator that (s)he sets the `MaxDelay`. If it is set then the strategy should cap the value regardless how it is calculated/retrieved IMHO.

martincostello (approved):

PR #1935, opened 2024-01-27
Fix S6608/IDE1006/SA1414/CA1508 warnings in Polly.Specs project

baranyaimate:

Contributes to [https://github.com/App-vNext/Polly/issues/1290](https://github.com/App-vNext/Polly/issues/1290)

Fix S6608/IDE1006/SA1414/CA1508 warnings in Polly.Specs project

## Confirm the following


test/Polly.Specs/RateLimit/RateLimitPolicySpecsBase.cs
```diff
@@ -11,7 +11,7 @@ public abstract class RateLimitPolicySpecsBase : RateLimitSpecsBase
         TimeSpan perTimeSpan,
         int maxBurst);
 
-    protected abstract (bool, TimeSpan) TryExecuteThroughPolicy(IRateLimitPolicy policy);
+    protected abstract (bool permitExecution, TimeSpan retryAfter) TryExecuteThroughPolicy(IRateLimitPolicy policy);
```
martincostello, line 14: We might currently have the rule for this off, but there's another rule that requires named tuples to use PascalCase (so `PermitExecution` and `RetryAfter`), so as you're adding these new it would be better to name them that way now so we don't need to rename them later.

baranyaimate: If I use PascaleCase I get a warning about the SA1316 rule.

martincostello: The documentation for the rule states that PascalCase is what should be used: [SA1316](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/f5f4ca9468fc29a4d34d50541afb2e496933feda/documentation/SA1316.md?plain=1#L29-L49)

Looks like the configuration is overriding the default here: https://github.com/App-vNext/Polly/blob/0064b4de2da37cc98400553834ae6a990625496b/eng/analyzers/Stylecop.json#L22

Can you remove that override as I think Pascal Case is what we should be using, provided that that change doesn't suddenly cause lots of new warnings.

martincostello (approved):
Thanks!

PR #1947, opened 2024-02-03
Fix CA2000/stalled suppressions

gintsk:
# Pull Request

Contributes to #1290

Fix CA2000 in `Polly.Specs` tests
Removed stalled suppressions

## Confirm the following

### DIFF
--- a/test/Polly.Specs/Caching/CacheAsyncSpecs.cs
+++ b/test/Polly.Specs/Caching/CacheAsyncSpecs.cs
@@ -450,24 +450,27 @@ public class CacheAsyncSpecs : IDisposable
 
         var cache = Policy.CacheAsync(new StubCacheProvider(), TimeSpan.MaxValue);
 
-        CancellationTokenSource tokenSource = new CancellationTokenSource();
-
         int delegateInvocations = 0;
-        Func<Context, CancellationToken, Task<string>> func = async (_, _) =>
+
+        using (CancellationTokenSource tokenSource = new CancellationTokenSource())
         {
-            // delegate does not observe cancellation token; test is whether CacheEngine does.
-            delegateInvocations++;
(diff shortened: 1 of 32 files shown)


test/Polly.Specs/Caching/CacheAsyncSpecs.cs
```diff
@@ -450,24 +450,27 @@ public async Task Should_honour_cancellation_even_if_prior_execution_has_cached(
 
         var cache = Policy.CacheAsync(new StubCacheProvider(), TimeSpan.MaxValue);
 
-        CancellationTokenSource tokenSource = new CancellationTokenSource();
-
         int delegateInvocations = 0;
-        Func<Context, CancellationToken, Task<string>> func = async (_, _) =>
+
+        using (CancellationTokenSource tokenSource = new CancellationTokenSource())
```
martincostello, line 455: I think it would be less noisy in the diff to use the syntax that doesn't need a block scope wherever possible, so:

```csharp
using CancellationTokenSource tokenSource = ...
...

In cases like this, `var` would also be a welcome improvement.

Comment (gintsk) [reply test/Polly.Specs/Caching/CacheAsyncSpecs.cs:455]: For tests testing cancellation scenarios, I used block scope, to explicitly indicate cancellation token usage limits (it is the same style as in other existing tests, i.e. `Should_execute_user_delegate_without_adding_extra_cancellation_behaviour` and many other)
```
//setup
...
using (CancellationTokenSource cts = new CancellationTokenSource())
{
...
}
//asserts
...
```

I didn't use `var` because it didn't match the existing style. But I can correct it to `var` and unify CancellationTokenSource variable naming to cts (in tests)
```
using (var cts = new CancellationTokenSource())
```

martincostello (approved):

PR #1953, opened 2024-02-05
Use var

martincostello:
Some more `var` usage missed from #1950.


gintsk:
Maybe it makes sense to use consistent variable naming within `Polly.Specs` ?
Changing these variations `var tokenSource = new CancellationTokenSource()`, `var cts = new CancellationTokenSource()`, `var cancellationSource = new CancellationTokenSource()`
to
`var cancellationTokenSource = new CancellationTokenSource()`

martincostello:
I considered that, but some also have custom names and it was just a quick drive-by change while I was preparing the next release, so I'm going to leave them as they are for the time being.

PR #1963, opened 2024-02-08
Fix CA1806

gintsk:
# Pull Request

Contributes to #1290

Formal fix for CA1806: Do not ignore method results

## Confirm the following

### DIFF
--- a/test/Polly.Specs/Caching/AbsoluteTtlSpecs.cs
+++ b/test/Polly.Specs/Caching/AbsoluteTtlSpecs.cs
@@ -6,10 +6,7 @@ public class AbsoluteTtlSpecs : IDisposable
     [Fact]
     public void Should_be_able_to_configure_for_near_future_time()
     {
-        Action configure = () =>
-        {
-            _ = new AbsoluteTtl(DateTimeOffset.UtcNow.Date.AddDays(1));
-        };
+        Action configure = () => _ = new AbsoluteTtl(DateTimeOffset.UtcNow.Date.AddDays(1));
 
         configure.Should().NotThrow();
     }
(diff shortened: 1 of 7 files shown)


test/Polly.Specs/Caching/AbsoluteTtlSpecs.cs
```diff
@@ -6,23 +6,32 @@ public class AbsoluteTtlSpecs : IDisposable
     [Fact]
     public void Should_be_able_to_configure_for_near_future_time()
     {
-        Action configure = () => new AbsoluteTtl(DateTimeOffset.UtcNow.Date.AddDays(1));
```
martincostello, line 9: An easier way to fix all of these issues without also changing all the formatting is to assign them to the discard variable, i.e. ` _ = new Foo()` etc.

martincostello: I meant more like this, assuming it compiles:

```csharp
Action configure = () => _ = new AbsoluteTtl(DateTimeOffset.UtcNow.Date.AddDays(1));
```

Then the diff is mostly just adding a bunch of `_ = `, than also changing the lines, adding braces etc.

gintsk: Agree. This will be cleaner.

martincostello (approved):

martincostello:
Thanks again!

test/Polly.Specs/Caching/AbsoluteTtlSpecs.cs
jnyrup: Or if you permit using `var` and C#10, we can let the compiler infer the return type to be `Func<AbsoluteTtl>` which also silences CA1806 as I recall.

```cs
var configure = () => new AbsoluteTtl(DateTimeOffset.UtcNow.Date.AddDays(1));
```

PR #1975, opened 2024-02-19
[Docs] Improve diagrams for hedging cancellation

martintmk:

The cancellation behavior was not fully described, this PR fixes this:

image

## Confirm the following

### DIFF
--- a/docs/strategies/timeout.md
+++ b/docs/strategies/timeout.md
@@ -165,11 +165,18 @@ sequenceDiagram
     D-->>D: Performs <br/>long-running <br/>operation
     T-->>T: Times out
     deactivate T
+
     T->>D: Propagates cancellation
+    D-->>D: Cancellation of callback
+    D->>T: Cancellation finished
     deactivate D
+
     T->>P: Throws <br/>TimeoutRejectedException
     P->>C: Propagates exception
 ```
+
+> [!NOTE]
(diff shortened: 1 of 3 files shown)


martincostello (approved):

peter-csala:
The same is true for here:
- https://github.com/App-vNext/Polly/blob/main/docs/strategies/timeout.md?plain=1#L168
- https://github.com/App-vNext/Polly/blob/main/docs/pipelines/index.md?plain=1#L283

docs/strategies/timeout.md
```diff
@@ -165,12 +165,19 @@ sequenceDiagram
...
+    D-->>D: Cancellation of callback
+    D->>T: Cancellation finished
     deactivate D
+
     T->>P: Throws <br/>TimeoutRejectedException
     P->>C: Propagates exception
 ```
 
+> [!NOTE]
+> Notice that the timeout waits until the callback  is cancelled before throwing `TimeoutRejectedException`. Therefore it's important for the callbacks to respect the cancellation token passed to the execution. If the cancellation token is not correctly respected, the timeout is unnecessarily delayed.
```
peter-csala, line 179:
```suggestion
> Notice that the timeout waits until the callback is cancelled before throwing `TimeoutRejectedException`. Therefore it's important for the callbacks to respect the cancellation token passed to the execution. If the cancellation token is not correctly respected, the timeout is unnecessarily delayed.
```

peter-csala: Just wondering: If we use `Therefore it's important` wording inside the note then would it make sense to use `[!IMPORTANT]` instead?

martintmk: Sure makes sense, changed to important.

PR #1994, opened 2024-02-25
Document outcome strategy anti-patterns

vany0114:
# Pull Request

* Document outcome strategy anti-patterns
* Deprecates `AddException` API in the `OutcomeGenerator`

As per our discussion [here](https://github.com/App-vNext/Polly/discussions/1978) we decided to document why injecting exceptions via the outcome strategy is an anti-pattern (in advanced scenarios when the usage of delegates is required):
* side effects on telemetry events, which might affect instrumentation and metrics
* goes against controlled chaos injection

As part of this change, I marked the `AddException` API in the `OutcomeGenerator` as deprecated to be consistent with what we're preaching, it does not make sense to say it is an anti-pattern and at the same time provide a built-in way to do so + if we expose an API to inject an exception as an outcome why have the fault strategy at all? why consumers would use it? or when to use one or another? it ends up creating confusion. 

## Confirm the following


src/Polly.Core/Simmy/Outcomes/OutcomeGenerator.cs
```diff
@@ -35,6 +35,9 @@ internal OutcomeGenerator(Func<int, int> weightGenerator)
     /// <param name="generator">The delegate that generates the exception.</param>
     /// <param name="weight">The weight assigned to this generator. Defaults to <c>100</c>.</param>
     /// <returns>The current instance of <see cref="OutcomeGenerator{TResult}"/>.</returns>
+#pragma warning disable S1133 // Deprecated code should be removed
+    [Obsolete("This method is deprecated and will be removed in the next version. Use Chaos fault strategy instead.")]
+#pragma warning restore S1133 // Deprecated code should be removed
     public OutcomeGenerator<TResult> AddException(Func<Exception> generator, int weight = DefaultWeight)
```
martintmk, line 41: Why obsolete this though? I consider this still a valid pattern. Imagine a very common scenario where we need to inject both types of faults `HttpResponseMessage` with 500 and `HttpRequestException`.

Current chaos outcome API is powerful enough to handle both cases using a single strategy. My thinking for anti-patters was using chaos outcome to inject exceptions only and no results. In that case, the chaos fault is a better option.

vany0114: The scenario you're talking about makes a lot of sense, but still, both have undesirable side effects:  telemetry and control. We could certainly modify the outcome strategy to report the exception in a different way, but to me we're mixing two different things, exceptions and "faulted" results, since the last one isn't necessary an exception.

On the other hand the biggest downside of that approach is the consumer is losing control, in the real world you usually have different configurations to inject different kinds of chaos, it does not feel right treating outcomes the same as exceptions, so imagine you have a 10% as injection rate for outcomes and 1% as injection rate for exceptions, if we allow injecting exceptions via the outcome strategy it will completely disregard the fault configuration.

martintmk: My biggest problem here is obsoleting API that was introduced just one version ago. In Polly v8, the `Outcome` can represents both results and exceptions. Even the signature of chaos outcome allows this:

https://github.com/App-vNext/Polly/blob/de6eb20fd8b808a3f821fc2ab1eef2eeb0b99fa7/src/Polly.Core/Simmy/Outcomes/ChaosOutcomeStrategyOptions.cs#L31C4-L31C104

So at this point, I gravity toward not breaking anything. If someone has a scenario where they need both results and exceptions injected in a single strategy, let's not prevent them. (we actually have some teams that are doing this)
Let's not forget that every additional strategy adds a slight overhead, so having the option to express results and exceptions with a single strategy is beneficial.

>  it does not feel right treating outcomes the same as exceptions, so imagine you have a 10% as injection rate for outcomes and 1% as injection rate for exceptions, if we allow injecting exceptions via the outcome strategy it will completely disregard the fault configuration.

The `OutcomeGenerator` API is flexible enough to allow assigning weight to an exception, so the above can still be expressed.

``` csharp
new OutcomeGenerator<HttpResponseMessage>()
    .AddResult(() => new HttpResponseMessage(System.Net.HttpStatusCode.InternalServerError), weight: 50)
    .AddException(() => new HttpRequestException(), weight: 2);
```

martincostello: I agree that obsoleting something immediately doesn't look great. If we did, it wouldn't be in the _next_ version, but a _future_ version.

vany0114: >Even the signature of chaos outcome allows this:
> 
> https://github.com/App-vNext/Polly/blob/de6eb20fd8b808a3f821fc2ab1eef2eeb0b99fa7/src/Polly.Core/Simmy/Outcomes/ChaosOutcomeStrategyOptions.cs#L31C4-L31C104

Correct, that's why the documentation states the antipattern may occur in advanced scenarios when the use of generator delegates is required.

> The OutcomeGenerator API is flexible enough to allow assigning weight to an exception, so the above can still be expressed.

Yeah partially I guess, because that's still injecting an exception using the outcome configuration, how about if the "knob" to inject exceptions is off? As I said in the docs this is very opinionated, but it's good having the discussion so that we can enrich the docs. And to be clear, I'm not against the AddException API, it's just that I feel we need to be cautious with that and not encourage people to use it w/o knowing the caveats, I decided to obsolete it based on the premise that treating faults as outcomes is an antipattern because of the reasons I'm explaining in this PR, mostly because of:
* side effects on telemetry events, which might affect instrumentation and metrics
* goes against controlled chaos injection

Said that, other than documenting the scenarios we can also document that method using the `remarks` or something like that.

docs/chaos/outcome.md
```diff
@@ -161,17 +161,151 @@ new ResiliencePipelineBuilder<HttpResponseMessage>()
...
+```
+<!-- endSnippet -->
+
+## Anti-patterns
+
+### Injecting faults (exceptions)
+
+❌ DON'T
+
+Use outcome strategies to inject faults in advanced scenarios which you need to inject outcomes using delegates. This is an opinionated anti-pattern since you can consider an exception as a result/outcome, however, there might be undesired implications when doing so, one of them is the telemetry events, which might end up affecting your metrics as the `ChaosOutcomeStrategy` reports both result and exceptions in the same way, and this could pose a problem for instrumentation purposes since it's clearer looking for fault injected events to be 100% sure where/when exceptions were injected, rather than have them mixed in the same "bag".
```
martincostello, line 184: Otherwise it's a very long sentence.

```suggestion
Use outcome strategies to inject faults in advanced scenarios which you need to inject outcomes using delegates. This is an opinionated anti-pattern since you can consider an exception as a result/outcome, however, there might be undesired implications when doing so, one of them is the telemetry events, which might end up affecting your metrics as the `ChaosOutcomeStrategy` reports both result and exceptions in the same way. This could pose a problem for instrumentation purposes since it's clearer looking for fault injected events to be 100% sure where/when exceptions were injected, rather than have them mixed in the same "bag".
```

docs/chaos/outcome.md
```diff
@@ -161,17 +161,151 @@ new ResiliencePipelineBuilder<HttpResponseMessage>()
...
+
+Also, you end up losing control of how/when to inject outcomes vs faults since this way does not allow you to control separately when to inject a fault vs an outcome.
```
martincostello, line 186:
```suggestion
Also, you end up losing control of how/when to inject outcomes vs. faults since this way does not allow you to separately control when to inject a fault vs. an outcome.
```

docs/chaos/outcome.md
```diff
@@ -161,17 +161,151 @@ new ResiliencePipelineBuilder<HttpResponseMessage>()
...
+
+<!-- snippet: chaos-outcome-anti-pattern-generator-inject-fault -->
+```cs
+var pipeline = new ResiliencePipelineBuilder<HttpResponseMessage>()
+    .AddChaosOutcome(new ChaosOutcomeStrategyOptions<HttpResponseMessage>
+    {
+        InjectionRate = 0.5, // same injection rate for both fault and outcome
```
martincostello, line 193:
```suggestion
        InjectionRate = 0.5, // Same injection rate for both fault and outcome
```

docs/chaos/outcome.md
```diff
@@ -161,17 +161,151 @@ new ResiliencePipelineBuilder<HttpResponseMessage>()
...
+        OutcomeGenerator = args =>
```
martincostello, line 194: Looks like this could be made `static`?

docs/chaos/outcome.md
```diff
@@ -161,17 +161,151 @@ new ResiliencePipelineBuilder<HttpResponseMessage>()
...
+            return default;
+        }
+    })
+    .Build();
+```
+<!-- endSnippet -->
+
+✅ DO
+
+The previous approach is tempting since it looks more succinct, but use the fault chaos instead as the [`ChaosFaultStrategy`](fault.md) correctly tracks telemetry events effectively as faults not just as any other outcome, also by separating them you can control the injection rate and enable/disable them separately which gives you more control when it comes to injecting chaos dynamically and in a controlled manner.
```
martincostello, line 228:
```suggestion
The previous approach is tempting since it looks more succinct, but use fault chaos instead as the [`ChaosFaultStrategy`](fault.md) correctly tracks telemetry events effectively as faults not just as any other outcome. Also by separating them, you can control the injection rate and enable/disable them separately which gives you more control when it comes to injecting chaos dynamically and in a controlled manner.
```

docs/chaos/outcome.md
```diff
@@ -161,17 +161,151 @@ new ResiliencePipelineBuilder<HttpResponseMessage>()
...
+
+<!-- snippet: chaos-outcome-pattern-generator-inject-fault -->
+```cs
+var randomThreshold = Random.Shared.Next(350);
+var pipeline = new ResiliencePipelineBuilder<HttpResponseMessage>()
+    .AddChaosFault(new ChaosFaultStrategyOptions
+    {
+        InjectionRate = 0.1, // different injection rate for faults
+        EnabledGenerator = args => ValueTask.FromResult(ShouldEnableFaults(args.Context)), // different settings might apply to inject faults
```
martincostello, line 237:
```suggestion
        InjectionRate = 0.1, // Different injection rate for faults
        EnabledGenerator = args => ValueTask.FromResult(ShouldEnableFaults(args.Context)), // Different settings might apply to inject faults
```

docs/chaos/outcome.md
```diff
@@ -161,17 +161,151 @@ new ResiliencePipelineBuilder<HttpResponseMessage>()
...
+        OnFaultInjected = static args =>
+        {
+            Console.WriteLine($"OnFaultInjected, Exception: {args.Fault.Message}, Operation: {args.Context.OperationKey}.");
+            return default;
+        }
+    })
+    .AddChaosOutcome(new ChaosOutcomeStrategyOptions<HttpResponseMessage>
+    {
+        InjectionRate = 0.5, // different injection rate for outcomes
+        EnabledGenerator = args => ValueTask.FromResult(ShouldEnableOutcome(args.Context)), // different settings might apply to inject outcomes
```
martincostello, line 257:
```suggestion
        EnabledGenerator = args => ValueTask.FromResult(ShouldEnableOutcome(args.Context)), // Different settings might apply to inject outcomes
```

docs/chaos/outcome.md
```diff
@@ -161,17 +161,151 @@ new ResiliencePipelineBuilder<HttpResponseMessage>()
...
+    {
+        OutcomeGenerator = new OutcomeGenerator<HttpResponseMessage>()
+            .AddException<HttpRequestException>(), // ⚠️ Avoid this ⚠️ 
     });
 ```
 <!-- endSnippet -->
+
+✅ DO
+
+Use fault strategies to properly inject the exception.
```
martincostello, line 297:
```suggestion
Use fault strategies to inject the exception.
```

docs/chaos/outcome.md
```diff
@@ -161,17 +161,151 @@ new ResiliencePipelineBuilder<HttpResponseMessage>()
...
+    .AddChaosFault(new ChaosFaultStrategyOptions
+    {
+        FaultGenerator = new FaultGenerator()
+            .AddException<HttpRequestException>(),
+    });
+```
+<!-- endSnippet -->
+
+> [!NOTE]
+> Even though the outcome strategy is flexible enough to allow you to inject outcomes as well as exceptions without the need to chain a fault strategy in the pipeline, use your judgment when doing so because of the caveats and side effects explained before around the telemetry and injection control.
```
martincostello, line 311:
```suggestion
> Even though the outcome strategy is flexible enough to allow you to inject outcomes as well as exceptions without the need to chain a fault strategy in the pipeline, use your judgment when doing so because of the caveats and side effects explained regarding telemetry and injection control.
```

martincostello (approved):

src/Snippets/Docs/Chaos.Outcome.cs
```diff
@@ -105,15 +106,138 @@ public static void OutcomeGeneratorDelegates()
...
+                EnabledGenerator = static args => ValueTask.FromResult(ShouldEnableFaults(args.Context)), // Different settings might apply to inject faults
+                FaultGenerator = args =>
+                {
+                    Exception? exception = randomThreshold switch
+                    {
+                        >= 250 and < 350 => new HttpRequestException("Chaos request exception."),
                         _ => null
                     };
 
+                    return new ValueTask<Exception?>(exception);
```
peter-csala, line 176: It is a minor thing but from my point of view we should have samples with consistent API usage.

Current we have:
```cs
return new ValueTask<Exception?>(exception); //inside FaultGenerator
...
return ValueTask.FromResult(outcome); //inside OutcomeGenerator
```

I would suggest to use either the ctor in both places or the `FromXYZ` methods.

src/Snippets/Docs/Chaos.Outcome.cs
```diff
@@ -105,15 +106,138 @@ public static void OutcomeGeneratorDelegates()
...
+                    return default;
+                }
+            })
+            .AddChaosOutcome(new ChaosOutcomeStrategyOptions<HttpResponseMessage>
+            {
+                InjectionRate = 0.5, // Different injection rate for outcomes
+                EnabledGenerator = static args => ValueTask.FromResult(ShouldEnableOutcome(args.Context)), // Different settings might apply to inject outcomes
+                OutcomeGenerator = args =>
+                {
+                    Outcome<HttpResponseMessage>? outcome = randomThreshold switch
```
peter-csala, line 190: In order to reduce repetitive code the `switch` expression should either return a `HttpStatusCode` or an `HttpResponseMessage`. Something like this:

```cs
HttpStatusCode statusCode = randomThreshold switch
{
    < 100 => HttpStatusCode.InternalServerError,
    < 150 => HttpStatusCode.TooManyRequests,
    < 250 => CalculateStatusCodeFromContext(args.Context),
    _ => HttpStatusCode.OK
};

var outcome = Outcome.FromResult(new HttpResponseMessage(statusCode));
return ValueTask.FromResult(outcome);
```

src/Snippets/Docs/Chaos.Outcome.cs
```diff
@@ -105,15 +106,138 @@ public static void OutcomeGeneratorDelegates()
...
+
+    public static void Pattern_GeneratorDelegateInjectFaultAndOutcome()
+    {
+        #region chaos-outcome-pattern-generator-inject-fault
+        var randomThreshold = Random.Shared.Next(350);
+        var pipeline = new ResiliencePipelineBuilder<HttpResponseMessage>()
+            .AddChaosFault(new ChaosFaultStrategyOptions
+            {
+                InjectionRate = 0.1, // Different injection rate for faults
+                EnabledGenerator = static args => ValueTask.FromResult(ShouldEnableFaults(args.Context)), // Different settings might apply to inject faults
```
peter-csala, line 167: I would suggest to change the signature of `ShouldEnableFaults` to return a `ValueTask<bool>` to streamline the example. I would suggest the same for `ShouldEnableOutcome`.

```cs
EnabledGenerator = static args => ShouldEnableFaults(args.Context), // Different settings might apply to inject faults
```

src/Snippets/Docs/Chaos.Outcome.cs
```diff
@@ -105,15 +106,138 @@ public static void OutcomeGeneratorDelegates()
...
+            })
+            .Build();
+
+        #endregion
+    }
+
+    public static void Pattern_GeneratorDelegateInjectFaultAndOutcome()
+    {
+        #region chaos-outcome-pattern-generator-inject-fault
+        var randomThreshold = Random.Shared.Next(350);
```
peter-csala, line 162: If we would move the `randomThreshold` to class level then we could mark it as `static`. That would allow us to use static anonymous functions for `FaultGenerator` and `OutcomeGenerator`.

docs/chaos/outcome.md
```diff
@@ -161,17 +161,151 @@ new ResiliencePipelineBuilder<HttpResponseMessage>()
...
+
+Use outcome strategies to inject only faults, use the [`ChaosFaultStrategy`](fault.md) instead.
+
+<!-- snippet: chaos-outcome-anti-pattern-only-inject-fault -->
+```cs
+new ResiliencePipelineBuilder<HttpResponseMessage>()
+    .AddChaosOutcome(new ChaosOutcomeStrategyOptions<HttpResponseMessage>
+    {
+        OutcomeGenerator = new OutcomeGenerator<HttpResponseMessage>()
+            .AddException<HttpRequestException>(), // ⚠️ Avoid this ⚠️
```
peter-csala, line 290: Please do **NOT** edit the snippets inside the markdown file.

It will be overwritten the very next time when someone runs the `dotnet mdsnippets` command.

I suggest to move the `// ⚠️ Avoid this ⚠️` comments into the source code as well.

peter-csala (changes requested):

docs/chaos/outcome.md
vany0114: Nice! I didn't know

vany0114:
Hi @martincostello @peter-csala I just addressed the rest of the feedback, please take a look again when you guys have the chance.

src/Snippets/Docs/Chaos.Outcome.cs
```diff
@@ -82,7 +85,7 @@ public static void OutcomeGenerator()
                 OutcomeGenerator = new OutcomeGenerator<HttpResponseMessage>()
                     .AddResult(() => new HttpResponseMessage(HttpStatusCode.InternalServerError)) // Result generator
                     .AddResult(() => new HttpResponseMessage(HttpStatusCode.TooManyRequests), weight: 50) // Result generator with weight
-                    .AddResult(context => CreateResultFromContext(context)) // Access the ResilienceContext to create result
+                    .AddResult(context => new HttpResponseMessage(CreateResultFromContext(context))) // Access the ResilienceContext to create result
```
peter-csala, line 88: The usage samples advocate to use one of the anti-patterns

```cs
...
.AddResult(context => new HttpResponseMessage(CreateResultFromContext(context))) // Access the ResilienceContext to create result
.AddException<HttpRequestException>(), // You can also register exceptions
```

Please update these `OutcomeGenerator`s by removing the`AddException` clauses.

src/Snippets/Docs/Chaos.Outcome.cs
```diff
@@ -103,17 +106,139 @@ public static void OutcomeGeneratorDelegates()
                     {
                         < 100 => Outcome.FromResult(new HttpResponseMessage(HttpStatusCode.InternalServerError)),
                         < 150 => Outcome.FromResult(new HttpResponseMessage(HttpStatusCode.TooManyRequests)),
-                        < 250 => Outcome.FromResult(CreateResultFromContext(args.Context)),
+                        < 250 => Outcome.FromResult(new HttpResponseMessage(CreateResultFromContext(args.Context))),
```
peter-csala, line 109: As far as I can see this `OutcomeGenerator` delegate could be marked as `static`

src/Snippets/Docs/Chaos.Outcome.cs
vany0114: When I originally opened the PR I had removed those guys since I had deprecated it, however, I put them back so that we support that scenario (injecting results and exceptions in the same strategy) but under their judgment after knowing the caveats/side effects explained in the anti-pattern sections and in the footer. I can remove that entirely from the docs but it would be hidden. Honestly, while I understand it is handy it feels like we're kinda discouraging people to use the right strategy to inject faults. 

We could also add a comment in the samples as I did in some snippets, like this: // ⚠️ Avoid this ⚠️ 
and maybe add a `remarks` on the API pointing to the docs? thoughts?

vany0114:
@martincostello I think this PR is ready to go BTW.

martincostello:
Yep, I'm just AFK a lot at the moment and haven't had the chance to re-review properly. It's in my todo queue.

martincostello (approved):

PR #2003, opened 2024-03-01
Update to .NET 9 SDK

martincostello:
A long-lived PR for testing .NET 9 until the stable release in November 2024. `net9.0` is explicitly _not_ being added as a new TFM for the packages we ship to NuGet.org.

Changes include:

- Update to the .NET 9 SDK.
- Add usage of `[DebuggerDisableUserUnhandledExceptions]` to avoid newer versions of the Visual Studio debugger for breaking for exceptions on async code paths we are intentionally handling.
- Target `net8.0` and `net9.0` in tests, samples and benchmarks.
- Remove `net6.0` target from tests, samples and benchmarks as it will both be out-of-support by November 2024.
- Fix/suppress new IDE0022/S3236 warnings.
- Only enable NuGet Audit for direct dependencies, not transient ones.

### DIFF
--- a/test/Polly.Core.Tests/Polly.Core.Tests.csproj
+++ b/test/Polly.Core.Tests/Polly.Core.Tests.csproj
@@ -6,8 +6,6 @@
     <ProjectType>Test</ProjectType>
     <Nullable>enable</Nullable>
     <Threshold>100</Threshold>
-    <!-- TODO Set back to 100 when failing tests re-enabled -->
-    <Threshold>99</Threshold>
     <NoWarn>$(NoWarn);S6966;SA1600;SA1204</NoWarn>
     <Include>[Polly.Core]*</Include>
     <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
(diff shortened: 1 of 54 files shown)


test/Polly.Core.Tests/Polly.Core.Tests.csproj
```diff
@@ -1,16 +1,17 @@
...
     <Nullable>enable</Nullable>
-    <Threshold>100</Threshold>
     <NoWarn>$(NoWarn);SA1600;SA1204</NoWarn>
     <Include>[Polly.Core]*</Include>
     <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
   </PropertyGroup>
 
   <ItemGroup>
+    <!-- HACK net9.0 compilation for the tests fails due to missing Task.Delay() extension for TimeProvider without this reference -->
+    <PackageReference Include="Microsoft.Bcl.TimeProvider" Condition="$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'net9.0'))" />
```
martincostello, line 14: @martintmk Any idea what's going on with this? Without an explicit reference to this package, I can't get the tests to compile for `net9.0`.

The only change I can see that might explain it is [this](https://github.com/dotnet/extensions/commit/16b40667278d24a23af0c839f4c277499b3fee19#diff-11865412b98c8eebb41fa5cd1c55c154b9b58ca5e7dbe41bd651124bdc5e1b9f) from https://github.com/dotnet/extensions/pull/5058.

PR #2032, opened 2024-03-26
Clearing out the only 1805 violation using the method listed here: ht…

cbredlow:
…tps://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1805#how-to-fix-violations

# Pull Request

Addressing the CA1805 as a part of [issue 1290](https://github.com/App-vNext/Polly/issues/1290)

Removed the NoWarn for CA1805, and changes the spinner to be static and have a lambda for the default value, as suggested here: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1805#how-to-fix-violations
## Confirm the following

### DIFF
--- a/src/Polly/RateLimit/LockFreeTokenBucketRateLimiter.cs
+++ b/src/Polly/RateLimit/LockFreeTokenBucketRateLimiter.cs
@@ -15,7 +15,7 @@ internal sealed class LockFreeTokenBucketRateLimiter : IRateLimiter
     private long addNextTokenAtTicks;
 
 #if !NETSTANDARD2_0
-    private SpinWait spinner = default;
+    private static SpinWait spinner => default;
 #endif
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


cbredlow:
@dotnet-policy-service agree

src/Polly/RateLimit/LockFreeTokenBucketRateLimiter.cs
```diff
@@ -15,7 +15,7 @@ internal sealed class LockFreeTokenBucketRateLimiter : IRateLimiter
     private long addNextTokenAtTicks;
 
 #if !NETSTANDARD2_0
-    private SpinWait spinner = default;
+    private static SpinWait spinner => default;
```
martincostello, line 18: In this case I don't believe this is the correct change to make here.

Making this a static property means that each access to the spinner effectively resets the internal state it has as all usage of it now gets a fresh value instead of the state being maintained for the lifetime of an instance of the rate limiter.

Either instead of `default` we should change the original code to use `new()`, and if that still raises warnings, then we should just explicitly suppress the CA1805 warning instead.

martincostello:
I'm going to close this due to activity. If you'd like to continue with this contribution, please open a new PR that addresses the comment here: https://github.com/App-vNext/Polly/pull/2032#discussion_r1538703370

Issue #2045, opened 2024-04-09
[Question]: How to use ExecuteAsync In VB.Net?


ricfwolff:
### What are you wanting to achieve?

Sorry to bring VB.Net... But some of us still need to work with it. :)

I'm trying to add Polly v8 to a vb.net project, but I'm facing a compiler error when I try to use the ExecuteAsync method:

<img width="754" alt="image" src="https://github.com/App-vNext/Polly/assets/9446182/ba43734e-47e1-4a1f-913a-989e3f66f657">

I noticed the ExecuteAsync method returns "ValueTask" instead of "Task". And it looks like that might be the problem...

### What code or approach do you have so far?

I also noticed that I can exchange ExecuteAsync for Execute and the compiler will stop complaining about it, but there's another post here on your GitHub explaining that we shouldn't do it, and anyway I tried and my RetryStrategy does not work. :(

Could someone please point me to what I'm doing wrong?

Thank you!

ricfwolff:
It does not work :(
Because it's expecting the _callback_ to be of ValueTask:

<img width="548" alt="image" src="https://github.com/App-vNext/Polly/assets/9446182/46afa94d-7041-4dd6-8d91-03262d930857">


But I cannot modify the callback because it'll throw compiler errors as well.
I'm not sure what's wrong here... 

Maybe a question for StackOverflow instead.

ricfwolff:
Good idea, I'll try that out and let you know. Thanks!

martincostello:
I just tried this out locally and it seems to work:

```vb
Imports System.Threading
Imports Polly

Module Program
    Sub Main()
        Demo().Wait()
    End Sub

    Async Function Demo() As Task
        ' The ResiliencePipelineBuilder creates a ResiliencePipeline
        ' that can be executed synchronously or asynchronously
        ' and for both void and result-returning user-callbacks.
        Dim pipeline = New ResiliencePipelineBuilder().AddTimeout(TimeSpan.FromSeconds(5)).Build()

        ' Synchronously
        pipeline.Execute(Sub()
                             ' Do something synchronously
                         End Sub)

        Await pipeline.ExecuteAsync(Function(token)
                                        Return New ValueTask(MyAsyncCallback(token))
                                    End Function, CancellationToken.None).AsTask()
    End Function

    Function ExecuteAsync(pipeline As ResiliencePipeline, callback As Func(Of CancellationToken, ValueTask), cancellationToken As CancellationToken) As Task
[...]

PR #2046, opened 2024-04-09
Add F# and VB.NET samples

martincostello:
Issue #2045 pointed out that the usage for Polly v8 isn't as terse as it is for C# due to the inability to directly await `ValueTask{<T>}`.

This PR adds some limited F# and VB.NET examples to the documentation that demonstrate how to bridge the gap between `ValueTask` and `Task` for Polly v8 pipelines.

### DIFF
--- a/samples/Intro.FSharp/Program.fs
+++ b/samples/Intro.FSharp/Program.fs
@@ -1,0 +1,49 @@
+﻿open System
+open System.Threading
+open System.Threading.Tasks
+open Polly
+
+let getBestFilmAsync(token: CancellationToken) =
+    task {
+        do! Task.Delay(1000, token) |> Async.AwaitTask
+        return "https://www.imdb.com/title/tt0080684/"
+    }
+
+let demo() =
+    async {
+        // The ResiliencePipelineBuilder creates a ResiliencePipeline
+        // that can be executed synchronously or asynchronously
+        // and for both void and result-returning user-callbacks.
+        let pipeline = ResiliencePipelineBuilder().AddTimeout(TimeSpan.FromSeconds(5)).Build()
+
+        // Synchronously
+        pipeline.Execute(Action(fun () -> Console.WriteLine("Hello, world!")))
+
+        // Asynchronously
+        // Note that the function is wrapped in a ValueTask for Polly to use as F# cannot
+        // await ValueTask directly, and AsTask() is used to convert the ValueTask returned by
+        // ExecuteAsync() to a Task so it can be awaited.
+        do! pipeline.ExecuteAsync(Func<CancellationToken, ValueTask>(fun token -> new ValueTask(task {
+            Console.WriteLine("Hello, world! Waiting for 1 second...")
+            do! Task.Delay(1000, token) |> Async.AwaitTask
+        })),
+        CancellationToken.None).AsTask() |> Async.AwaitTask
+
+        // Synchronously with result
+        let someResult = pipeline.Execute(Func<CancellationToken, string>(fun token -> "some-result"))
+
+        // Asynchronously with result
+        // Note that the function is wrapped in a ValueTask<string> for Polly to use as F# cannot
+        // await ValueTask directly, and AsTask() is used to convert the ValueTask<string> returned by
+        // ExecuteAsync() to a Task<string> so it can be awaited.
+        let! bestFilm = pipeline.ExecuteAsync(
+            Func<CancellationToken, ValueTask<string>>(fun token -> new ValueTask<string>(getBestFilmAsync(token))),
+            CancellationToken.None).AsTask() |> Async.AwaitTask
+
+        Console.WriteLine("Link to the best film: {0}", bestFilm)
+    }
+
+[<EntryPoint>]
+let main _ =
+    demo() |> Async.RunSynchronously
+    0
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -8,6 +8,7 @@
     <PackageVersion Include="coverlet.msbuild" Version="6.0.1" />
     <PackageVersion Include="BenchmarkDotNet" Version="0.13.12" />
     <PackageVersion Include="FluentAssertions" Version="6.12.0" />
+    <PackageVersion Include="FSharp.Core" Version="8.0.200" />
     <PackageVersion Include="GitHubActionsTestLogger" Version="2.3.3" />
     <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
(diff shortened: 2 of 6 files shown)


martincostello:
Thoughts @martintmk and @peter-csala?

peter-csala:
> Thoughts @martintmk and @peter-csala?

I'm not a VB or F# expert but I think we should either add direct support to these languages or say that they are not supported. We can't expect that everyone will read our samples and apply the `AsTask` workaround everywhere.

I'm not even sure that this solutio would work for more complex logic as well. The sample codes are too simple to make a fair judgement whether this workaround is applicable everywhere.

martincostello:
> not supported

I wouldn't say it's not supported, it's just difficult. We did however neglect to consider the impact on VB.NET and F#, which is our bad, but given we're not familiar with those languages in the first place, I don't think it's a failure on our part.

> The sample codes are too simple to make a fair judgement whether this workaround is applicable everywhere.

I asked @slang25 who knows F# well to review my initial sample, and it seems to be fairly trivial to make it work compared to VB (adding `|> ValueTask`).

I intentionally left them simple so that someone who considers themselves an "expert" can make their own judgement on whether they want to use it that way or not if they read the documentation. A VB guru might just say to themselves "this is too hard" and not use it, which is fine, if unfortunate. A "novice" might decide they're fine if they need to use it in one place and haven't ever used Polly before. The "this is too hard" outcome is what the original user who brought this to my attention went with.

As noted in the PR comment, we could consider adding some extension methods that do the `Task => ValueTask => Task` wrapping for the user at the expense of the additional allocations/overhead that entails.

I don't think there's otherwise anything we can do about it other than accept it's not a good experience in those languages and that maybe one day F# will catch up to C# in terms of native `ValueTask` await support. Going back to `Task` would undo the main point of why Polly v8 was created in the first place, as well as be yet another breaking change.

martincostello:
Yeah, I was thinking about putting them in a new package rather than in Core. Was more the API shape idea I was looking for feedback on initially.

peter-csala:
> Yeah, I was thinking about putting them in a new package rather than in Core. Was more the API shape idea I was looking for feedback on initially.

Does it impact only the `ExecuteAsync` and `ExecuteOutcomeAsync` overloads or anything where the API expects a `ValueTask` like `ShouldHandle`, `DelayGenerator` or `OnRetry`?

martincostello:
Any method returning `ValueTask` or `ValueTask<T>` - VB and F# can't await anything other than `Task` or `Task<T>`.

baronfel:
Out of the box F# cannot create ValueTasks with nice syntax, but there are a few ways to get to something usable. Here's a quick sample showing my two preferred ways:

```fsharp
module Polly.FSharp

open Polly
open System
open System.Threading
open System.Threading.Tasks

let pipeline =
    ResiliencePipelineBuilder()
        .AddRetry(Retry.RetryStrategyOptions())
        .AddTimeout(TimeSpan.FromSeconds 5.)
        .Build()

let token = CancellationToken.None

// this creates a Task using the F# built-in Task builder, then creates a valueTask from it
// this is more allocate-y but works out of the box
let builtinValue =
    task {
        let! result = pipeline.ExecuteAsync((fun token -> (task { return 42 }) |> ValueTask), token)
        return result
    }


// this uses the valueTask builder from IcedTasks to have 'direct' use of ValueTasks
// lower allocations, but requires adding a library (though one that is pretty widely used in the F# community)
// `dotnet add package IcedTasks`
open IcedTasks

let vTaskResult =
    valueTask {
        let! result = pipeline.ExecuteAsync((fun token -> valueTask { return 42 }), token)
        return result
    }
```

cc @theangrybyrd

martincostello:
Let's move the discussion of any possible new APIs to #2061.

I'll look to apply some further feedback to the F# sample based on @baronfel's suggestions so we can look to get this merged to provide some help to future C#/VB users trying to adopt Polly.

martincostello:
@baronfel @TheAngryByrd I've updated the F# sample to use IcedTasks as suggested - let me know if there's anything further that can be tightened up.

samples/Intro.FSharp/Program.fs
```diff
@@ -0,0 +1,66 @@
...
+open System
+open System.Threading
+open System.Threading.Tasks
+open IcedTasks
+open Polly
+
+let getBestFilmAsync token =
+    task {
+        do! Task.Delay(1000, token)
+        return "https://www.imdb.com/title/tt0080684/"
```
TheAngryByrd, line 11: Good choice :)

TheAngryByrd (approved):
Looks good from F# point of view :)

samples/Intro.FSharp/Program.fs
```diff
@@ -0,0 +1,66 @@
...
+                do! pipeline.ExecuteAsync(
+                    fun token ->
+                        valueTask {
+                            printfn "Hello, world! Waiting for 2 seconds..."
+                            do! Task.Delay(1000, token)
+                            printfn "Wait complete."
+                        }
+                    , token
+                )
+            }
```
TheAngryByrd, line 43: Oops after I comment I found something. This has an extra `valueTask` wrapper that doesn't need to be here. 
```suggestion

        do! pipeline.ExecuteAsync(
            fun token ->
                valueTask {
                    printfn "Hello, world! Waiting for 2 seconds..."
                    do! Task.Delay(1000, token)
                    printfn "Wait complete."
                }
            , token
        )            
```

samples/Intro.FSharp/Program.fs
```diff
@@ -0,0 +1,66 @@
...
+        // Note that Polly expects a ValueTask<T> to be returned, so the function uses the valueTask builder
+        // from IcedTasks to make it easier to use ValueTask<T>. See https://github.com/TheAngryByrd/IcedTasks.
+        let! bestFilm =
+            valueTask {
+                let! url = pipeline.ExecuteAsync((fun token -> valueTask {
+                    let! url = getBestFilmAsync(token)
+                    return url
+                }), token)
+                return url
+            }
```
TheAngryByrd, line 58:
```suggestion
        let! bestFilm = 
          pipeline.ExecuteAsync((fun token -> valueTask {
                    let! url = getBestFilmAsync(token)
                    return url
                }), token)
```

TheAngryByrd (changes requested):
Oops after I comment I found something. This has an extra valueTask wrapper in places that doesn't need to be here.

martincostello:
Thanks for the feedback everyone. The docs are now live here: https://www.pollydocs.org/use-with-fsharp-and-visual-basic.html

PR #2056, opened 2024-04-12
Port fixes from .NET 9 testing

martincostello:
Port various changes/fixes from #2003 that don't depend on .NET 9.

- Fix IDE0057 warnings.
- Suppress xUnit1031 warnings in tests that explicitly test synchronous behaviours.
- Remove properties that are implied by `IsAotCompatible=true`.
- Only mark as AoT-compatible for `net8.0` where it's actually supported.
- Fix test not being declared in a namespace.
- Bump the samples' reference to ourself to the latest version.
- Bump `Microsoft.Extensions.TimeProvider.Testing` to 8.4.0.
- Use an MSBuild property to allow for easy conditions for multi-targeting.

### DIFF
--- a/src/Polly.Core/Polly.Core.csproj
+++ b/src/Polly.Core/Polly.Core.csproj
@@ -11,12 +11,8 @@
     <LegacySupport>true</LegacySupport>
     <InjectSharedSources>true</InjectSharedSources>
   </PropertyGroup>
-  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))">
-    <EnableAotAnalyzer>true</EnableAotAnalyzer>
-    <EnableSingleFileAnalyzer>true</EnableSingleFileAnalyzer>
-    <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
+  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
     <IsAotCompatible>true</IsAotCompatible>
-    <IsTrimmable>true</IsTrimmable>
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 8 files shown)


src/Polly.Core/Polly.Core.csproj
```diff
@@ -11,12 +11,8 @@
     <LegacySupport>true</LegacySupport>
     <InjectSharedSources>true</InjectSharedSources>
   </PropertyGroup>
-  <PropertyGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net6.0'))">
```
martintmk, line 14: So no AOT on .NET 6

src/Polly/Utilities/KeyHelper.cs
```diff
@@ -6,5 +6,9 @@ internal static class KeyHelper
     private const int GuidPartLength = 8;
 
     public static string GuidPart() =>
+#if NET6_0_OR_GREATER
+        Guid.NewGuid().ToString()[..GuidPartLength];
```
martintmk, line 10: Isn't this just adding extra complexity?

src/Polly.Core/Polly.Core.csproj
martincostello: It just isn't being marked as it is explicitly, but in the #2003 PR with the .NET 9 SDK if you enable it it errors out saying it's not supported.

Either I can turn it back on, but we'll lose it again in November, or we turn it off now.

src/Polly/Utilities/KeyHelper.cs
martincostello: Is it more code? Yes. Is it complicated? No, it's in two small places.

It fails the build with the .NET 9 SDK as it wasn't smart enough to detect it in .NET 8, so we need to do _something_.

Alternatively, I can just suppress the warning and put it back how it was.

src/Polly.Core/Polly.Core.csproj
martintmk: Ahh ok then :)

src/Polly/Utilities/KeyHelper.cs
martintmk: no problem then, thanks for explanation

martintmk (approved):

PR #2063, opened 2024-04-18
Fix ToC

martincostello:
Fix weird navbar rendering caused by #2046.


martincostello:
Checked by running the output locally.

PR #2060, opened 2024-04-15
[Docs] Unify strategy descriptions and add Telemetry sections

peter-csala:
# Pull Request

## The issue or feature being addressed

I have found the strategies documentation paged a bit inconsistent:
- The `About` section sometimes contains a short description while other times doesn't.
- I have found some of the description fields of the `Defaults` sections rather vague.

## Details on the issue fix or feature implementation

- Unified Timeout + added Telemetry
- Unified Retry + added Telemetry
- Unified Fallback + added Telemetry
- Unified Rate Limiter + added Telemetry
- Unified Hedging + added Telemetry
- Unified Circuit Breaker + added Telemetry

## Confirm the following

- [x]  I started this PR by branching from the head of the default branch
- [x]  I have targeted the PR to merge into the default branch
- [ ]  I have included unit tests for the issue/feature
- [x]  I have successfully run a local build

PR #2069, opened 2024-04-22
Add unit to execution time in telemetry events

peter-csala:
# Pull Request

Follow up of #2060 

- Execution time did not have an unit in the reported telemetry events

## Confirm the following

### DIFF
--- a/src/Polly.Extensions/Telemetry/Log.cs
+++ b/src/Polly.Extensions/Telemetry/Log.cs
@@ -66,7 +66,7 @@ internal static partial class Log
                 "Result: '{Result}', " +
                 "Handled: '{Handled}', " +
                 "Attempt: '{Attempt}', " +
-                "Execution Time: '{ExecutionTimeMs}'",
+                "Execution Time: {ExecutionTime}ms",
         EventName = "ExecutionAttempt",
         SkipEnabledCheck = true)]
     public static partial void ExecutionAttempt(
@@ -79,6 +79,6 @@ internal static partial class Log
         object? result,
         bool handled,
         int attempt,
-        double executionTimeMs,
+        double executionTime,
         Exception? exception);
 }
--- a/src/Polly.Extensions/Telemetry/Log.cs
+++ b/src/Polly.Extensions/Telemetry/Log.cs
@@ -13,7 +13,7 @@ internal static partial class Log
     internal const string SourceWithoutStrategy = "Source: '{PipelineName}/{PipelineInstance}'";
     internal const string OperationKey = "Operation Key: '{OperationKey}'";
     internal const string Result = "Result: '{Result}'";
-    internal const string ExecutionTime = "Execution Time: {ExecutionTime}ms";
+    internal const string ExecutionTime = "Execution Time: {ExecutionTimeMs}ms";
 
     [LoggerMessage(
         EventId = 0,
@@ -62,7 +62,7 @@ internal static partial class Log
         string pipelineInstance,
         string? operationKey,
         object? result,
-        double executionTime,
+        double executionTimeMs,
         Exception? exception);
 
     [LoggerMessage(
@@ -86,6 +86,6 @@ internal static partial class Log
         object? result,
         bool handled,
         int attempt,
-        double executionTime,
+        double executionTimeMs,
         Exception? exception);
 }
--- a/src/Polly.Extensions/Telemetry/Log.cs
+++ b/src/Polly.Extensions/Telemetry/Log.cs
@@ -8,13 +8,20 @@ namespace Polly.Telemetry;
 [ExcludeFromCodeCoverage]
 internal static partial class Log
 {
+    internal const string Separator = ", ";
+    internal const string SourceWithStrategy = "Source: '{PipelineName}/{PipelineInstance}/{StrategyName}'";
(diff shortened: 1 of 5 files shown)


peter-csala:
I've just realized that the Chaos strategies are also reporting telemetry events.
- Do we want to extend those pages as well with `Telemetry` section?
- Do we want to list those events on the `Telemetry` page's [event names](https://www.pollydocs.org/advanced/telemetry.html#event-names) section?

src/Polly.Extensions/Telemetry/Log.cs
```diff
@@ -61,12 +68,12 @@ internal static partial class Log
...
         Message = "Execution attempt. " +
-                "Source: '{PipelineName}/{PipelineInstance}/{StrategyName}', " +
-                "Operation Key: '{OperationKey}', " +
-                "Result: '{Result}', " +
+                SourceWithStrategy + Separator +
+                OperationKey + Separator +
+                Result + Separator +
                 "Handled: '{Handled}', " +
                 "Attempt: '{Attempt}', " +
-                "Execution Time: '{ExecutionTimeMs}'",
```
martincostello, line 69: I think this is a breaking change so we shouldn't actually make it.

If users are using the log event parameter names to emit telemetry, this would cause the metrics's name to change and might break things, such as alerts configured based on the metrics' names and values.

Am I correct on my understanding here @martintmk, as I seem to remember this was what you were planning to do in Teams applications with the v8 telemetry. Or have I misunderstood?

martincostello:
> * Do we want to extend those pages as well with `Telemetry` section?

I would say yes to both so there's parity in the documentation.

src/Polly.Extensions/Telemetry/Log.cs
peter-csala: This `Log` class is marked as `internal`. I don't understand how can this cause a breaking change? Can you please rephrase?

martincostello: The names of the parameters (`ExecutionTimeMs` => `ExecutionTime`) are emitted as part of the structured logging events, which (IIRC) are then consumed to emit metrics.

martintmk: I think changing `ExecutionTimeMs` to `ExecutionTime` is doable. People shouldn't really on log parameter names for anything important. (I hope at least :D)

peter-csala: I can revert the renaming of the parameter, but with that the log message pattern will look a bit clumsy and we will continue to have inconsistent parameter naming across two types of log events.

martincostello: It's possible I'm getting confused with this sort of guidance: https://github.com/dotnet/aspnetcore/pull/18494#discussion_r379603655

@martintmk How are you generating your metrics from Polly?

peter-csala:
> > * Do we want to extend those pages as well with `Telemetry` section?
> 
> I would say yes to both so there's parity in the documentation.

Okay, I'll do that in a separate PR.

src/Polly.Extensions/Telemetry/Log.cs
martintmk: For us, the [metrics](https://www.pollydocs.org/advanced/telemetry.html#metrics) are important and effectively fixed, so no breaking changes here. 

For log properties we can be little bit more relaxed. Currently, I don't think anyone relies on this exact name so I am ok with the rename. (old name still works for me though)

martintmk (approved):

src/Polly.Extensions/Telemetry/Log.cs
martincostello: OK looks like this isn't a breaking change to any metrics, however I think on reflection it's better to keep the unit in the logging property name. Otherwise if looking at the property in a tool (Aspire, Datadog, Kibana...) you're not immediately going to be able to tell the unit as it's just a number. If it were a `TimeSpan` it would be self-explanatory. We can still keep the `ms` in the template for when it's rendered/formatted.

Here's how Polly logging looks in Kibana:

image

With this change, it would change to `ExecutionTime=35.595` and wouldn't be clear if it's milliseconds or seconds (or something else).

peter-csala: If we were shooting for self-explanatory field names then `handled`, `attempt`, or `result` are also can be considered as vague field names.

Anyway, I used `ExecutionTimeMs` for `ExecutionAttempt` and `PipelineExecuted` as well.

martincostello (approved):

src/Polly.Extensions/Telemetry/Log.cs
martincostello: It's less self-explanatory, and more ambiguity. "Attempts" don't have a natural unit, where as things like time, distance, temperature, etc. do, and getting them wrong [can be very expensive](https://en.wikipedia.org/wiki/Mars_Climate_Orbiter#:~:text=An%20investigation%20attributed%20the%20failure%20to%20a%20measurement%20mismatch%20between%20two%20measurement%20systems%3A%20SI%20units%20(metric)%20by%20NASA%20and%20US%20customary%20units%20by%20spacecraft%20builder%20Lockheed%20Martin.%5B3%5D), so it's better to be explicit where the value doesn't inherently convey the units.

PR #2071, opened 2024-04-23
[Docs] Add telemetry section to chaos strategies documentation pages

peter-csala:
# Pull Request

Follow up of the following PRs:
- #2060 
- #2069 

- Unified Behavior page
- Unified Fault page
- Unified Latency page
- Unified Outcome page
- Updated chaos index page
- Updated telemetry page as well

## Confirm the following


peter-csala:
@vany0114 

I was planning with the Behavior Chaos and I've just realized if the `BehaviorGenerator` throws an exception then neither the telemetry will be reported nor the `OnBehaviorInjected` will be called.

```cs
var builder = new ResiliencePipelineBuilder<int>() 
{ 
	Name = "MyPipeline",
	InstanceName = "MyPipelineInstance"
}; 
var pipeline = builder
.AddChaosBehavior(new ChaosBehaviorStrategyOptions()
{
	InjectionRate = 1,
	Enabled = true,
	BehaviorGenerator = static args =>
    {
       	Console.WriteLine("Custom Behavior Injected");
		//throw new CustomException(); // uncomment to reproduce
		return default;
    },
	OnBehaviorInjected = static args =>
   	{
		Console.WriteLine("OnBehaviorInjected, Operation: {0}.", args.Context.OperationKey);
		return default;
	}
})
.ConfigureTelemetry(new TelemetryOptions
{
	LoggerFactory = LoggerFactory.Create(builder => builder.AddConsole())
}) 
.Build();

ResilienceContext resilienceContext = ResilienceContextPool.Shared.Get("MyBehaviorInjectedOperation");

try
{
	await pipeline.ExecuteAsync<int>(ctx => { return ValueTask.FromResult<int>(1); }, resilienceContext);
}
catch(Exception ex)
{
	Console.WriteLine(ex.Message);
}

ResilienceContextPool.Shared.Return(resilienceContext);
```
Dotnet fiddle link: https://dotnetfiddle.net/EaycSE

The standard strategies do report telemetry even in a case of exception. They also call the `OnXYZ` callback as well.

Does this strategy behave differently intentionally? 

cc: @martincostello , @martintmk

docs/chaos/behavior.md
```diff
@@ -68,10 +68,10 @@ var pipeline = new ResiliencePipelineBuilder()
 
 ## Defaults
 
-| Property             | Default Value | Description                                    |
-|----------------------|---------------|------------------------------------------------|
-| `OnBehaviorInjected` | `null`        | Action executed when the behavior is injected. |
-| `BehaviorGenerator`  | `null`        | Custom behavior to be injected.                |
+| Property             | Default Value | Description                                                                                                                  |
+|----------------------|---------------|------------------------------------------------------------------------------------------------------------------------------|
+| `BehaviorGenerator`  | `null`        | This **mandatory** delegate allows you to inject custom behavior by utilizing information that is only available at runtime. |
```
martincostello, line 73: Personally I don't think the emphasis here is needed, and it looks "shouty". "Required" would fit better with language we've used elsewhere too?

docs/chaos/latency.md
```diff
@@ -90,15 +92,45 @@ var pipeline = new ResiliencePipelineBuilder()
...
 ## Defaults
 
-| Property            | Default Value | Description                                            |
-|---------------------|---------------|--------------------------------------------------------|
-| `Latency`           | `30 seconds`  | A `TimeSpan` indicating the delay to be injected.      |
-| `LatencyGenerator`  | `null`        | Generates the latency to inject for a given execution. |
-| `OnLatencyInjected` | `null`        | Action executed when latency is injected.              |
+| Property            | Default Value | Description                                                                                                          |
+|---------------------|---------------|----------------------------------------------------------------------------------------------------------------------|
+| `Latency`           | 30 seconds    | Defines a **static** delay to be injected.                                                                           |
```
martincostello, line 97: Fixed or constant to avoid confusion with `static`?

docs/chaos/behavior.md
peter-csala: Done

docs/chaos/latency.md
peter-csala: The same wording is used in case of [Circuit Breaker](https://www.pollydocs.org/strategies/circuit-breaker.html#defaults) and [Timeout](https://www.pollydocs.org/strategies/timeout.html#defaults) as well.

Shall I rename there as well?

peter-csala:
Based on my experience the `Result` part of the telemetry events are never populated for the chaos strategies. IMHO we could populate it in case of `Fault` and `Outcome`.

What do you guys think?
cc: @martincostello , @martintmk

docs/chaos/latency.md
peter-csala: Done

docs/advanced/telemetry.md
```diff
@@ -140,14 +140,20 @@ Tags:
...
+| `OnHedging`             | [Link](../strategies/hedging.md#telemetry)         |
+| `OnTimeout`             | [Link](../strategies/timeout.md#telemetry)         |
+| `OnCircuitClosed`       | [Link](../strategies/circuit-breaker.md#telemetry) |
+| `OnCircuitOpened`       | [Link](../strategies/circuit-breaker.md#telemetry) |
+| `OnCircuitHalfOpened`   | [Link](../strategies/circuit-breaker.md#telemetry) |
+| `OnRateLimiterRejected` | [Link](../strategies/rate-limiter.md#telemetry)    |
+| `Chaos.OnFault`         | [Link](../chaos/fault.md#telemetry)                |
+| `Chaos.OnOutcome`       | [Link](../chaos/outcome.md#telemetry)              |
+| `Chaos.OnLatency`       | [Link](../chaos/latency.md#telemetry)              |
+| `Chaos.OnBehavior`      | [Link](../chaos/behavior.md#telemetry)             |
```
martincostello, line 156: Do we need a table here? Could it just be a list where each item is itself a link?

vany0114:
Sorry for the delay here guys, will take a look as soon as I can.

vany0114:
> I hope I do something really silly but for me the Outcome Chaos strategy does not seem to inject exception.
[...]

@peter-csala you're right! And I it is because of the way we're handling the outcome when it is an exception since originally the outome strategy wasn't designed to return an exception as an outcome, so instead it is returning the default value ot `T`. 

I added the [OnOutcomeInjected delegate](https://dotnetfiddle.net/Ir3nOf) to the fiddle you shared and look at the results:

```
info: Polly[0]
      Resilience event occurred. EventName: 'Chaos.OnOutcome', Source: '(null)/(null)/Chaos.Outcome', Operation Key: '', Result: ''
OnOutcomeInjected, Outcome: 0, Operation: .
OnOutcomeInjected, Exception: Exception of type 'System.Exception' was thrown., Operation: .
Dumping object(Int32)
 0
```

We would need to do a change in the `Outcome` strategy to fully support it, I can help with that PR, something like this:
image

One caveat there that undersocres the this usage as an anti-patter is that the telemetry event was registered as an empty result, instead of the exception, we could also do something like this before returning the exception to mitigate it:

```C#
_telemetry.Report(new(ResilienceEventSeverity.Information, ChaosOutcomeConstants.OnOutcomeInjectedEvent), context, outcome.Value.Exception);
```

vany0114:
> I was playing with the Behavior Chaos and I've just realized if the `BehaviorGenerator` throws an exception then neither the telemetry will be reported nor the `OnBehaviorInjected` will be called.
> The standard strategies do report telemetry even in a case of exception. They also call the `OnXYZ` callback as well.
> 
> Does this strategy behave differently intentional?

@peter-csala Yep that's intended, if the generator throws an exception we're not handling that, we just let the exception to bubble up, I spot checked a couple of Polly strategies and I could see that they are handling errors in the generators and tracking those as telemetry events 🤔 can you please provide an example when you say that the standard strategies report telemetry even in a case of exception?

peter-csala:
I will create a dedicated discussion thread for each question just to keep the PR focused.

I'll update this comment with the links.

peter-csala:
> @peter-csala Yep that's intended, if the generator throws an exception we're not handling that, we just let the exception to bubble up, I spot-checked a couple of Polly strategies and I couldn't see that they are handling errors in the generators and tracking those as telemetry events 🤔 can you please provide an example when you say that the standard strategies report telemetry even in a case of exception?

Here is an example in case of Fallback, both the telemetry is reported as well as the `OnFallback` is called

```cs
var pipeline = new ResiliencePipelineBuilder<int>()
.AddFallback(new FallbackStrategyOptions<int>
{
	ShouldHandle = static args => PredicateResult.True(),
	FallbackAction = static args => throw new Exception("Fallback boom"),
	OnFallback = static args => 
	{
		Console.WriteLine("OnFallback is called");
		return default;
	}
})
.ConfigureTelemetry(new TelemetryOptions
{
	LoggerFactory = LoggerFactory.Create(builder => builder.AddConsole())
}) 
.Build();

var resilienceContext = ResilienceContextPool.Shared.Get();
try
{
	var result = await pipeline.ExecuteAsync<int>(ctx => { return ValueTask.FromResult<int>(1); }, resilienceContext);
	result.Dump();
}
catch(Exception ex)
{
	Console.WriteLine(ex.Message);
}
ResilienceContextPool.Shared.Return(resilienceContext);
```
```none

[1mwarn: Polly[0]
      Resilience event occurred. EventName: 'OnFallback', Source: '(null)/(null)/Fallback', Operation Key: '', Result: '1'
OnFallback is called
Fallback boom
```
https://dotnetfiddle.net/S7AjzT

peter-csala:
@vany0114 Are you fine with this documentation PR?

docs/chaos/latency.md
```diff
@@ -90,15 +92,45 @@ var pipeline = new ResiliencePipelineBuilder()
...
+| `LatencyGenerator`  | `null`        | This delegate allows you to **dynamically** inject delay by utilizing information that is only available at runtime. |
+| `OnLatencyInjected` | `null`        | If provided then it will be invoked after the latency injection occurred.                                            |
 
 > [!NOTE]
 > If both `Latency` and `LatencyGenerator` are specified then `Latency` will be ignored.
 
+---
+
+> [!IMPORTANT]
+> Please note that if the calculated latency is negative (regardless if it's fixed or dynamic) then it will not be injected.
```
vany0114, line 107: 👍

vany0114 (approved):
🙏

Issue #1859, opened 2023-12-14
[Question]: Customize telemetry event severity?


wjrogers:
### What are you wanting to achieve?

I am using Polly v8 RateLimiter strategies to pro-actively restrict my client's request rate. I noticed that the OnRateLimiterRejected event is logged with Error severity. I want rate limiting events to be logged with at most Information severity because I am imposing the limit on myself. Is there a way to customize event severity?

### What code or approach do you have so far?

I poked around the source code and tried a few things like handling `RateLimiterStrategyOptions.OnRejected` or calling `builder.ConfigureTelemetry(new TelemetryOptions())`, but I could not find a way to modify the event severity. All the properties exposed through these APIs are read-only.

martincostello:
As far as I'm aware, as you've indicated, there's no way to change the severities at this time. It seems a reasonable suggestion to be able to allow this to be customised though.

martintmk:
Thinking we could introduce something like this:

``` csharp
public class TelemetryOptions
{
    // ... existing properties

    public IDictionary<string, ResilienceEventSeverity> EventSeverityOverrides { get; set; }
}
```

BY default, it would be empty. If specified one could override the built-in severity for events. Also, I think we can downgrade the `OnRejected` to `Warning`. Error seems too high indeed.

Usage:

``` csharp
services.Configure<TelemetryOptions>(options =>
{
    options.EventSeverityOverrides["OnRateLimiterRejected"] = ResilienceEventSeverity.Information;
}); 
```

peter-csala:
☝🏻 seems a bit error-prone. Non-existing key overrides would result in `KeyNotFoundException`. 

Introducing hierarchical constants might be a better option IMHO.
```
options.EventSeverityOverrides[TelemetryNames.RateLimiter.OnRejected] = ResilienceEventSeverity.Information;
```

martincostello:
Constants or not, we wouldn't get exceptions because we'd use `TryGetValue()` not the indexer.

martintmk:
Since this is for advanced scenarios, the user can put whatever strings they want into the dictionary. If the name does not match up with existing telemetry events, we won't do any overrides.

Alternatively, we can allow delegate to be used (faster API and more flexible):

``` csharp
public class TelemetryOptions
{
    // ... existing properties
    public Func<ResilienceEvent, ResilienceEventSeverity>? ResilienceEventSeverityProvider { get; set; }
}
```

wjrogers:
[...]

Issue #1958, opened 2024-02-07
[Question]: How can I suppress Execution Attempt log when using Polly V8 ResiliencePipelineBuilder?


raul-portela:
### What are you wanting to achieve?

I'd like to suppress a log that's creating unnecessary clutter. Can it be done?

### What code or approach do you have so far?

public class ScreenshotStreamWithRetries
{
    private readonly ILogger<ScreenshotStreamWithRetries> _logger;

    public ScreenshotStreamWithRetries(ILogger<ScreenshotStreamWithRetries> logger)
    {
        _logger = logger;
    }

    public static string Name => "screenshot-stream-with-retries";

    public ResiliencePipelineBuilder<Stream> Configure(ResiliencePipelineBuilder<Stream> pipelineBuilder)
    {
        return pipelineBuilder.AddRetry(new RetryStrategyOptions<Stream>
        {
            MaxRetryAttempts = 3,
            Delay = TimeSpan.FromSeconds(2),
            BackoffType = DelayBackoffType.Exponential,
            ShouldHandle = new PredicateBuilder<Stream>()
                .Handle<Exception>(),
            OnRetry = retryArguments =>
            {               
                _logger.LogWarning($"Screenshot failed with exception {retryArguments.Outcome.Exception.Message}. " +
                             $"Waiting {retryArguments.Duration} before next retry. Retry attempt {retryArguments.AttemptNumber}");
                return ValueTask.CompletedTask;
            }
        });
    }
}

### Additional context

In a .NET 8 project I'm using Polly V8 and with ResiliencePipelineBuilder, and this log shows always, even if there is no retry:

Execution attempt. Source: 'screenshot-stream-with-retries/(null)/Retry', Operation Key: 'null', Result: 'System.IO.MemoryStream', Handled: 'false', Attempt: '0', Execution Time: '730.3186'

martincostello:
You could:

1. Change the Polly log level to `Warning` or above.
2. Use a dynamic filter to remove the message based on your criteria.

Otherwise this sounds like a feature request similar to #1859.

serhatataman:
> That solved the issue. Thanks!!!

Hi, could you please share how you solved the issue? with a code example if possible. Thank you :)

PR #2072, opened 2024-04-24
Allow changing the severity of resilience events

martintmk:
# Pull Request

#1859 

This can be used to customize the severity of resilience events as:

``` csharp
services.AddResiliencePipeline("my-strategy", (builder, context) =>
{
    // Create a new instance of telemetry options by using copy-constructor of the global ones.
    // This ensures that common configuration is preserved.
    var telemetryOptions = new TelemetryOptions(context.GetOptions<TelemetryOptions>());

    telemetryOptions.SeverityProvider = ev =>
    {
        if (ev.EventName == "OnRetry")
        {
            // Decrease the severity of particular event.
            return ResilienceEventSeverity.Debug;
        }

        return ev.Severity;
    };

    builder.AddTimeout(TimeSpan.FromSeconds(1));

    // Override the telemetry configuration for this pipeline.
    builder.ConfigureTelemetry(telemetryOptions);
});
```

## Confirm the following

### DIFF
--- a/docs/advanced/telemetry.md
+++ b/docs/advanced/telemetry.md
@@ -264,15 +264,15 @@ services.AddResiliencePipeline("my-strategy", (builder, context) =>
     // This ensures that common configuration is preserved.
     var telemetryOptions = new TelemetryOptions(context.GetOptions<TelemetryOptions>());
 
-    telemetryOptions.SeverityProvider = @event =>
+    telemetryOptions.SeverityProvider = args =>
     {
-        if (@event.Event.EventName == "OnRetry")
+        if (args.Event.EventName == "OnRetry")
         {
             // Decrease the severity of particular event.
             return ResilienceEventSeverity.Debug;
         }
 
-        return @event.Event.Severity;
+        return args.Event.Severity;
     };
 
     builder.AddTimeout(TimeSpan.FromSeconds(1));
--- a/docs/advanced/telemetry.md
+++ b/docs/advanced/telemetry.md
@@ -264,15 +264,15 @@ services.AddResiliencePipeline("my-strategy", (builder, context) =>
     // This ensures that common configuration is preserved.
     var telemetryOptions = new TelemetryOptions(context.GetOptions<TelemetryOptions>());
 
-    telemetryOptions.SeverityProvider = ev =>
+    telemetryOptions.SeverityProvider = @event =>
     {
-        if (ev.Event.EventName == "OnRetry")
+        if (@event.Event.EventName == "OnRetry")
         {
             // Decrease the severity of particular event.
             return ResilienceEventSeverity.Debug;
         }
 
-        return ev.Event.Severity;
+        return @event.Event.Severity;
     };
 
     builder.AddTimeout(TimeSpan.FromSeconds(1));
--- a/docs/advanced/telemetry.md
+++ b/docs/advanced/telemetry.md
@@ -251,3 +251,34 @@ Each resilience strategy can generate telemetry data through the [`ResilienceStr
 To leverage this telemetry data, users should assign a `TelemetryListener` instance to `ResiliencePipelineBuilder.TelemetryListener` and then consume the `TelemetryEventArguments`.
 
 For common scenarios, it is expected that users would make use of `Polly.Extensions`. This extension enables telemetry configuration through the `ResiliencePipelineBuilder.ConfigureTelemetry(...)` method, which processes `TelemetryEventArguments` to generate logs and metrics.
+
+## Customizing the severity of telemetry events
+
+To customize the severity of telemetry events, set the [`SeverityProvider`](xref:Polly.Telemetry.TelemetryOptions.SeverityProvider) delegate that allows changing the default severity of resilience events:
+
+<!-- snippet: telemetry-severity-override -->
+```cs
+services.AddResiliencePipeline("my-strategy", (builder, context) =>
+{
+    // Create a new instance of telemetry options by using copy-constructor of the global ones.
+    // This ensures that common configuration is preserved.
+    var telemetryOptions = new TelemetryOptions(context.GetOptions<TelemetryOptions>());
+
+    telemetryOptions.SeverityProvider = ev =>
+    {
+        if (ev.Event.EventName == "OnRetry")
+        {
+            // Decrease the severity of particular event.
+            return ResilienceEventSeverity.Debug;
+        }
+
+        return ev.Event.Severity;
+    };
+
+    builder.AddTimeout(TimeSpan.FromSeconds(1));
+
+    // Override the telemetry configuration for this pipeline.
+    builder.ConfigureTelemetry(telemetryOptions);
+});
+```
+<!-- endSnippet -->
--- a/docs/advanced/telemetry.md
+++ b/docs/advanced/telemetry.md
@@ -264,18 +264,15 @@ services.AddResiliencePipeline("my-strategy", (builder, context) =>
     // This ensures that common configuration is preserved.
     var telemetryOptions = new TelemetryOptions(context.GetOptions<TelemetryOptions>());
 
-    telemetryOptions.SeverityProvider = args =>
+    telemetryOptions.SeverityProvider = args => args.Event.EventName switch
     {
-        if (args.Event.EventName == "OnRetry")
-        {
-            // Decrease the severity of particular event.
-            return ResilienceEventSeverity.Debug;
-        }
-
-        return args.Event.Severity;
+        // Decrease severity of specific events
+        "OnRetry" => ResilienceEventSeverity.Debug,
+        "ExecutionAttempt" => ResilienceEventSeverity.Debug,
+        _ => args.Event.Severity
     };
 
-    builder.AddTimeout(TimeSpan.FromSeconds(1));
+    builder.AddRetry(new RetryStrategyOptions());
 
     // Override the telemetry configuration for this pipeline.
     builder.ConfigureTelemetry(telemetryOptions);
--- a/src/Polly.Extensions/Telemetry/TelemetryOptions.cs
+++ b/src/Polly.Extensions/Telemetry/TelemetryOptions.cs
@@ -78,5 +78,5 @@ public class TelemetryOptions
     /// <value>
     /// The default value is <see langword="null"/>.
     /// </value>
-    public Func<ResilienceEvent, ResilienceEventSeverity>? SeverityProvider { get; set; }
+    public Func<SeverityProviderArguments, ResilienceEventSeverity>? SeverityProvider { get; set; }
 }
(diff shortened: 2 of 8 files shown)


src/Polly.Extensions/Telemetry/TelemetryOptions.cs
```diff
@@ -48,4 +71,12 @@ public class TelemetryOptions
...
         _ => result,
     };
+
+    /// <summary>
+    /// Gets or sets the resilience event severity provider that allows customizing the severity of resilience events.
+    /// </summary>
+    /// <value>
+    /// The default value is <see langword="null"/>.
+    /// </value>
+    public Func<ResilienceEvent, ResilienceEventSeverity>? SeverityProvider { get; set; }
```
martintmk, line 81:
```suggestion
    public Func<SeverityProviderArguments, ResilienceEventSeverity>? SeverityProvider { get; set; }
```

Shall we wrap it to allow adding more stuff to it in the future if needed?

cc @martincostello

martincostello: I'm erring to us not needing that as I wonder what we'd do that wasn't inside the event that was in `TelemetryEventArguments<TResult, TArgs>` but that didn't need the type arguments, but at the same time is safer than not if we did just in case.

src/Snippets/Docs/Telemetry.cs
```diff
@@ -134,4 +134,36 @@ internal sealed class CustomMeteringEnricher : MeteringEnricher
...
+
+            telemetryOptions.SeverityProvider = ev =>
+            {
+                if (ev.EventName == "OnRetry")
+                {
+                    // Decrease the severity of particular event.
+                    return ResilienceEventSeverity.Debug;
+                }
+
+                return ev.Severity;
```
martincostello, line 158:
```suggestion
            telemetryOptions.SeverityProvider = @event =>
            {
                if (@event.EventName == "OnRetry")
                {
                    // Decrease the severity of particular event.
                    return ResilienceEventSeverity.Debug;
                }

                return @event.Severity;
```

martincostello (commented):
Seems fine to me - just needs docs and tests!

src/Polly.Extensions/Telemetry/TelemetryOptions.cs
martintmk: Maybe we can pass `ResilienceContext` for extra flexibility. I wonder whether we are not overdoing it though.

src/Snippets/Docs/Telemetry.cs
```diff
@@ -134,4 +134,36 @@ internal sealed class CustomMeteringEnricher : MeteringEnricher
...
+
+        services.AddResiliencePipeline("my-strategy", (builder, context) =>
+        {
+            // Create a new instance of telemetry options by using copy-constructor of the global ones.
+            // This ensures that common configuration is preserved.
+            var telemetryOptions = new TelemetryOptions(context.GetOptions<TelemetryOptions>());
+
+            telemetryOptions.SeverityProvider = ev =>
+            {
+                if (ev.EventName == "OnRetry")
```
peter-csala, line 152: I think if we would expose all the telemetry event name constants via a simple static class then the branching logic would safer. It would also help users of the API to use switch expression.

src/Snippets/Docs/Telemetry.cs
```diff
@@ -134,4 +134,36 @@ internal sealed class CustomMeteringEnricher : MeteringEnricher
...
+
+        #region telemetry-severity-override
+
+        services.AddResiliencePipeline("my-strategy", (builder, context) =>
+        {
+            // Create a new instance of telemetry options by using copy-constructor of the global ones.
+            // This ensures that common configuration is preserved.
+            var telemetryOptions = new TelemetryOptions(context.GetOptions<TelemetryOptions>());
+
+            telemetryOptions.SeverityProvider = ev =>
```
peter-csala, line 150: Would this approach allow the users to override the severity of the `ExecutionAttempt` event as well?

wjrogers:
This would meet my needs. Thanks for the heads up!

src/Snippets/Docs/Telemetry.cs
martintmk: Yup, but the information about the strategy type the produces it is missing.

src/Snippets/Docs/Telemetry.cs
martintmk: It would require to expose a lot of public members for a limited audience. Let's wait and see. My vote would be to not expose anything unless absolutely necessary. Consumers of this API can just hardcode the strings, these won't change anyway as they are considered public contract as part of telemetry.

martincostello: In this specific case you could also do `nameof(RetryStrategyOptions<>.OnRetry)` (I think).

peter-csala: The `nameof` won't work for all cases. For example 
- in case of rate limiter `OnRateLimiterRejected` event name vs `OnRejected` options property
- in case of circuit breaker `OnCircuitHalfOpened` event name vs `OnHalfOpened` options property
- in case of latency `Chaos.OnLatency` event name vs `OnLatencyInjected` options property
- etc.

peter-csala: > It would require to expose a lot of public members for a limited audience. Let's wait and see. My vote would be to not expose anything unless absolutely necessary. Consumers of this API can just hardcode the strings, these won't change anyway as they are considered public contract as part of telemetry.

If you are saying that it is already implicitly part of the public contract then I don't see why it is a problem to make it explicit?

src/Snippets/Docs/Telemetry.cs
peter-csala: I was asking this because of this issue: https://github.com/App-vNext/Polly/issues/1958

src/Snippets/Docs/Telemetry.cs
```diff
@@ -147,15 +147,15 @@ public static void SeverityOverrides()
             // This ensures that common configuration is preserved.
             var telemetryOptions = new TelemetryOptions(context.GetOptions<TelemetryOptions>());
 
-            telemetryOptions.SeverityProvider = ev =>
+            telemetryOptions.SeverityProvider = @event =>
```
martincostello, line 150: I guess now this would be better named `args`?

test/Polly.Extensions.Tests/Telemetry/TelemetryListenerImplTests.cs
```diff
@@ -453,9 +453,33 @@ public void PipelineExecuted_ShouldBeSkipped()
         events.Should().HaveCount(0);
     }
 
+    [Fact]
+    public void SeverityProvider_EnsureRespected()
```
martincostello, line 457: Is it worth adding a test that explicitly validates the severity isn't changed if `SeverityProvider` is null?

test/Polly.Extensions.Tests/Telemetry/TelemetryOptionsTests.cs
```diff
@@ -21,4 +23,29 @@ public void Ctor_EnsureDefaults()
         using var response = new HttpResponseMessage(HttpStatusCode.OK);
         options.ResultFormatter(resilienceContext, response).Should().Be(200);
     }
+
+    [Fact]
+    public void CopyCtor_Ok()
```
martincostello, line 28: Is it worth us adding an assert that validates how many properties `TelemetryOptions` has so that if we add more in future a test fails to act as a reminder to ensure the copy constructor is implemented correctly?

I've been hit my this sort of thing recently 😄: https://github.com/domaindrivendev/Swashbuckle.AspNetCore/pull/2796

test/Polly.Extensions.Tests/Telemetry/TelemetryListenerImplTests.cs
martintmk: Done.

test/Polly.Extensions.Tests/Telemetry/TelemetryOptionsTests.cs
martintmk: Done.

src/Snippets/Docs/Telemetry.cs
martintmk: Done.

src/Snippets/Docs/Telemetry.cs
martintmk: Just more public area to cover :)

src/Snippets/Docs/Telemetry.cs
martintmk: Added `SeverityProviderArguments` that make this API more flexible and future-proof.

docs/advanced/telemetry.md
```diff
@@ -251,3 +251,34 @@ Each resilience strategy can generate telemetry data through the [`ResilienceStr
...
+        if (args.Event.EventName == "OnRetry")
+        {
+            // Decrease the severity of particular event.
+            return ResilienceEventSeverity.Debug;
+        }
+
+        return args.Event.Severity;
+    };
+
+    builder.AddTimeout(TimeSpan.FromSeconds(1));
```
peter-csala, line 278: This example seems a bit pointless because
- you have a single Timeout strategy
- and you override the `OnRetry` telemetry event's severity

Can you either override the `OnTimeout` or add a Retry strategy to the pipeline?

test/Polly.Extensions.Tests/Telemetry/TelemetryOptionsTests.cs
```diff
@@ -28,7 +28,7 @@ public void Ctor_EnsureDefaults()
     public void CopyCtor_Ok()
     {
         var options = new TelemetryOptions
-        {
+        {
```
martincostello:
```suggestion
        {
```

docs/advanced/telemetry.md
martintmk: Done.

martincostello (approved):

martincostello:
@martintmk Are you planning on any other PRs anytime soon? If not, once this is merged we can do an 8.4.0 to release this and the other recent changes.

pisethdanh:
Hi @martintmk,

Thank you for supporting this feature!

We currently configure Polly using the `AddStandardResilienceHandler` from the `Microsoft.Extensions.Http.Resilience` package:
e.g.
```csharp
public static IHttpStandardResiliencePipelineBuilder AddResilienceStrategies(this IHttpClientBuilder builder) => builder
        .AddStandardResilienceHandler()
        .Configure((options, serviceProvider) =>
        {
            // configure ...
        });
```

We add the standard resilience strategies to each typed http client.

Is it possible to set the event severity with the above approach?

martintmk:
You can configure global `TelemetryOptions` to configure the severity for all resilience pipelines.

```
var serviceCollection = new ServiceCollection()
    .Configure<TelemetryOptions>(options =>
    {
        // Configure options here
    });
```

PR #2097, opened 2024-05-05
Fix potential github action smells

ceddy4395:
# Pull Request

Hey! 🙂
I want to contribute the following changes to your workflow:

- Use 'if' for upload-artifact action
- Avoid uploading artifacts on forks
- Prevent running issue/PR actions on forks
- Avoid deploying jobs on forks

(These changes are part of a research Study at TU Delft looking at GitHub Action Smells. [Find out more](https://ceddy4395.github.io/research/gha-smells.html))

## Confirm the following

### DIFF
--- a/.github/workflows/after-release.yml
+++ b/.github/workflows/after-release.yml
@@ -16,7 +16,7 @@ jobs:
 
   update-changelog:
     runs-on: [ ubuntu-latest ]
-    if: github.event_name == 'release' && github.repository == 'App-vNext/Polly'
+    if: github.event_name == 'release' && github.event.repository.fork == false
 
     concurrency:
       group: '${{ github.workflow }}-changelog'
@@ -122,7 +122,7 @@ jobs:
     concurrency:
       group: '${{ github.workflow }}-milestone'
       cancel-in-progress: false
-    if: github.repository == 'App-vNext/Polly'
+    if: github.event.repository.fork == false
     permissions:
       issues: write
 
--- a/.github/workflows/after-release.yml
+++ b/.github/workflows/after-release.yml
@@ -16,7 +16,7 @@ jobs:
 
   update-changelog:
     runs-on: [ ubuntu-latest ]
-    if: github.event_name == 'release'
+    if: github.event_name == 'release' && github.repository == 'App-vNext/Polly'
 
     concurrency:
       group: '${{ github.workflow }}-changelog'
@@ -122,7 +122,7 @@ jobs:
     concurrency:
       group: '${{ github.workflow }}-milestone'
       cancel-in-progress: false
-
+    if: github.repository == 'App-vNext/Polly'
     permissions:
       issues: write
 
--- a/.github/workflows/after-release.yml
+++ b/.github/workflows/after-release.yml
@@ -118,11 +118,10 @@ jobs:
 
   update-milestone:
     runs-on: [ ubuntu-latest ]
-
+    if: github.event.repository.fork == false
     concurrency:
       group: '${{ github.workflow }}-milestone'
       cancel-in-progress: false
-    if: github.event.repository.fork == false
     permissions:
       issues: write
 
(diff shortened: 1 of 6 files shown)


.github/workflows/after-release.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
   update-changelog:
     runs-on: [ ubuntu-latest ]
-    if: github.event_name == 'release'
+    if: github.event_name == 'release' && github.repository == 'App-vNext/Polly'
```
martincostello, line 19: These could all be changed to `github.event.repository.fork == false` - then if the repo were ever moved/renamed, the semantics would remain the same without a need to update all the naming.

.github/workflows/build.yml
```diff
@@ -77,7 +77,7 @@ jobs:
       run: ./build.ps1
 
     - name: Upload Coverage Reports
-      if: always()
+      if: always() && github.repository == 'App-vNext/Polly'
```
martincostello, line 80: What's the benefit of not uploading the artifacts if someone decides to run this workflow in their fork?

.github/workflows/build.yml
```diff
@@ -107,7 +109,7 @@ jobs:
 
     - name: Upload signing file list
       uses: actions/upload-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
-      if: matrix.os_name == 'windows'
+      if: matrix.os_name == 'windows'
```
martincostello, line 112:
```suggestion
      if: matrix.os_name == 'windows'
```

gintsk:
@martincostello is it really neccessary to run `code-scan`, `ossf-scorecard` and `stale`, etc actions in forks?
It seems that after recent changes all github workflows are enabled for running in forks.

martincostello:
I didn't specifically disable it because my understanding was that GitHub automatically disables actions for new forks. This was something I _used to_ do in all my own repos, but I've stopped because I thought the default didn't need it anymore. See https://github.com/orgs/community/discussions/26704#discussioncomment-3252979

If that isn't the case we can look to turn it off explicitly, but if we do so I would want to code to do it the way I suggested in my review, not with hard-coding the repository name: https://github.com/App-vNext/Polly/pull/2097#discussion_r1590397248

.github/workflows/build.yml
ceddy4395: I would consider it to be more efficient in terms of runtime and space used. Furthermore, especially for the codedov uploads which will fail for sure because they don't have access to the secrets.

.github/workflows/after-release.yml
ceddy4395: That's a very good point, thank you for the feedback! :smile:

.github/workflows/build.yml
martincostello: This step doesn't upload anything to codecov - they're being uploaded to the artifacts of the workflow run itself.

.github/workflows/after-release.yml
```diff
@@ -122,7 +122,7 @@ jobs:
     concurrency:
       group: '${{ github.workflow }}-milestone'
       cancel-in-progress: false
-
+    if: github.event.repository.fork == false
```
martincostello, line 125: For consistency, please place after `runs-on`.

.github/workflows/build.yml
```diff
@@ -86,19 +86,21 @@ jobs:
 
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@5ecb98a3c6b747ed38dc09f787459979aebb39be # v4.3.1
+      if: always() && github.event.repository.fork == false
```
martincostello, line 89: Please revert all the changes except this one - if someone wants to run our builds in their fork, then they should be able to access the artifacts produced by them.

martincostello (approved):

PR #2111, opened 2024-05-15
Fix S3872 warning

iamdmitrij:
# Pull Request

#1290

- [x] Fix [S3872](https://sonarsource.github.io/rspec/#/rspec/S3872/csharp) by renaming `timeout` parameter to `duration` in `Policy.Timeout` APIs
- [x] Rename `timeout` -> `duration` in `Policy.TimeoutAsync` async APIs to unify naming

## Confirm the following

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -18,6 +18,10 @@ indent_style = space
 indent_size = 2
 tab_width = 2
 
+# S3872 // Interfaces should not be empty
+[**/Timeout/{TimeoutSyntax,TimeoutTResultSyntax}.cs]
+dotnet_diagnostic.S3872.severity = suggestion
+
 # C# files
 [*.cs]
 
--- a/src/Polly/Polly.csproj
+++ b/src/Polly/Polly.csproj
@@ -7,7 +7,7 @@
     <ProjectType>Library</ProjectType>
     <MutationScore>70</MutationScore>
     <IncludePollyUsings>true</IncludePollyUsings>
-    <NoWarn>$(NoWarn);S3872;SA1414;S3215</NoWarn>
+    <NoWarn>$(NoWarn);SA1414;S3215</NoWarn>
     <NoWarn>$(NoWarn);IDE1006;CA1062;CA1068;S4039;CA1063;CA1031;CA1051</NoWarn>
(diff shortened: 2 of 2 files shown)


martincostello (changes requested):
We can't accept any change that requires you touching the `PublicAPI.Shipped.txt` files, as those would be breaking changes.

You would need to suppress any warnings generated instead.

iamdmitrij:
Ok, I've disabled this rule only for impacted classes in `.editoconfig`. If new major versions gets released, it'd be good idea to fix it permanently.

.editorconfig
```diff
@@ -18,6 +18,10 @@ indent_style = space
 indent_size = 2
 tab_width = 2
 
+# S3872 // Interfaces should not be empty
```
martincostello, line 21: This doesn't appear to be the right description for what the original change was?

It's better (IMHO) to suppress these in the code with `#pragma warning disable/restore` pairs at the specific points of usage, rather the hiding it away in the editorconfig.

iamdmitrij: You are right, in this case we don't have many violations for this rule, so it makes sense to keep it in code. Also, it can serve as a reminder to fix it in the future.

martincostello (approved):

src/Polly/Timeout/TimeoutSyntax.cs
```diff
@@ -109,7 +109,9 @@ public static TimeoutPolicy Timeout(int seconds, TimeoutStrategy timeoutStrategy
     /// <param name="timeout">The timeout.</param>
     /// <returns>The policy instance.</returns>
     /// <exception cref="ArgumentOutOfRangeException">timeout;Value must be a positive TimeSpan (or Timeout.InfiniteTimeSpan to indicate no timeout).</exception>
+#pragma warning disable S3872
```
martincostello, line 112: The build failures are annoying - I've seen them before where dotnet format doesn't complain locally but does in CI. I think to fix them you need to move the disable to above the /// comments.

iamdmitrij: Thanks, works like a charm now.

martincostello:
Thanks again!

PR #2172, opened 2024-07-03
Fix SA1414 warning 

Zombach:
# Pull Request

[#1290](https://github.com/App-vNext/Polly/issues/1290)

 * [x]  Suppress [SA1414](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1414.md) in the code or fix the warning

## Confirm the following

### DIFF
--- a/src/Polly/Caching/IAsyncCacheProvider.cs
+++ b/src/Polly/Caching/IAsyncCacheProvider.cs
@@ -16,7 +16,7 @@ public interface IAsyncCacheProvider
     /// A <see cref="Task{TResult}" /> promising as Result a tuple whose first element is a value indicating whether
     /// the key was found in the cache, and whose second element is the value from the cache (null if not found).
     /// </returns>
-    Task<(bool, object?)> TryGetAsync(string key, CancellationToken cancellationToken, bool continueOnCapturedContext);
+    Task<(bool CacheHit, object? Result)> TryGetAsync(string key, CancellationToken cancellationToken, bool continueOnCapturedContext);
 
     /// <summary>
     /// Puts the specified value in the cache asynchronously.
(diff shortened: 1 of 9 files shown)


src/Polly/Caching/IAsyncCacheProvider.cs
```diff
@@ -16,7 +16,7 @@ public interface IAsyncCacheProvider
     /// A <see cref="Task{TResult}" /> promising as Result a tuple whose first element is a value indicating whether
     /// the key was found in the cache, and whose second element is the value from the cache (null if not found).
     /// </returns>
-    Task<(bool, object?)> TryGetAsync(string key, CancellationToken cancellationToken, bool continueOnCapturedContext);
+    Task<(bool CacheHit, object? Result)> TryGetAsync(string key, CancellationToken cancellationToken, bool continueOnCapturedContext);
```
martincostello, line 19: We can't make changes like this to the public interface. This is why the build is failing, because the public API no longer matches the baseline.

For public members, the warning just has to be suppressed. Otherwise it would be a breaking change for users.

src/Polly/PublicAPI.Shipped.txt
martincostello, line 1: As noted in my previous comment, all these changes should be reverted.

If we wanted to make them (which we don't), they would go in the `Unshipped` file because these members don't exist in any shipped version.

src/Polly/Caching/IAsyncCacheProvider.cs
Zombach: Really. I've done something stupid

martincostello (approved):

PR #2174, opened 2024-07-03
Fix CA1822 warning 

Zombach:
# Pull Request

[#1290](https://github.com/App-vNext/Polly/issues/1290)

 * [x]  Suppress [CA1822](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1822) in the code or fix the warning

## Confirm the following

There are also two questions.
Maybe it's better instead of all this.
Separately leave <NoWarn>$(NoWarn);CA1822</NoWarn> in .csproj and add a comment about what not to touch,
// We do not need to fix these as it would break, BenchmarkDotNet supports only non-static, non-sealed public types with instance methods marked as [Benchmark]

And the second one. I noticed that all lambda methods can use a group of methods.
Can I do this and are these changes necessary?

Before
image
After
image


martincostello:
Based on [other benchmarks](https://github.com/App-vNext/Polly/blob/main/bench/Polly.Core.Benchmarks/CreationBenchmark.cs) to looks like `static` might actually work.

However, if it doesn't then I think overall it's better to leave these as-is, as otherwise we're effectively turning it off in every file in the project, but more verbosely.

Zombach:
> Based on [other benchmarks](https://github.com/App-vNext/Polly/blob/main/bench/Polly.Core.Benchmarks/CreationBenchmark.cs) to looks like `static` might actually work.
> 
> However, if it doesn't then I think overall it's better to leave these as-is, as otherwise we're effectively turning it off in every file in the project, but more verbosely.

I'll check it now. 

Most likely the problem is that the class itself cannot be static. That's why there is suppression 1052

Zombach:
> Based on [other benchmarks](https://github.com/App-vNext/Polly/blob/main/bench/Polly.Core.Benchmarks/CreationBenchmark.cs) to looks like `static` might actually work.
> 
> However, if it doesn't then I think overall it's better to leave these as-is, as otherwise we're effectively turning it off in every file in the project, but more verbosely.

However, this correction cannot be made.

image

martincostello:
As adding file-level pragmas is more noise than just the `NoWarn` in the case of this project, I don't think this specific change is making a net-improvement.

Zombach:
> As adding file-level pragmas is more noise than just the `NoWarn` in the case of this project, I don't think this specific change is making a net-improvement.

I completely agree.

Then this option will suit you?

Or I'm closing the PR

martincostello:
Thanks for the attempt, but in this case I think it's better to leave things as-is.

I'll update the linked issue to scope it to just the unit test class so someone isn't lead to try and change this project as the benchmarks are effectively "done" with respect to the refactorings (i.e. those suppressed are intentional).

martincostello:
https://github.com/App-vNext/Polly/issues/1290#issuecomment-2208393227

Issue #2215, opened 2024-07-19
[Cleanup the Polly codebase] Warning CA1062 


Zombach:
# Checklist

## The issue or feature being addressed

[#1290 Cleanup codebase](https://github.com/App-vNext/Polly/issues/1290)
Namely suppress [CA1062 Validate arguments](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1062) in the code or fix the warning

## Details on the issue fix or feature implementation

Since the task is big, 
This must be done in several stages. 
The first stage is warning suppression.
Then each next PR - processing of one file and adding tests

- [x] [Warning Suppression CA1062](https://github.com/App-vNext/Polly/pull/2216)
- [x] [src/Polly/Bulkhead/AsyncBulkheadPolicy.cs](https://github.com/App-vNext/Polly/pull/2237)
- [x] [src/Polly/Bulkhead/BulkheadPolicy.cs](https://github.com/App-vNext/Polly/pull/2236)
- [x] [src/Polly/Caching/AsyncCachePolicy.cs](https://github.com/App-vNext/Polly/pull/2239)
- [x] [src/Polly/Caching/AsyncCacheSyntax.cs](https://github.com/App-vNext/Polly/pull/2242)
- [x] [src/Polly/Caching/AsyncCacheTResultSyntax.cs](https://github.com/App-vNext/Polly/pull/2243)
- [x] [src/Polly/Caching/CachePolicy.cs](https://github.com/App-vNext/Polly/pull/2238)
- [x] [src/Polly/Caching/CacheSyntax.cs](https://github.com/App-vNext/Polly/pull/2240)
- [x] [src/Polly/Caching/CacheTResultSyntax.cs](https://github.com/App-vNext/Polly/pull/2241)
- [x] [src/Polly/Caching/ContextualTtl.cs](https://github.com/App-vNext/Polly/pull/2217)
- [x] [src/Polly/Caching/DefaultCacheKeyStrategy.cs](https://github.com/App-vNext/Polly/pull/2218)
- [x] [src/Polly/CircuitBreaker/AsyncCircuitBreakerPolicy.cs](https://github.com/App-vNext/Polly/pull/2235)
- [x] [src/Polly/CircuitBreaker/CircuitBreakerPolicy.cs](https://github.com/App-vNext/Polly/pull/2234)
[...]

PR #2216, opened 2024-07-19
Warning Suppression CA1062

Zombach:
# Pull Request

[#2215](https://github.com/App-vNext/Polly/issues/2215)

 * [x]  Suppress [CA1062](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1062) in the code or fix the warning

## Confirm the following

- [x] Preparing to solve warning CA1062.

### DIFF
--- a/src/Polly/Bulkhead/AsyncBulkheadPolicy.cs
+++ b/src/Polly/Bulkhead/AsyncBulkheadPolicy.cs
@@ -4,6 +4,7 @@ namespace Polly.Bulkhead;
 /// <summary>
 /// A bulkhead-isolation policy which can be applied to delegates.
 /// </summary>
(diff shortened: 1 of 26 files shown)


martincostello:
Why suppress rather than fix in this case?

Zombach:
> Why suppress rather than fix in this case?

The problem is that there is a lot of this warning. And if you do not suppress this warning in advance. If you disable this warning, a lot of noise is created. Which is very much in the way. And when correcting the warning, it is necessary to additionally cover the code with tests. Which in turn amounts to a huge PR of 1000+ lines. That's why I divided the task into subtasks, as described in #2215 . And now anyone can take a file as a separate file, disable warning suppression in it and fully work it out. Without making too much noise.

martincostello (approved):

PR #2217, opened 2024-07-20
Fix warning CA1062#ContextualTtl

Zombach:
# Pull Request

[#2215](https://github.com/App-vNext/Polly/issues/2215)

*  [x]  Suppress [CA1062](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1062) in src/Polly/Caching/ContextualTtl.cs or fix the warning

## Confirm the following

### DIFF
--- a/src/Polly/Caching/ContextualTtl.cs
+++ b/src/Polly/Caching/ContextualTtl.cs
@@ -4,7 +4,6 @@ namespace Polly.Caching;
 /// <summary>
 /// Defines a ttl strategy which will cache items for a TimeSpan which may be influenced by data in the execution context.
 /// </summary>
-#pragma warning disable CA1062 // Validate arguments of public methods
 public class ContextualTtl : ITtlStrategy
 {
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/Polly/Caching/ContextualTtl.cs
```diff
@@ -28,6 +27,11 @@ public class ContextualTtl : ITtlStrategy
     /// <returns>TimeSpan.</returns>
     public Ttl GetTtl(Context context, object? result)
     {
+        if (context is null)
```
Zombach, line 30: @martincostello
Which is better to use? "==" or the pattern "is"
The "is" mapping pattern works faster, but mostly the code uses "=="

martincostello: I would use `is` for new code where it is equivalent to `== null`. Don't worry about updating existing code to match - it was written before `is` was a thing.

martincostello (approved):

PR #2245, opened 2024-07-23
Remove code analysis hack

martincostello:
Try and repro dotnet/roslyn#70368 again.


martincostello:
Couldn't repro anymore after several attempts, so going to merge. We can always add the workaround back if it happens again.

PR #2253, opened 2024-07-29
Fix S4225 warnings of extension methods for object class

sukreshmanda:
# Pull Request

This includes warning suppression for IDE0022 as it is blocking the build

https://github.com/App-vNext/Polly/issues/1290

- Adding extension methods on object type is not recommended . 
- Using **_Dictionary<string, object>_** will solve the issue. So no need for **_test/Polly.Specs/Helpers/ObjectExtensions.cs_**

## Confirm the following


test/Polly.Specs/Helpers/ObjectExtensions.cs
```diff
@@ -1,11 +0,0 @@
-﻿namespace Polly.Specs.Helpers;
-
-public static class ObjectExtensions
```
martincostello, line 3: I think this change as-is is undoing the intent of the helper method, which is to avoid the repetitive code to create dictionaries.

Instead maybe add a _different_ extension method that achieves the same effect.

Something like this:

```csharp
public static class DictionaryHelpers
{
    public static Dictionary<TKey, TValue> Create<TKey, TValue>(TKey key, TValue value)
        => new Dictionary<TKey, TValue>(1) { [key] = value };
}
```

test/Polly.Specs/Caching/CacheAsyncSpecs.cs
```diff
@@ -1,4 +1,6 @@
-﻿namespace Polly.Specs.Caching;
+﻿using static Polly.Specs.DictionaryHelpers;
```
martincostello, line 1: If you made this a static global using for this project, we wouldn't need to add this to all these files:

```xml
<ItemGroup>
  <Using Include="Polly.Specs.DictionaryHelpers" Static="true" />
</ItemGroup>
```

test/Polly.Specs/DictionaryHelpers.cs
```diff
@@ -0,0 +1,10 @@
+namespace Polly.Specs;
+
+public static class DictionaryHelpers
+{
+    public static Dictionary<TKey, object> CreateDictionary<TKey>(TKey key, object value)
+        where TKey : notnull => new() { { key, value } };
+
+    public static Dictionary<TKey, object> CreateDictionary<TKey>(TKey key1, object value1, TKey key2, object value2)
+        where TKey : notnull => new() { { key1, value1 }, { key2, value2 } };
```
martincostello, line 9: I think the newer indexer syntax is nicer:

```suggestion
        where TKey : notnull => new() { [key] = value };

    public static Dictionary<TKey, object> CreateDictionary<TKey>(TKey key1, object value1, TKey key2, object value2)
        where TKey : notnull => new() { [key1] = value1, [key2] = value2 };
```

sukreshmanda: Done

test/Polly.Specs/Caching/CacheAsyncSpecs.cs
sukreshmanda: Done

test/Polly.Specs/Polly.Specs.csproj
```diff
@@ -27,6 +27,7 @@
     <Using Include="Xunit" />
     <Using Include="Xunit.Abstractions" />
     <Using Include="Xunit.Sdk" />
+    <Using Include="Polly.Specs.DictionaryHelpers" Static="true" />
```
martincostello, line 30: Could you put this in the appropriate place in the sort please?

test/Polly.Specs/CircuitBreaker/AdvancedCircuitBreakerAsyncSpecs.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Scenario = Polly.Specs.Helpers.PolicyExtensionsAsync.ExceptionAndOrCancellationScenario;
+﻿using static Polly.Specs.DictionaryHelpers;
```
martincostello, line 1: There's still a bunch of these that can be removed - otherwise this looks much neater 👍

test/Polly.Specs/Polly.Specs.csproj
sukreshmanda: Are they in sorted order by name?

martincostello: Yes - if you look at the `ItemGroup` you see that they go `F => N => P => X` in alphabetical order by the value of `Include`.

sukreshmanda: Got it thanks

martincostello (approved):
Thanks!

PR #2254, opened 2024-07-31
Use [DebuggerDisableUserUnhandledExceptions]

martincostello:
- Add usage of `[DebuggerDisableUserUnhandledExceptions]` in `StrategyHelper` and `RetryEngine` to avoid newer versions of the Visual Studio debugger for breaking for exceptions we are intentionally handling.
- Add copy of `[DebuggerDisableUserUnhandledExceptions]` for use in downlevel versions of .NET that do not contain the attribute.

[_Break for Async User-Unhandled exceptions in the Visual Studio Debugger_](https://devblogs.microsoft.com/visualstudio/break-for-async-user-unhandled-exceptions-in-the-visual-studio-debugger/)

/cc @halter73 @gregg-miskelly

### DIFF
--- a/src/Polly.Core/Hedging/Controller/TaskExecution.cs
+++ b/src/Polly.Core/Hedging/Controller/TaskExecution.cs
@@ -199,6 +199,7 @@ internal sealed class TaskExecution<T>
         _stopExecutionTimestamp = 0;
     }
 
+    [DebuggerDisableUserUnhandledExceptions]
     private async Task ExecuteSecondaryActionAsync(Func<ValueTask<Outcome<T>>> action)
     {
         Outcome<T> outcome;
@@ -218,6 +219,7 @@ internal sealed class TaskExecution<T>
 
     private async Task ExecuteCreateActionException(Exception e) => await UpdateOutcomeAsync(Polly.Outcome.FromException<T>(e)).ConfigureAwait(Context.ContinueOnCapturedContext);
 
+    [DebuggerDisableUserUnhandledExceptions]
     private async Task ExecutePrimaryActionAsync<TState>(Func<ResilienceContext, TState, ValueTask<Outcome<T>>> primaryCallback, TState state)
     {
         Outcome<T> outcome;
--- a/src/Polly.Core/Utils/StrategyHelper.cs
+++ b/src/Polly.Core/Utils/StrategyHelper.cs
@@ -4,6 +4,7 @@
 
 internal static class StrategyHelper
 {
+    [DebuggerDisableUserUnhandledExceptions]
     public static ValueTask<Outcome<TResult>> ExecuteCallbackSafeAsync<TResult, TState>(
         Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> callback,
         ResilienceContext context,
@@ -29,6 +30,7 @@ internal static class StrategyHelper
             return new ValueTask<Outcome<TResult>>(Outcome.FromException<TResult>(e));
         }
 
+        [DebuggerDisableUserUnhandledExceptions]
         static async ValueTask<Outcome<T>> AwaitTask<T>(ValueTask<Outcome<T>> task, bool continueOnCapturedContext)
         {
             try
--- a/src/Polly/Caching/AsyncCacheEngine.cs
+++ b/src/Polly/Caching/AsyncCacheEngine.cs
@@ -3,6 +3,7 @@ namespace Polly.Caching;
 
 internal static class AsyncCacheEngine
 {
+    [DebuggerDisableUserUnhandledExceptions]
     internal static async Task<TResult> ImplementationAsync<TResult>(
         IAsyncCacheProvider<TResult> cacheProvider,
         ITtlStrategy<TResult> ttlStrategy,
--- a/src/Polly/Retry/RetryEngine.cs
+++ b/src/Polly/Retry/RetryEngine.cs
@@ -3,6 +3,7 @@ namespace Polly.Retry;
 
 internal static class RetryEngine
 {
+    [DebuggerDisableUserUnhandledExceptions]
     internal static TResult Implementation<TResult>(
         Func<Context, CancellationToken, TResult> action,
         Context context,
--- a/src/Polly.Core/Outcome.TResult.cs
+++ b/src/Polly.Core/Outcome.TResult.cs
@@ -90,5 +90,4 @@ public readonly struct Outcome<TResult>
         ExceptionDispatchInfo?.Throw();
         return Result!;
     }
-
 }
--- a/src/Polly.Core/Polly.Core.csproj
+++ b/src/Polly.Core/Polly.Core.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net9.0;net8.0;net6.0;netstandard2.0;net472;net462</TargetFrameworks>
+    <TargetFrameworks>net8.0;net6.0;netstandard2.0;net472;net462</TargetFrameworks>
     <AssemblyTitle>Polly.Core</AssemblyTitle>
     <RootNamespace>Polly</RootNamespace>
     <Nullable>enable</Nullable>
--- a/src/Polly.Core/Polly.Core.csproj
+++ b/src/Polly.Core/Polly.Core.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net8.0;net6.0;netstandard2.0;net472;net462</TargetFrameworks>
+    <TargetFrameworks>net9.0;net8.0;net6.0;netstandard2.0;net472;net462</TargetFrameworks>
     <AssemblyTitle>Polly.Core</AssemblyTitle>
     <RootNamespace>Polly</RootNamespace>
     <Nullable>enable</Nullable>
@@ -34,6 +34,10 @@
     <PackageReference Include="System.Threading.Tasks.Extensions" Condition="!$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'netcoreapp3.1'))" />
     <PackageReference Include="System.ValueTuple" Condition="$([MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)')) == '.NETFramework'" />
     <PackageReference Include="System.ComponentModel.Annotations" Condition="!$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'netcoreapp3.1'))" />
+  </ItemGroup>
+
+  <ItemGroup Condition="!$([MSBuild]::IsTargetFrameworkCompatible($(TargetFramework), 'net9.0'))">
+    <Compile Include="$(MSBuildThisFileDirectory)..\Shared\DebuggerDisableUserUnhandledExceptionsAttribute.cs" Link="DebuggerDisableUserUnhandledExceptionsAttribute.cs" />
   </ItemGroup>
 
 </Project>
--- a/src/Polly.Core/ResilienceContext.cs
+++ b/src/Polly.Core/ResilienceContext.cs
@@ -74,9 +74,11 @@ public sealed class ResilienceContext
         Properties.AddOrReplaceProperties(context.Properties);
     }
 
+#pragma warning disable S3236 // Remove this argument from the method call; it hides the caller information.
     [ExcludeFromCodeCoverage]
     [Conditional("DEBUG")]
     internal void AssertInitialized() => Debug.Assert(IsInitialized, "The resilience context is not initialized.");
+#pragma warning restore S3236 // Remove this argument from the method call; it hides the caller information.
 
     internal ResilienceContext Initialize<TResult>(bool isSynchronous)
     {
--- a/src/Polly.Core/ResiliencePipeline.Async.cs
+++ b/src/Polly.Core/ResiliencePipeline.Async.cs
@@ -25,7 +25,7 @@ public partial class ResiliencePipeline
         InitializeAsyncContext(context);
 
         var outcome = await Component.ExecuteCore(
-            static async (context, state) =>
+            [DebuggerDisableUserUnhandledExceptions] static async (context, state) =>
             {
                 try
                 {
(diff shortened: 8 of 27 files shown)


src/Polly.Core/Utils/StrategyHelper.cs
```diff
@@ -4,6 +4,10 @@
 
 internal static class StrategyHelper
 {
+    [DebuggerDisableUserUnhandledExceptions]
+#if !DEBUG
+    [DebuggerNonUserCode]
```
gregg-miskelly, line 9: You don't need `[DebuggerNonUserCode]`. Sorry, I should have added a comment to my sample -- that was only needed in my sample because I was adding code to a single project and I wanted the debugger to treat some of it as if it was library code.

src/Polly.Core/Utils/StrategyHelper.cs
```diff
@@ -4,6 +4,10 @@
...
+#endif
```
martincostello, line 10:
```suggestion
```

src/Polly/Retry/RetryEngine.cs
```diff
@@ -3,6 +3,10 @@ namespace Polly.Retry;
 
 internal static class RetryEngine
 {
+    [DebuggerDisableUserUnhandledExceptions]
+#if !DEBUG
+    [DebuggerNonUserCode]
+#endif
```
martincostello, line 9:
```suggestion
```

gregg-miskelly:
When I try this locally, in the sync case, I am seeing the exception caught in the various lambdas such as this one: https://github.com/App-vNext/Polly/blob/8a8f3d8fa167f3c6da22bd6c93e9c8f7d5e4226d/src/Polly.Core/ResiliencePipeline.Sync.cs#L27

gregg-miskelly:
I am seeing something similar in async where it is the lambda's in ResiliencePipeline.Async.cs that is actually calling the user's delegate and catching the exception, so that is where the `[DebuggerDisableUserUnhandledExceptions]` attribute needs to go.

martincostello:
Thanks for hunting those down. I didn't do a proper dig through the code (clearly) find the closes place to the user's original code.

Tomorrow I'll look at moving the attributes around.

martincostello:
Using a search for `catch (Exception ` as a guide, I've added a bunch of additional attributes to various methods.

Are there any other heuristics I can use to find likely useful places for this? Otherwise I'll need to rely on you both to see if it's behaving how you'd expect 😄

src/Polly.Core/Hedging/Controller/TaskExecution.cs
```diff
@@ -155,7 +155,7 @@ private async Task HandleOnHedgingAsync(ResilienceContext primaryContext, int at
         Func<ResilienceContext, TState, ValueTask<Outcome<TResult>>> primaryCallback,
         ResilienceContext primaryContext,
         TState state,
-        int attempt) => new(primaryContext, Context, attempt, (context) => primaryCallback(context, state));
+        int attempt) => new(primaryContext, Context, attempt, [DebuggerDisableUserUnhandledExceptions] (context) => primaryCallback(context, state));
```
gregg-miskelly, line 158: You shouldn't need it here since there is no `catch`

martincostello, line 158:
```suggestion
        int attempt) => new(primaryContext, Context, attempt, (context) => primaryCallback(context, state));
```

src/Polly/Bulkhead/AsyncBulkheadEngine.cs
```diff
@@ -3,6 +3,7 @@ namespace Polly.Bulkhead;
 
 internal static class AsyncBulkheadEngine
 {
+    [DebuggerDisableUserUnhandledExceptions]
```
gregg-miskelly, line 6: >[DebuggerDisableUserUnhandledExceptions] [](http://example.com/codeflow?start=4&length=40)

You don't need this since there is no catch

src/Polly/Bulkhead/BulkheadEngine.cs
```diff
@@ -3,6 +3,7 @@ namespace Polly.Bulkhead;
 
 internal static class BulkheadEngine
 {
+    [DebuggerDisableUserUnhandledExceptions]
```
gregg-miskelly, line 6: >[DebuggerDisableUserUnhandledExceptions] [](http://example.com/codeflow?start=4&length=40)

Same

src/Polly/Caching/AsyncCacheEngine.cs
```diff
@@ -3,6 +3,7 @@ namespace Polly.Caching;
 
 internal static class AsyncCacheEngine
 {
+    [DebuggerDisableUserUnhandledExceptions]
```
gregg-miskelly, line 6: >[DebuggerDisableUserUnhandledExceptions] [](http://example.com/codeflow?start=4&length=40)

Sorry if this is obvious, but just to be clear -- the reason to have this here would be if you expect `cacheProvider` or possibly `onCachePut` to be implemented in user code, and exceptions coming out of user code should be considered ignorable, such as if they will be retried. If that isn't the case, you don't need this even though there is a `catch`

gregg-miskelly:
> Using a search for `catch (Exception ` as a guide, I've added a bunch of additional attributes to various methods.

Searching for `catch (Exception ` is a great way to start. `DebuggerDisableUserUnhandledExceptionsAttribute` only makes sense in methods with a `catch` and I would be surprised if there are any methods in Polly that catches a specific exception that would need this attribute. 

Once you have found a method like that, then the follow up question is if there is a way for the exception being caught to be one which (1) is from user code -and- (2) should usually be ignored as Polly is very likely to retry the operation.
 
If you want to be able to test this work, Visual Studio 2022 version 17.12 preview 1 will support `DebuggerDisableUserUnhandledExceptionsAttribute`. 17.12 preview 1 isn't public yet, but it should be fairly soon. If you aren't familiar with Visual Studio previews, you can install them side by side your regular builds from [the preview channel](https://visualstudio.microsoft.com/vs/preview/).

You can also somewhat test things today thusly --
* In cases where the exception is thrown by a synchronous user method, you can test this with any version of .NET and Visual Studio.
* In cases where the exception is thrown by an asynchronous user method, you can test this if you use a test app running on .NET 9 preview 6 + Visual Studio 2022 version 17.11, which can be obtained from the preview channel.

Sorry if this is obvious, but the basic way to test is:
1. Create a new test executable that supplies a call back method that throws
2. Ensure that Debug->Options->Enable Just My Code is **checked**.
3. Ensure that the test app is using a Release-configuration version of Polly
4. Run the test app under the debugger

If you are using 17.11 (async) / 17.11 or earlier (sync) you should see the debugger break "User-Unhandled". You want to make sure you have added `DebuggerDisableUserUnhandledExceptions` to whichever frame on the call stack has the `catch` block that will first catch the exception.

If you are using 17.12, you want to see that the debugger no longer breaks user unhandled.

I hope this helps and thanks so much for working on this!

src/Polly/Caching/AsyncCacheEngine.cs
martincostello: All are user-configurable:

https://github.com/App-vNext/Polly/blob/2ff84c8db3b4e6268bd9424a262365c3346c93ed/src/Polly/Caching/AsyncCacheSyntax.cs#L146-L172

Whether they're ignorable would depend on how the user has composed their resilience policy (for example, wrapping a cache policy with retries).

src/Polly/NoOp/NoOpEngineAsync.cs
```diff
@@ -3,6 +3,7 @@ namespace Polly.NoOp;
 
 internal static partial class NoOpEngine
 {
+    [DebuggerDisableUserUnhandledExceptions]
```
martincostello, line 6:
```suggestion
```

martincostello:
@gregg-miskelly Thanks for the information. I already have previews installed, so I think I'll just wait until I get updated to 17.12 and then run out test suites inside the IDE in Release mode and see where the debugger breaks (or not).

gregg-miskelly (approved):

gregg-miskelly:
> @gregg-miskelly Thanks for the information. I already have previews installed, so I think I'll just wait until I get updated to 17.12 and then run out test suites inside the IDE in Release mode and see where the debugger breaks (or not).

FYI 17.12 preview 1 was released today

gregg-miskelly (approved):

martincostello:
@gregg-miskelly I was having a play around with this branch in 17.11 and 17.12 and I didn't really notice any difference with and without the attribute.

I'm probably doing something stupid, but I followed the instructions you posted the other week.

Do you have a simple GitHub repo that I could clone that shows the desired behavioural difference so I can check I've got things configured properly before I re-test with Polly?

gregg-miskelly:
@martincostello I am sorry. Unfortunately, the preview 1 version of this work has a rather large bug that we are working to fix for preview 2. If you want to work around it rather than waiting for preview 2 --
1. Start debugging, and hit a breakpoint before the exception is raised
2. Open the modules window
3. Load symbols for the module containing the `[DebuggerDisableUserUnhandledExceptions]`
4. Verify the value of 'Optimized', 'User Code', and 'Symbol Status' look like this:

image

martincostello:
Just come back to this, and I'm still failing to see a difference in behaviour with `17.12.0 Preview 2.0`. I used the [Polly-Samples](https://github.com/App-vNext/Polly-Samples) repo and ran the [EFCore](https://github.com/App-vNext/Polly-Samples/blob/main/PollyDemos/Demo16_EntityFramework-WithRetryNTimes.cs) sample.

My understanding was that with Polly 8.4.0, Visual Studio would break [here](https://github.com/App-vNext/Polly-Samples/blob/d8b8633a988f7c505dcbbee32afa47bdad3c835d/PollyDemos/Demo16_EntityFramework-WithRetryNTimes.cs#L60), but with the changes from this branch it wouldn't. It practice, I'm seeing the same behaviour with both - Visual Studio doesn't break and I just see the exceptions logged to the console.

I'm consuming Polly via a local NuGet package source using the `.nupkg` files out of CI so it's more representative of what a user would experience (rather than Polly locally compiled). I disabled Just My Code to step into Polly.Core, and then the Modules window shows this:

image

I feel like I must be missing something here.

Could you share a simple GitHub project that's set up in a way that should illustrate the difference? Then I can pull this branch's code into it and verify if I see the same.

gregg-miskelly:
> My understanding was that with Polly 8.4.0, Visual Studio would break [here](https://github.com/App-vNext/Polly-Samples/blob/d8b8633a988f7c505dcbbee32afa47bdad3c835d/PollyDemos/Demo16_EntityFramework-WithRetryNTimes.cs#L60), but with the changes from this branch it wouldn't. 

Your understanding of the expected behavior is correct. What version of the .NET Runtime is the process you are debugging using? For async-user unhandled, you need to modify the client to use .NET 9. Also, if you are on .NET 9, do you have Just My Code enabled?

martincostello:
Ah, that's probably what I missed - I didn't update the sample to .NET 9 and was just using the CI build for net8.0 with the copy of the attribute in it. I'll update the host project tomorrow and try again.

martincostello:
Yep, that's what it was.

Using https://github.com/App-vNext/Polly-Samples/pull/107, the debugger breaks [here](https://github.com/App-vNext/Polly-Samples/blob/c5c8fecc05dec5737760cea6d74f80fed3c0b74f/PollyDemos/Demo16_EntityFramework-WithRetryNTimes.cs#L60) when an exception is thrown now when the app is targeting `net9.0`. With `net8.0`, it just gets logged to the console and the debugger doesn't break.

That suggests to me that I'm missing one or more attributes as it should be being ignored and not breaking as it's being retried?

martincostello:
This is what I see in Visual Studio:

image

gregg-miskelly:
> That suggests to me that I'm missing one or more attributes as it should be being ignored and not breaking as it's being retried?

Agreed. If show external code, and inspect the frames on the call stack, you can hopefully find the frame with the `catch` block.

martincostello:
@gregg-miskelly I thought I'd found one I'd missed, but ingesting that change still doesn't seem to have changed the behaviour - would you mind cloning https://github.com/App-vNext/Polly-Samples/pull/107 locally and taking a look to see what I'm missing?

gregg-miskelly:
@martincostello Thanks for continuing to work on this!

The attribute usage in Polly is correct. The issue is that there is a non-user async method on the stack above Polly's catch handler (Microsoft.EntityFrameworkCore.dll!Microsoft.EntityFrameworkCore.ExecutionStrategyExtensions.ExecuteAsync.AnonymousMethod__4_0). Any async method will have a compiler-generated catch handler. In hindsight this should have been obvious to us, but we didn't think about those compiler-generated catch blocks when we added `DebuggerDisableUserUnhandledExceptions` support. So the debugger sees the catch handler from EntityFramework and thinks this exception should be user-unhandled. We think it shouldn't be too hard to teach the debugger to ignore these catch handlers and let the exception continue propagating. We hope we will be able to fix this for Preview 3, but that is shutting down soon so we will see.

Thanks again and sorry for all the problems.

martincostello:
No worries, that's the point of testing 😄

Let me know when a fix makes it into a public preview of 17.12 and then I'll try again.

martincostello:
One more question: do we need to use the attribute in methods with `finally` blocks without a `catch`?

Running the other demos in the same project, there's another one that is breaking in the debugger that has [this method on the stack](https://github.com/martincostello/Polly/blob/52ff7603487bf20037b4819b6047dcfd5c41b32d/src/Polly.Core/ResiliencePipeline.AsyncT.cs#L160-L191) - do we need to mark those methods with the attribute too, or is this similar to the EFCore case?

gregg-miskelly:
> One more question: do we need to use the attribute in methods with `finally` blocks without a `catch`?

No, you only need to worry about `catch` blocks.

> Running the other demos in the same project, there's another one that is breaking in the debugger that has [this method on the stack](https://github.com/martincostello/Polly/blob/52ff7603487bf20037b4819b6047dcfd5c41b32d/src/Polly.Core/ResiliencePipeline.AsyncT.cs#L160-L191) - do we need to mark those methods with the attribute too, or is this similar to the EFCore case?

It looks like the same issue as the EFCore case.

gregg-miskelly:
FYI: The fix to better handle intermediate async methods made it in time for VS 17.12 Preview 3.

martincostello:
Looks like things are now working as expected with 17.12.0 Preview 3.0.

Will try with `net8.0` Polly with `net9.0` samples to see if the duck-typed attribute works next - if it does, then we don't need Polly to add a `net9.0` TFM.

martincostello:
All working as expected with the latest public preview of Visual Studio. These changes will be included in the 8.5.0 release we're aiming to publish next month once the 9.0.100 SDK is available.

PR #2255, opened 2024-07-31
Re-enable failing tests

martincostello:
Disable TFM parallelism for `dotnet test` in the Polly.Core project to see if it is the cause of the two failing tests.

See [_What's new in the SDK for .NET 9 - Run tests in parallel_](https://learn.microsoft.com/dotnet/core/whats-new/dotnet-9/sdk#run-tests-in-parallel).


martincostello:
Looks like the new parallel behaviour in .NET 9 SDK plus the additional 3rd TFM on Windows was what was causing the failures.

PR #2298, opened 2024-09-20
Run Stryker on Linux

martincostello:
Run Stryker on Linux instead of Windows to see if it makes things faster.


martincostello:
Still not working.

PR #2313, opened 2024-09-26
Fix issue 2288

### LINKED ISSUE #2288: Revert to xunit's default parallel algorithm in Polly.Core.Tests
xunit 2.8.0+ changes the default algorithm use to run async tests in parallel: [docs](https://xunit.net/docs/running-tests-in-parallel#algorithms)

This change caused an existing test to fail when dependabot upgraded the project's version of xunit in #2186. This is the specific test that failed:

https://github.com/App-vNext/Polly/blob/957444f8a3ce923b298e9dee1ec653447ec4091d/test/Polly.Core.Tests/Hedging/HedgingStrategyOptionsTests.cs#L22

Having asked over in the xunit repository (see https://github.com/xunit/xunit/discussions/3024), a workaround was made to revert back to the previous default algorithm to fix the test and ingest the dependency update: https://github.com/App-vNext/Polly/pull/2186/commits/35967a314561380bf534d3cdaf2eca1a3570ba3c

While this fixes the CI and allows us to stay up-to-date with xunit, it isn't the optimal solution.

Instead we should determine how to refactor the test to assert on the required intent (that the hedged action executes on a different thread) of the test so that we can revert the changes in the commit referenced above and instead use xunit's default parallel algorithm (option 3 in the linked discussion).

gabidabet:
# Pull Request

Fixes #2288.

I used a semaphore slim to control the number of parallel tasks that can run.

## Confirm the following

### DIFF
--- a/test/Polly.Core.Tests/Hedging/HedgingStrategyOptionsTests.cs
+++ b/test/Polly.Core.Tests/Hedging/HedgingStrategyOptionsTests.cs
@@ -48,7 +48,7 @@ public class HedgingStrategyOptionsTests
         semaphore
             .Wait(TimeSpan.FromSeconds(20))
             .Should()
-            .BeTrue($"The test thread failed to enter the {nameof(semaphore)}, the hedging callback didn't executed");
+            .BeTrue("The test thread failed to complete within the timeout");
         (await task).Result.Should().Be(99);
     }
 
--- a/test/Polly.Core.Tests/Hedging/HedgingStrategyOptionsTests.cs
+++ b/test/Polly.Core.Tests/Hedging/HedgingStrategyOptionsTests.cs
@@ -45,7 +45,10 @@ public class HedgingStrategyOptionsTests
         }))!;
 
         var task = action();
-        semaphore.Wait();
+        semaphore
+            .Wait(TimeSpan.FromSeconds(20))
+            .Should()
+            .BeTrue($"The test thread failed to enter the {nameof(semaphore)}, the hedging callback didn't executed");
         (await task).Result.Should().Be(99);
     }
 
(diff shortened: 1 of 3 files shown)


martincostello:
This seems to resolve the original issue, but I've asked the xunit maintainer who proposed the original workaround to take a look to give their opinion. If we don't hear from them within a week, I'll merge this as-is.

bradwilson:
Yeah, this seems like a reasonable way to ensure that there is no dependency on the current thread.

The only suggestion I'll make is to change `semaphore.Wait();` into one that waits a limited amount of time (long enough to not give false positives, which can be facilitated by putting the test into a test collection that is marked as non-parallelized), and then failing if the wait did not succeed. This will allow the test to wait and fail if somehow the code stop behaving appropriately, rather than waiting forever (and requiring a "long running tests" detection pass to determine which test is problematic).

martincostello:
Good suggestion - thanks Brad!

I reckon about 20 seconds should be long enough for our purposes to guard against a "forever wait".

gabidabet:
Ok I will add a timeout of 20s.

test/Polly.Core.Tests/Hedging/HedgingStrategyOptionsTests.cs
```diff
@@ -39,11 +40,16 @@ public async Task HedgingActionGenerator_EnsureDefaults(bool synchronous)
...
             return Outcome.FromResultAsValueTask(99);
         }))!;
 
-        action.Should().NotBeNull();
-        (await action()).Result.Should().Be(99);
+        var task = action();
+        semaphore
+            .Wait(TimeSpan.FromSeconds(20))
+            .Should()
+            .BeTrue($"The test thread failed to enter the {nameof(semaphore)}, the hedging callback didn't executed");
```
martincostello, line 51:
```suggestion
            .BeTrue("The test thread failed to complete within the timeout");
```

PR #2324, opened 2024-10-02
Add `RetryAfter` to `BrokenCircuitException`.

### LINKED ISSUE #2284: [Feature request]: Indicate time of possible circuit closure in `BrokenCircuitException`

Related to #506.

Imagine a worker service pulling messages from a queue (e.g. Azure Service Bus) and issuing circuit-breaker protected calls to downstream HTTP endpoints. If the downstream services go down and the circuit breaks, the worker has no easy way to know when the circuit breaker would possibly become closed again.

#506 recommends applications to stop processing messages for some time, and I would agree. The problem is "for how long". Knowing when the circuit breaker is likely still open without relying on actually performing and failing the request is important for message queue scenarios, because excessive retries and failures can quickly exhaust message delivery limit and deadletter the messages unnecessarily.

Applications can surely take a guess, but it still needs this piece of information to make an informed decision on how long to cease message processing. It wouldn't help if the circuit breaker is configured to break for 2 days and the application blindly guesses 2 hours.

Indicate the time when the circuit breaker can possibly become closed again in `BrokenCircuitException`. Possible API:

```diff
  public class BrokenCircuitException : ExecutionRejectedException
  {
+     public TimeSpan? RetryAfter { get; }
  }
```

This matches `RetryAfter` from `RateLimiterRejectedException`. Pulling it up to `ExecutionRejectedException` is not advisable because it does not apply to all derived exception types.

For `IsolatedCircuitException`, set this to `null`.

### Additional context

_No response_

DL444:
# Pull Request

Closes #2284

This PR adds `RetryAfter` property to `BrokenCircuitException` to indicate time of possible circuit closure.

The implementation stores the time span from the time when the exception is thrown to the time when the circuit breaker should enter half-open state. This can cause inaccuracy if the exception is stowed and consumed at a later time, as the value would become stale. This route is chosen over storing the time of circuit closure because it

- Maintains API consistency with `RateLimiterRejectedException`, another derived type of `ExecutionRejectedException`, whose `RetryAfter` was already defined to return `TimeSpan`.
- Avoids performing time computation in the exception.
- Avoids demonstrating inconsistent behavior when the value is observed multiple times.

Tests are expanded to cover the new APIs. Specific test policy proposed and implemented:

- Direct tests against the exception types (`BrokenCircuitExceptionTests`, `IsolatedCircuitExceptionTests`) are expanded to cover the new API;
- Issue tests are kept as-is because they are supposed to focus on the specific relevant issues.
- Each specs test category (`CircuitBreakerSpecs`, `CircuitBreakerAsyncSpecs`, `AdvancedCircuitBreakerSpecs`, `AdvancedCircuitBreakerAsyncSpecs`, `CircuitBreakerTResultSpecs`, `CircuitBreakerTResultMixedResultExceptionSpecs`, `CircuitBreakerTResultAsyncSpecs`) implements one test for the value of the new `RetryAfter` property.
- All other reference sites of the exception in specs tests confirm nullity of the property.

Let me know if either the implementation or the tests need rework.

It is also observed that `BrokenCircuitException` supports binary serialization on .NET Framework. I'm unfortunately not familiar with its details, and as a deprecated technology, information about it online is scarce. I'm not sure if this change would have any negative implications from a compatibility standpoint.

## Confirm the following


martintmk:
I would keep this feature to `Polly.Core` only as the `Polly` V7 should be limited to bug fixes only. Wdyt @martincostello, @peter-csala ?

peter-csala:
> I would keep this feature to `Polly.Core` only as the `Polly` V7 should be limited to bug fixes only. Wdyt @martincostello, @peter-csala ?

Yes, that's my understand as well. Your implementation should be limited to V8 only.

peter-csala:
Also, please don't forget to update the documentation as well: https://github.com/App-vNext/Polly/blob/main/docs/strategies/circuit-breaker.md

src/Polly.Core/CircuitBreaker/BrokenCircuitException.cs
```diff
@@ -22,6 +22,14 @@ public BrokenCircuitException()
     {
     }
 
+    /// <summary>
+    /// Initializes a new instance of the <see cref="BrokenCircuitException"/> class.
+    /// </summary>
+    /// <param name="retryAfter">The time during which the circuit will be broken.</param>
```
martincostello, line 28: While true, I think it's describing a related concept rather than the literal meaning of the value. Something like the below might be better:

```suggestion
    /// <param name="retryAfter">The period after which the circuit will close.</param>
```

Whatever the change is to, it should be applied to all the constructors not just this one.

src/Polly.Core/CircuitBreaker/BrokenCircuitException.cs
```diff
@@ -54,4 +79,13 @@ protected BrokenCircuitException(SerializationInfo info, StreamingContext contex
...
 #pragma warning restore RS0016 // Add public types and members to the declared API
+
+    /// <summary>
+    /// Gets the amount of time before circuit can become closed.
+    /// </summary>
+    /// <remarks>
+    /// This is set when the exception object was constructed and might be inaccurate if consumed at a later time.
+    /// Can be <see langword="null"/> if not provided or when the circuit was manually isolated.
+    /// </remarks>
+    public TimeSpan? RetryAfter { get; }
```
martincostello, line 90:
```suggestion
    /// <summary>
    /// Gets the amount of time before the circuit can become closed, if known.
    /// </summary>
    /// <remarks>
    /// This value is specified when the instance is constructed and may be inaccurate if consumed at a later time.
    /// Can be <see langword="null"/> if not provided or if the circuit was manually isolated.
    /// </remarks>
    public TimeSpan? RetryAfter { get; }
```

src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
```diff
@@ -307,8 +307,8 @@ private void SetLastHandledOutcome_NeedsLock(Outcome<T> outcome)
 
     private BrokenCircuitException CreateBrokenCircuitException() => _breakingException switch
     {
-        Exception exception => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, exception),
-        _ => new BrokenCircuitException(BrokenCircuitException.DefaultMessage)
+        Exception exception => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, exception, _blockedUntil - _timeProvider.GetUtcNow()),
+        _ => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, _blockedUntil - _timeProvider.GetUtcNow())
     };
```
martincostello, line 312: Rather than do the same thing in both branches, let's compute it once:

```
{
    var retryAfter = _blockedUntil - _timeProvider.GetUtcNow();
    return _breakingException switch
    {
        Exception exception => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, exception, retryAfter),
        _ => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter),
    };
}
```

test/Polly.Core.Tests/CircuitBreaker/IsolatedCircuitExceptionTests.cs
```diff
@@ -7,10 +7,16 @@ public class IsolatedCircuitExceptionTests
...
-        new IsolatedCircuitException().Message.Should().Be("The circuit is manually held open and is not allowing calls.");
-        new IsolatedCircuitException("Dummy.", new InvalidOperationException()).Message.Should().Be("Dummy.");
-        new IsolatedCircuitException("Dummy.", new InvalidOperationException()).InnerException.Should().BeOfType<InvalidOperationException>();
+        IsolatedCircuitException dummyMessageException = new("Dummy.");
+        dummyMessageException.Message.Should().Be("Dummy.");
+        dummyMessageException.RetryAfter.Should().BeNull();
+        IsolatedCircuitException defaultException = new();
+        defaultException.Message.Should().Be("The circuit is manually held open and is not allowing calls.");
+        defaultException.RetryAfter.Should().BeNull();
+        IsolatedCircuitException dummyMessageExceptionWithInnerException = new("Dummy.", new InvalidOperationException());
```
martincostello, line 16: A question of style, but I prefer the `var foo = new Bar()` style over `Bar foo = new()` for variables.

src/Polly.Core/CircuitBreaker/BrokenCircuitException.cs
```diff
@@ -41,6 +57,15 @@ public BrokenCircuitException(string message, Exception inner)
     {
     }
 
+    /// <summary>
+    /// Initializes a new instance of the <see cref="BrokenCircuitException"/> class.
+    /// </summary>
+    /// <param name="message">The message that describes the error.</param>
+    /// <param name="inner">The inner exception.</param>
+    /// <param name="retryAfter">The time during which the circuit will be broken.</param>
+    public BrokenCircuitException(string message, Exception inner, TimeSpan retryAfter)
```
martintmk, line 66:
```suggestion
    public BrokenCircuitException(string message, TimeSpan retryAfter, Exception inner)
```

keep the inner exception as last parameter

DL444:
The source branch is rebased, revised, and force-pushed to incorporate the following requested changes:

- Reverts all changes made to `Polly` and `Polly.Specs` projects as new features are no longer accepted for them.
- Expands relevant documentations to cover the new feature.
- Addresses miscellaneous code review comments.

src/Polly.Core/CircuitBreaker/BrokenCircuitException.cs
```diff
@@ -54,4 +79,13 @@ protected BrokenCircuitException(SerializationInfo info, StreamingContext contex
...
 #pragma warning restore RS0016 // Add public types and members to the declared API
+
+    /// <summary>
+    /// Gets the amount of time before the circuit can become closed, if known.
+    /// </summary>
+    /// <remarks>
+    /// This value is specified when the instance is constructed and may be inaccurate if consumed at a later time.
+    /// Can be <see langword="null"/> if not provided or if the circuit was manually isolated.
+    /// </remarks>
+    public TimeSpan? RetryAfter { get; }
```
martincostello, line 90: We should add a test like this one to verify that binary serialization works correctly for .NET Framework usage:

https://github.com/App-vNext/Polly/blob/439cf1515d84c9ecaa1c6379def2cb3c47151df7/test/Polly.Core.Tests/Timeout/TimeoutRejectedExceptionTests.cs#L21-L31

Getting it to pass probably needs similar code to the below adding to ensure the property value is serialized correctly:

https://github.com/App-vNext/Polly/blob/439cf1515d84c9ecaa1c6379def2cb3c47151df7/src/Polly.Core/Timeout/TimeoutRejectedException.cs#L78-L89

test/Polly.Core.Tests/CircuitBreaker/BrokenCircuitExceptionTests.cs
```diff
@@ -7,10 +7,27 @@ public class BrokenCircuitExceptionTests
...
+        dummyMessageWithRetryAfterException.Message.Should().Be("Dummy.");
+        dummyMessageWithRetryAfterException.RetryAfter.Should().Be(retryAfter);
+        var dummyMessageExceptionWithInnerException = new BrokenCircuitException("Dummy.", new InvalidOperationException());
+        dummyMessageExceptionWithInnerException.Message.Should().Be("Dummy.");
+        dummyMessageExceptionWithInnerException.InnerException.Should().BeOfType<InvalidOperationException>();
+        dummyMessageExceptionWithInnerException.RetryAfter.Should().BeNull();
+        var dummyMessageExceptionWithInnerExceptionAndRetryAfter = new BrokenCircuitException("Dummy.", retryAfter, new InvalidOperationException());
+        dummyMessageExceptionWithInnerExceptionAndRetryAfter.Message.Should().Be("Dummy.");
+        dummyMessageExceptionWithInnerExceptionAndRetryAfter.InnerException.Should().BeOfType<InvalidOperationException>();
+        dummyMessageExceptionWithInnerExceptionAndRetryAfter.RetryAfter.Should().Be(retryAfter);
```
martincostello, line 30: For readability:

```suggestion
        var retryAfter = new TimeSpan(1, 0, 0);

        var defaultException = new BrokenCircuitException();
        defaultException.Message.Should().Be("The circuit is now open and is not allowing calls.");
        defaultException.RetryAfter.Should().BeNull();

        var retryAfterException = new BrokenCircuitException(retryAfter);
        retryAfterException.Message.Should().Be($"The circuit is now open and is not allowing calls. It can be retried after '{retryAfter}'.");
        retryAfterException.RetryAfter.Should().Be(retryAfter);

        var dummyMessageException = new BrokenCircuitException("Dummy.");
        dummyMessageException.Message.Should().Be("Dummy.");
        dummyMessageException.RetryAfter.Should().BeNull();

        var dummyMessageWithRetryAfterException = new BrokenCircuitException("Dummy.", retryAfter);
        dummyMessageWithRetryAfterException.Message.Should().Be("Dummy.");
        dummyMessageWithRetryAfterException.RetryAfter.Should().Be(retryAfter);

        var dummyMessageExceptionWithInnerException = new BrokenCircuitException("Dummy.", new InvalidOperationException());
        dummyMessageExceptionWithInnerException.Message.Should().Be("Dummy.");
        dummyMessageExceptionWithInnerException.InnerException.Should().BeOfType<InvalidOperationException>();
        dummyMessageExceptionWithInnerException.RetryAfter.Should().BeNull();

        var dummyMessageExceptionWithInnerExceptionAndRetryAfter = new BrokenCircuitException("Dummy.", retryAfter, new InvalidOperationException());
        dummyMessageExceptionWithInnerExceptionAndRetryAfter.Message.Should().Be("Dummy.");
        dummyMessageExceptionWithInnerExceptionAndRetryAfter.InnerException.Should().BeOfType<InvalidOperationException>();
        dummyMessageExceptionWithInnerExceptionAndRetryAfter.RetryAfter.Should().Be(retryAfter);
```

test/Polly.Core.Tests/CircuitBreaker/IsolatedCircuitExceptionTests.cs
```diff
@@ -7,10 +7,16 @@ public class IsolatedCircuitExceptionTests
...
+        var dummyMessageException = new IsolatedCircuitException("Dummy.");
+        dummyMessageException.Message.Should().Be("Dummy.");
+        dummyMessageException.RetryAfter.Should().BeNull();
+        var defaultException = new IsolatedCircuitException();
+        defaultException.Message.Should().Be("The circuit is manually held open and is not allowing calls.");
+        defaultException.RetryAfter.Should().BeNull();
+        var dummyMessageExceptionWithInnerException = new IsolatedCircuitException("Dummy.", new InvalidOperationException());
+        dummyMessageExceptionWithInnerException.Message.Should().Be("Dummy.");
+        dummyMessageExceptionWithInnerException.InnerException.Should().BeOfType<InvalidOperationException>();
+        dummyMessageExceptionWithInnerException.RetryAfter.Should().BeNull();
```
martincostello, line 19:
```suggestion
        var dummyMessageException = new IsolatedCircuitException("Dummy.");
        dummyMessageException.Message.Should().Be("Dummy.");
        dummyMessageException.RetryAfter.Should().BeNull();

        var defaultException = new IsolatedCircuitException();
        defaultException.Message.Should().Be("The circuit is manually held open and is not allowing calls.");
        defaultException.RetryAfter.Should().BeNull();

        var dummyMessageExceptionWithInnerException = new IsolatedCircuitException("Dummy.", new InvalidOperationException());
        dummyMessageExceptionWithInnerException.Message.Should().Be("Dummy.");
        dummyMessageExceptionWithInnerException.InnerException.Should().BeOfType<InvalidOperationException>();
        dummyMessageExceptionWithInnerException.RetryAfter.Should().BeNull();
```

test/Polly.Core.Tests/CircuitBreaker/BrokenCircuitExceptionTests.cs
peter-csala: Or you could create parameterized test instead. Something like this
```cs
[Theory]
[InlineData(null, false, "The circuit is now open and is not allowing calls.")]
[InlineData(null, true, "The circuit is now open and is not allowing calls. It can be retried after.")]
[InlineData("Dummy.", false, "Dummy.")]
[InlineData("Dummy.", true, "Dummy.")]
public void bce_without_inner_exception(string? errorMessage, bool useRetryAfter, string expectedMessage)
{
    //Arrange
    TimeSpan? retryAfter = useRetryAfter ? new(1, 0, 0) : null;
    List<object> args = new();
    if (errorMessage is not null) { args.Add(errorMessage); }
    if (useRetryAfter) { args.Add(retryAfter); }
    
    //Act
    var exception = (BrokenCircuitException)Activator.CreateInstance(typeof(BrokenCircuitException), args.ToArray());
    
    //Assert
    exception.Message.Should().Be(expectedMessage);
    exception.RetryAfter.Should().Be(retryAfter);
}
```
The same could be applied for those cases where you pass an inner `Exception`

martincostello: IMHO there's too much conditional logic in that example.

Either it's one test broken up, or multiple tests (one for each constructor).

DL444: I may opt for separating the tests across multiple methods to make things more manageable.

DL444:
New commits are added to:
- Implement binary serialization support for the new property.
- Refactor exception constructor tests to improve readability.

src/Polly.Core/CircuitBreaker/BrokenCircuitException.cs
```diff
@@ -76,6 +76,31 @@ public BrokenCircuitException(string message, TimeSpan retryAfter, Exception inn
...
+
+        // https://github.com/dotnet/runtime/issues/42460
+        SerializationInfoEnumerator enumerator = info.GetEnumerator();
+        while (enumerator.MoveNext())
+        {
+            SerializationEntry entry = enumerator.Current;
+            if (string.Equals(entry.Name, "RetryAfter", StringComparison.Ordinal))
+            {
+                var ticks = (long)entry.Value;
+                RetryAfter = new TimeSpan(ticks);
```
martincostello, line 89: Can break here as we've found the only member we're looking for, though it does make me wonder why we're doing this approach at all rather than just always setting `RetryAfter` into the `SerializationInfo` and then read it out with `GetValue()`? Do we need to make this ultra-definsive about people passing in a `SerializationInfo` that wasn't populated by us?

test/Polly.Core.Tests/CircuitBreaker/BrokenCircuitExceptionTests.cs
```diff
@@ -5,34 +5,79 @@ namespace Polly.Core.Tests.CircuitBreaker;
...
+        BrokenCircuitException roundtripResult = BinarySerializationUtil.SerializeAndDeserializeException(exception);
+        roundtripResult.Should().NotBeNull();
+        roundtripResult.Message.Should().Be(TestMessage);
+        roundtripResult.InnerException.Should().BeOfType<InvalidOperationException>();
+        roundtripResult.RetryAfter.Should().BeNull();
+    }
 #endif
+
+    private const string TestMessage = "Dummy.";
+    private static readonly TimeSpan TestRetryAfter = new(1, 0, 0);
```
martincostello, line 82: More readable IMHO:

```suggestion
    private static readonly TimeSpan TestRetryAfter = TimeSpan.FromHours(1);
```

src/Polly.Core/CircuitBreaker/BrokenCircuitException.cs
DL444: Because the property is nullable and the current implementation does not add the value when it's null. Also I think we'd better defend against users who deserialize from persisted data generated by previous versions. 

As to why not `break`ing, I think it avoids future problems where we add more properties later without revisiting and removing the `break`. But that's opinion.

martincostello: Ok, in that case let's just add the break. The need to remove it will become apparent in testing if we ever add any additional properties.

martincostello (approved):

PR #2331, opened 2024-10-07
Seperate mutation tests

### LINKED ISSUE #1755: [Feature request]: Speed-up the build by executing the mutation tests in a separate workflow

I think we can speed-up the build by extracting the mutation steps from https://github.com/App-vNext/Polly/blob/main/.github/workflows/build.yml into separate workflows.

I believe we can introduce two new workflows:

- `mutation-tests-core.yml` - in this workflow we would test all new Polly v8 projects.
- `mutation-tests-legacy.yml` - in this workflow we would test only Polly project.

This way the mutation tests is decreased by half on top of of reduced build times. With this change our PR builds could be done in less than 20 minutes.

### Additional context

_No response_

ShawnWu33:
# Pull Request

Resolves #1755 

- Create two new workflow to cover mutation test core and mutation test legacy.
- Modify cake scirpt to accomodate another two target "MutationCore" and "MutationLegacy"
- Remove mutation test in build workflow

Tested these three workflow in my repo, below is what they lookslike during execution:
Build workflow: https://github.com/ShawnWu33/Polly/actions/runs/11216605547
Mutation core workflow: https://github.com/ShawnWu33/Polly/actions/runs/11216605554
Mutation legacy workflow: https://github.com/ShawnWu33/Polly/actions/runs/11216605545

## Confirm the following


.github/workflows/mutation-tests-core.yml
```diff
@@ -0,0 +1,63 @@
+name: mutation-tests-core
```
martincostello, line 1: Rather than two workflows with one job each, we should have one workflow with two jobs. Then there's less duplicative code in the workflow files.

In fact, using a matrix (like the build does for the operating system) would only need one job to be defined, which would then fan out for the two sets of mutation tests.

ShawnWu33: Yes, I definitely agree that putting them in one workflow can reduce some duplication in workflow.

I put them in this way was because in @martintmk issue description mentioned there are two workflows.

I thought initially maybe we want to put just core as required and legacy as optional in the future when we don't support legacy.

If we can confirm that we need just one workflow I can make the change.

martincostello: Just one workflow with two jobs which specify which set to run via a matrix please.

.github/workflows/mutation-tests.yml
```diff
@@ -0,0 +1,76 @@
+name: mutation-tests
+
+on:
+  push:
+    branches: [ main, release/* ]
+    tags: [ '*' ]
```
martincostello, line 6: There's no need to run these for tags - they were skipped previously as it massively slowed down the time to publish a new release.

.github/workflows/mutation-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+  DOTNET_SYSTEM_CONSOLE_ALLOW_ANSI_COLOR_REDIRECTION: 1
+  NUGET_XMLDOC_MODE: skip
+  TERM: xterm
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: Run Mutation Test ${{ matrix.target }}
```
martincostello, line 25:
```suggestion
  mutations:
    name: ${{ matrix.name }}
```

.github/workflows/build.yml
```diff
@@ -26,10 +26,6 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60
```
martincostello, line 27: With the mutation tests moved out, we can probably reduce this to 20 minutes.

.github/workflows/mutation-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+      id: setup-dotnet
+
+    - name: Setup NuGet cache
+      uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+      with:
+        path: ~/.nuget/packages
+        key: ${{ runner.os }}-nuget-${{ hashFiles('**/*.csproj', '**/*.props') }}
+        restore-keys: ${{ runner.os }}-nuget-
+
+    - name: Build and Run Mutation Tests
```
martincostello, line 62:
```suggestion
    - name: Run mutation tests
```

.github/workflows/mutation-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+      id: build
+      shell: pwsh
+      run: ./build.ps1 -Target Mutation${{ matrix.target }}
+
+    - name: downcase target
+      run: |
+        echo "REPO=${GITHUB_REPOSITORY,,}" >>${GITHUB_ENV}
+
```
martincostello, line 70: This whole step appears to be redundant.

.github/workflows/mutation-tests-core.yml
ShawnWu33: Updated.

.github/workflows/mutation-tests.yml
ShawnWu33: Sure. Updated.

.github/workflows/mutation-tests.yml
ShawnWu33: Thanks. Updated.

.github/workflows/build.yml
ShawnWu33: Thanks. Updated.

.github/workflows/mutation-tests.yml
ShawnWu33: Thanks. Updated.

.github/workflows/mutation-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+    - name: Setup NuGet cache
+      uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+      with:
+        path: ~/.nuget/packages
+        key: ${{ runner.os }}-nuget-${{ hashFiles('**/*.csproj', '**/*.props') }}
+        restore-keys: ${{ runner.os }}-nuget-
+
+    - name: Build and Run Mutation Tests
+      id: build
```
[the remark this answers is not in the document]
ShawnWu33: Thanks. Updated.

.github/workflows/mutation-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+
+    - name: Setup .NET SDKs
+      uses: actions/setup-dotnet@6bd8b7f7774af54e05809fcc5431931b3eb1ddee # v4.0.1
+      with:
+        dotnet-version: |
+          6.0.x
+
+    - name: Setup .NET SDK
+      uses: actions/setup-dotnet@6bd8b7f7774af54e05809fcc5431931b3eb1ddee # v4.0.1
+      id: setup-dotnet
```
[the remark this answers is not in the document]
ShawnWu33: Thanks. Updated.

.github/workflows/mutation-tests.yml
```diff
@@ -0,0 +1,69 @@
...
+    - name: Setup NuGet cache
+      uses: actions/cache@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+      with:
+        path: ~/.nuget/packages
+        key: ${{ runner.os }}-nuget-${{ hashFiles('**/*.csproj', '**/*.props') }}
+        restore-keys: ${{ runner.os }}-nuget-
+
+    - name: Run mutation tests
+      shell: pwsh
+      run: ./build.ps1 -Target Mutation${{ matrix.target }}
```
martincostello, line 62:
```suggestion
      shell: pwsh
      env:
        MUTATION_TARGET: 'Mutation${{ matrix.target }}'
      run: ./build.ps1 -Target ${env:MUTATION_TARGET}
```

build.cake
```diff
@@ -159,56 +159,29 @@ Task("__RunTests")
...
-        var strykerPath = Context.Tools.Resolve("Stryker.CLI.dll");
-        var mutationScore = XmlPeek(proj, "/Project/PropertyGroup/MutationScore/text()", new XmlPeekSettings { SuppressWarning = true });
-        var score = int.Parse(mutationScore);
-
-        Information($"Running mutation tests for '{proj}'. Test Project: '{testProj}'");
+Task("__RunLegacyMutationTests")
+    .Does((context) =>
+{
+    var oldDirectory = context.Environment.WorkingDirectory;
+    context.Environment.WorkingDirectory = MakeAbsolute(Directory("test"));
```
martincostello, line 180: Why do we need to change the working directory when we didn't before?

ShawnWu33: Previously we surrounding the whole test block with this.
https://github.com/App-vNext/Polly/blob/a50d47230c4b6587f8257f91f47cf491ef39a019/build.cake#L176-L185

Just because now we seperate the test in two place so this shown as new code.

martincostello: I still don't quite follow - dotnet test should be being invoked from the repo root and the working directory _shouldn't_ matter because we specify `--project` as a full path?

ShawnWu33: Yes you are correct. I got that directory changes removed.

.github/workflows/mutation-tests.yml
```diff
@@ -0,0 +1,71 @@
...
+  DOTNET_SYSTEM_CONSOLE_ALLOW_ANSI_COLOR_REDIRECTION: 1
+  NUGET_XMLDOC_MODE: skip
+  TERM: xterm
+
+permissions:
+  contents: read
+
+jobs:
+  mutations:
+    name: ${{ matrix.name }}
```
martincostello, line 24: Just to make the status check names for our protection rules a bit more specific:

```suggestion
    name: 'mutations-${{ matrix.name }}'
```

martincostello (approved):

martincostello:
Just one final edit please, then this should be good to go. I'll just need to get the required statuses updated.

martincostello:
@joelhulen Could you edit the branch protections to make the following required please?

- `mutations-core`
- `mutations-legacy`
- `lint`
- `dependency-review`
- `docs`
- `validate-packages`

martincostello:
Thanks again for this @ShawnWu33 - this shaves a good 10-15 minutes off the total time between opening a PR and being able to merge it now.

Issue #2345, opened 2024-10-14
[Feature request]: Determine policy which caused rejection


lonix1:
### Is your feature request related to a specific problem? Or an existing feature?

I've spent some time learning how Polly works, and specifically how to combine policies (which is probably a common need in a non-trivial production system).

I've found a pain point in how Polly handles rejections (excuse the pun).

Suppose one combines rate limiting and concurrency control, without queuing (or after reaching the queue limit), and there is a rejection:
- if rate limited: read delay from `RateLimiterRejectedException.RetryAfter` then retry
- if concurrency limited: return to caller with error

To do that one needs to know the source of the rejection.

### Describe the solution you'd like

There are two workarounds, both bad:
- If the `RateLimiterRejectedException.RetryAfter` is null, one can infer that it was concurrency limited (rather than fixed window, sliding window or token bucket). But this is a dangerous assumption as the library could add more policies in the future, or change it's internals.
- One could use a variation of the roundabout approach [shown here](https://github.com/App-vNext/Polly/discussions/1777#discussioncomment-7506465) (though I'm not even certain it works... I couldn't get it to work). Assuming it works, it's very messy and overly complicated. I wouldn't want to maintain such code if I only ever touch Polly code once or twice a year.

The ideal solution is mentioned in that thread, which is to add a new property to the exception, which identifies the source of the rejection.

Perhaps `RateLimiterRejectedException.Source` could be a string, equal to a configurable `Name` property for the policy, or if unset then equal to `policy.GetType().Name`.

### Additional context

Thank you for considering it!

peter-csala:
Just to clarify: which Polly version are we talking about? The `RateLimiterRejectedException` was introduced in the V8 API. V7 API uses _policies_ whereas the V8 uses _strategies_. I educated guess is that you are using the V8 API, can you please confirm it?

peter-csala:
Could you please describe the desired behavior in a bit more detail (how many strategies are chain, in what order, how are they configured, etc.)? Or you could also share an ideal pipeline setup.

peter-csala:
Well, as far as I know there is only a single place where Polly throws `RateLimiterRejectedException`:

https://github.com/App-vNext/Polly/blob/d36058d0ac55ac221c1f0d9405aff932998c045a/src/Polly.RateLimiting/RateLimiterResilienceStrategy.cs#L68

I think we could extend the RLRE with a `Source` property and populate it similarly like the Telemetry event's `Source`:
```none
[...]

PR #2346, opened 2024-10-15
Add TelemetrySource to ExecutionRejectedException

peter-csala:
# Pull Request

#2345 

### Done
- Added `TelemetrySource` property to the `ExecutionRejectedException`
- Extended `ResilienceStrategyTelemetry` with a `UpdateTelemetrySource` method
- Updated several strategies to populate `TelemetrySource` on the thrown exception

## Confirm the following

### DIFF
--- a/src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
+++ b/src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
@@ -96,7 +96,7 @@ internal sealed class CircuitStateController<T> : IDisposable
         lock (_lock)
         {
             var exception = new IsolatedCircuitException();
-            _telemetry.UpdateTelemetrySource(exception);
+            _telemetry.SetTelemetrySource(exception);
             SetLastHandledOutcome_NeedsLock(Outcome.FromException<T>(exception));
             OpenCircuitFor_NeedsLock(Outcome.FromResult<T>(default), TimeSpan.MaxValue, manual: true, context, out task);
             _circuitState = CircuitState.Isolated;
@@ -159,7 +159,7 @@ internal sealed class CircuitStateController<T> : IDisposable
 
         if (exception is not null)
         {
-            _telemetry.UpdateTelemetrySource(exception);
+            _telemetry.SetTelemetrySource(exception);
             return Outcome.FromException<T>(exception);
         }
 
@@ -311,12 +311,12 @@ internal sealed class CircuitStateController<T> : IDisposable
     private BrokenCircuitException CreateBrokenCircuitException()
     {
         TimeSpan retryAfter = _blockedUntil - _timeProvider.GetUtcNow();
-        BrokenCircuitException exception = _breakingException switch
+        var exception = _breakingException switch
         {
             Exception ex => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter, ex),
             _ => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter)
         };
-        _telemetry.UpdateTelemetrySource(exception);
+        _telemetry.SetTelemetrySource(exception);
         return exception;
     }
 
--- a/src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
+++ b/src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
@@ -95,7 +95,9 @@ internal sealed class CircuitStateController<T> : IDisposable
 
         lock (_lock)
         {
-            SetLastHandledOutcome_NeedsLock(Outcome.FromException<T>(new IsolatedCircuitException()));
+            var exception = new IsolatedCircuitException();
+            _telemetry.UpdateTelemetrySource(exception);
+            SetLastHandledOutcome_NeedsLock(Outcome.FromException<T>(exception));
             OpenCircuitFor_NeedsLock(Outcome.FromResult<T>(default), TimeSpan.MaxValue, manual: true, context, out task);
             _circuitState = CircuitState.Isolated;
         }
@@ -123,7 +125,7 @@ internal sealed class CircuitStateController<T> : IDisposable
     {
         EnsureNotDisposed();
 
-        Exception? exception = null;
+        BrokenCircuitException? exception = null;
         bool isHalfOpen = false;
 
         Task? task = null;
@@ -157,6 +159,7 @@ internal sealed class CircuitStateController<T> : IDisposable
 
         if (exception is not null)
         {
+            _telemetry.UpdateTelemetrySource(exception);
             return Outcome.FromException<T>(exception);
         }
 
@@ -308,11 +311,13 @@ internal sealed class CircuitStateController<T> : IDisposable
     private BrokenCircuitException CreateBrokenCircuitException()
     {
         TimeSpan retryAfter = _blockedUntil - _timeProvider.GetUtcNow();
-        return _breakingException switch
+        BrokenCircuitException exception = _breakingException switch
         {
-            Exception exception => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter, exception),
+            Exception ex => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter, ex),
             _ => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter)
         };
+        _telemetry.UpdateTelemetrySource(exception);
+        return exception;
     }
 
     private void OpenCircuit_NeedsLock(Outcome<T> outcome, bool manual, ResilienceContext context, out Task? scheduledTask)
--- a/src/Polly.Core/ExecutionRejectedException.cs
+++ b/src/Polly.Core/ExecutionRejectedException.cs
@@ -2,6 +2,8 @@
 using System.Runtime.Serialization;
 #endif
 
+using Polly.Telemetry;
+
 namespace Polly;
 
 /// <summary>
@@ -49,4 +51,10 @@ public abstract class ExecutionRejectedException : Exception
     }
 #endif
 #pragma warning restore RS0016 // Add public types and members to the declared API
+
+    /// <summary>
+    /// Gets the source of the strategy which has thrown the exception, if known.
+    /// </summary>
+    public virtual ResilienceTelemetrySource? TelemetrySource { get; internal set; }
+
 }
--- a/src/Polly.Core/ExecutionRejectedException.cs
+++ b/src/Polly.Core/ExecutionRejectedException.cs
@@ -56,5 +56,4 @@ public abstract class ExecutionRejectedException : Exception
     /// Gets the source of the strategy which has thrown the exception, if known.
     /// </summary>
     public virtual ResilienceTelemetrySource? TelemetrySource { get; internal set; }
-
 }
--- a/src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
+++ b/src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
@@ -26,7 +26,7 @@ public sealed class ResilienceStrategyTelemetry
     /// <summary>
     ///  Sets the source of the telemetry on the provided exception.
     /// </summary>
-    /// <param name="exception">The to-be-updated exception.</param>
+    /// <param name="exception">The to-be-set exception.</param>
     [EditorBrowsable(EditorBrowsableState.Never)]
     public void SetTelemetrySource(ExecutionRejectedException exception)
     {
--- a/src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
+++ b/src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
@@ -24,11 +24,11 @@ public sealed class ResilienceStrategyTelemetry
     internal ResilienceTelemetrySource TelemetrySource { get; }
 
     /// <summary>
-    ///  Updates the source of the telemetry on the provided exception.
+    ///  Sets the source of the telemetry on the provided exception.
     /// </summary>
     /// <param name="exception">The to-be-updated exception.</param>
     [EditorBrowsable(EditorBrowsableState.Never)]
-    public void UpdateTelemetrySource(ExecutionRejectedException exception)
+    public void SetTelemetrySource(ExecutionRejectedException exception)
     {
         Guard.NotNull(exception);
 
--- a/src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
+++ b/src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
@@ -1,3 +1,5 @@
+using System.ComponentModel;
+
 namespace Polly.Telemetry;
 
 /// <summary>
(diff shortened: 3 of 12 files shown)


src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
```diff
@@ -19,7 +19,11 @@ internal ResilienceStrategyTelemetry(ResilienceTelemetrySource source, Telemetry
 
     internal TelemetryListener? Listener { get; }
 
-    internal ResilienceTelemetrySource TelemetrySource { get; }
+    /// <summary>
+    /// Gets BlahBlahBlah.
+    /// </summary>
+    /// <value>dummy.</value>
+    public ResilienceTelemetrySource TelemetrySource { get; } // Fix visibility
```
peter-csala, line 26: If I add the following line to the `Polly.Core.csproj`:
```xml
<InternalsVisibleToProject Include="Polly.RateLimiting" />
```
then I don't have to change the access modifier from `internal` to `public`. 

But it causes lots of compile time error, like complaining about `RateLimiterResilienceStrategy`'s `ExecuteCore` should be `proctected internal` not just `proctected`, etc..

Do you have any other idea than making this property `public`?

src/Polly.RateLimiting/RateLimiterRejectedException.cs
```diff
@@ -21,6 +23,19 @@ public RateLimiterRejectedException()
     {
     }
 
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RateLimiterRejectedException"/> class.
+    /// </summary>
+    /// <param name="telemetrySource">The source pipeline and strategy names.</param>
+    public RateLimiterRejectedException(ResilienceTelemetrySource telemetrySource)
```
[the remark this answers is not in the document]
martincostello: Code analysers recommend a standard set of constructors for publicly-visible types:

- `.ctor()`
- `.ctor(string)`
- `.ctor(string, Exception)`
- `.ctor(SerializationInfo, StreamingContext)`

Because we can't predict what our users might be doing in private code we can't see, we should add at a minimum what someone would need to also populate any new property themselves if they wanted to.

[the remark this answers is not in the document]
peter-csala: Based on this I would say, it is enough for now to introduce those constructors that we need. Future requests asking for other overloads can be easily added. Can we agree that we extend this exception class only with these two overloads?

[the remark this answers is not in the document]
martincostello: Only those two means you can't create an instance of the exception with all the possible property values, so we would need what you need to make this feature work _plus_ enough extras to allow for a fully constructed object (which could be just one).

src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
martincostello: As far as I can see, we're currently doing this:

1. Exposing the source;
2. Using the source to pass to an exception;
3. Having the exception turn the context into a string;
4. Expose the string on the exception.

If that is the case, would this be better?

1. Add a method to the context to turn it into a string (i.e. have it own its own logic to do that so it's not far from the type and could misalign if we changed the context in the future and forgot we were stringifying it elsewhere);
2. Use that method to expose that string representation on the telemetry;
3. Pass that string to the exception;
4. Expose that string on the exception.

The above avoids the visibility problem, and makes the source string come from Polly.Core as the source-of-truth, rather than making something visible solely for rate limiting to make its own decisions about.

peter-csala: With this approach we end up with following error:
> Type 'RateLimiterRejectedException' already defines a member called 'RateLimiterRejectedException' with the same parameter types

```
public RateLimiterRejectedException(string telemetrySource)
```

martincostello: So let's add a constructor that doesn't cause that problem then 😄 - for example, by manually specifying an exception message `new RateLimiterRejectedException("The source {blah} go boom", source)`.

I think it would be much neater to have to manually construct a message because we've already used `.ctor(string)` for an explicit exception than have to expose the source just to avoid that.

src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
```diff
@@ -21,6 +21,22 @@ internal ResilienceStrategyTelemetry(ResilienceTelemetrySource source, Telemetry
 
     internal ResilienceTelemetrySource TelemetrySource { get; }
 
+ #pragma warning disable S4049
```
martincostello, line 24: Rather than suppress, I suggest words such as `As`, `To`, `Create` or `Convert` instead of `Get`.

src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
```diff
@@ -21,6 +21,22 @@ internal ResilienceStrategyTelemetry(ResilienceTelemetrySource source, Telemetry
...
+ #pragma warning disable CA1024
+    /// <summary>
+    /// Returns the concatenated telemetry source.
+    /// </summary>
+    /// <returns>The concatenated telemetry source.</returns>
```
martincostello, line 29: I would say "concatenated" is an implementation detail - instead I'd say something like "Returns a string representation of the source of the telemetry".

src/Polly.RateLimiting/RateLimiterRejectedException.cs
```diff
@@ -84,10 +131,16 @@ public RateLimiterRejectedException(string message, TimeSpan retryAfter, Excepti
     private RateLimiterRejectedException(SerializationInfo info, StreamingContext context)
         : base(info, context)
     {
-        var value = info.GetDouble("RetryAfter");
-        if (value >= 0.0)
+        var retryAfter = info.GetDouble(nameof(RetryAfter));
+        if (retryAfter >= 0.0)
```
martincostello, line 135: Seeing this again, I wonder if this should be `default` or `0d`?

src/Polly.RateLimiting/RateLimiterResilienceStrategy.cs
```diff
@@ -65,7 +65,8 @@ protected override async ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState
             await OnLeaseRejected(new OnRateLimiterRejectedArguments(context, lease)).ConfigureAwait(context.ContinueOnCapturedContext);
         }
 
-        var exception = retryAfter.HasValue ? new RateLimiterRejectedException(retryAfter.Value) : new RateLimiterRejectedException();
+        var source = _telemetry.GetTelemetrySource();
+        var exception = retryAfter.HasValue ? new RateLimiterRejectedException(source, retryAfter.Value) : new RateLimiterRejectedException(source);
```
martincostello, line 69: Needs changing to use the new constructors? Otherwise `source` is being used as the message.

peter-csala: :tired_face: :tired_face: :tired_face:

src/Polly.RateLimiting/RateLimiterRejectedException.cs
```diff
@@ -38,13 +38,33 @@ public RateLimiterRejectedException(string message)
     {
     }
 
+    /// <summary>
+    /// Initializes a new instance of the <see cref="RateLimiterRejectedException"/> class.
+    /// </summary>
+    /// <param name="message">The message that describes the error.</param>
+    /// <param name="telemetrySource">The source pipeline and strategy names.</param>
```
martincostello, line 45: Let's just make this something like "A string representing the source of the telemetry". The format is opaque to the exception itself.

src/Polly.RateLimiting/RateLimiterRejectedException.cs
```diff
@@ -74,6 +116,11 @@ public RateLimiterRejectedException(string message, TimeSpan retryAfter, Excepti
     /// </remarks>
     public TimeSpan? RetryAfter { get; }
 
+    /// <summary>
+    /// Gets the name of the strategy which has thrown the exception.
```
martincostello, line 120:
```suggestion
    /// Gets the source of the strategy which has thrown the exception, if known.
```

src/Polly.RateLimiting/RateLimiterRejectedException.cs
```diff
@@ -96,14 +149,8 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
...
-        if (RetryAfter.HasValue)
-        {
-            info.AddValue("RetryAfter", RetryAfter.Value.TotalSeconds);
-        }
-        else
-        {
-            info.AddValue("RetryAfter", -1.0);
-        }
+        info.AddValue(nameof(TelemetrySource), TelemetrySource ?? "(null)/(null)/(null)");
+        info.AddValue(nameof(RetryAfter), RetryAfter.HasValue ? RetryAfter.Value.TotalSeconds : -1.0);
```
martincostello, line 153: Not to do with your changes, but I wonder if reading and writing ticks rather than total seconds would be better in terms of resolution?

src/Polly.RateLimiting/RateLimiterResilienceStrategy.cs
```diff
@@ -5,6 +5,7 @@ namespace Polly.RateLimiting;
 
 internal sealed class RateLimiterResilienceStrategy : ResilienceStrategy, IDisposable, IAsyncDisposable
 {
+    private const string Message = "The operation could not be executed because it was rejected by the rate limiter.";
```
martincostello, line 8: As it's only used in one method, I think it would be better to declare this constant inline in the method body.

src/Polly.RateLimiting/RateLimiterRejectedException.cs
martincostello: Ignore this - I thought we'd recently added this but it hadn't been released yet, but actually that was `BrokenCircuitException`, so changing this would break compatibility.

src/Polly.RateLimiting/RateLimiterRejectedException.cs
```diff
@@ -74,6 +116,11 @@ public RateLimiterRejectedException(string message, TimeSpan retryAfter, Excepti
     /// </remarks>
     public TimeSpan? RetryAfter { get; }
 
+    /// <summary>
+    /// Gets the source of the strategy which has thrown the exception, if known.
+    /// </summary>
+    public string? TelemetrySource { get; }
```
martintmk, line 122: Why not use telemetry source here? The strategy name is not enough to uniquely identify the strategy.

martincostello: See https://github.com/App-vNext/Polly/pull/2346#discussion_r1803160193

src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
martintmk: Why not just expose this property? 

https://github.com/App-vNext/Polly/blob/ea15b525317d96af934ada7003b9ff3b6f6d52af/src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs#L22

The `ResilienceStrategyTelemetry` is already public. I would indeed suggest to make the constructor of `ResilienceTelemetrySource` public, there is no reason for us to hinder the testability.

src/Polly.RateLimiting/RateLimiterRejectedException.cs
martintmk: Thanks, added a comment :)

src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
peter-csala: @martintmk That was the original plan. But Martin suggested to do not expose that property rather have a method which returns a string representation of the `TelemetrySource`. Both of the approaches feel a bit clumsy so I don't have a preference.

What do you guys think? Which direction should we go?

martintmk: I played around with the idea on how we can uniformly introduce this to all exceptions:

https://github.com/App-vNext/Polly/pull/2352

wdyt?

martintmk:
LGTM, great job! 

How about to update our docs too in follow-up?

martintmk:
@martincostello is this ok to merge?

src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
```diff
@@ -308,11 +311,13 @@ private void SetLastHandledOutcome_NeedsLock(Outcome<T> outcome)
     private BrokenCircuitException CreateBrokenCircuitException()
     {
         TimeSpan retryAfter = _blockedUntil - _timeProvider.GetUtcNow();
-        return _breakingException switch
+        BrokenCircuitException exception = _breakingException switch
```
martincostello, line 314: `var`?

src/Polly.Core/ExecutionRejectedException.cs
```diff
@@ -49,4 +51,10 @@ protected ExecutionRejectedException(SerializationInfo info, StreamingContext co
     }
 #endif
 #pragma warning restore RS0016 // Add public types and members to the declared API
+
+    /// <summary>
+    /// Gets the source of the strategy which has thrown the exception, if known.
+    /// </summary>
+    public virtual ResilienceTelemetrySource? TelemetrySource { get; internal set; }
+
```
martincostello, line 59:
```suggestion
```

src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
```diff
@@ -21,6 +23,18 @@ internal ResilienceStrategyTelemetry(ResilienceTelemetrySource source, Telemetry
...
+    ///  Updates the source of the telemetry on the provided exception.
+    /// </summary>
+    /// <param name="exception">The to-be-updated exception.</param>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public void UpdateTelemetrySource(ExecutionRejectedException exception)
+    {
+        Guard.NotNull(exception);
+
+        exception.TelemetrySource = TelemetrySource;
+    }
```
martincostello, line 36: Rather than `Update` maybe `Set` would be a better verb?

src/Polly.RateLimiting/RateLimiterResilienceStrategy.cs
```diff
@@ -65,7 +65,11 @@ protected override async ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState
             await OnLeaseRejected(new OnRateLimiterRejectedArguments(context, lease)).ConfigureAwait(context.ContinueOnCapturedContext);
         }
 
-        var exception = retryAfter.HasValue ? new RateLimiterRejectedException(retryAfter.Value) : new RateLimiterRejectedException();
+        var exception = retryAfter.HasValue
+            ? new RateLimiterRejectedException(retryAfter.Value)
+            : new RateLimiterRejectedException();
```
martincostello, line 70: Then we don't pay (IIRC) pay the penalty of `.Value` checking `HasValue` internally?

```suggestion
        var exception = retryAfter is { } retryAfterValue
            ? new RateLimiterRejectedException(retryAfterValue)
            : new RateLimiterRejectedException();
```

src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
```diff
@@ -24,11 +24,11 @@ internal ResilienceStrategyTelemetry(ResilienceTelemetrySource source, Telemetry
     internal ResilienceTelemetrySource TelemetrySource { get; }
 
     /// <summary>
-    ///  Updates the source of the telemetry on the provided exception.
+    ///  Sets the source of the telemetry on the provided exception.
     /// </summary>
     /// <param name="exception">The to-be-updated exception.</param>
```
martincostello, line 29:
```suggestion
    /// <param name="exception">The to-be-set exception.</param>
```

martincostello (approved):

PR #2352, opened 2024-10-21
Introduce `ExecutionRejectedException.TelemetrySource` property

martintmk:
Just an experiment on how we can uniformly introduce telemetry source to all our exceptions.

The idea is that rather to introduce many new overloads to exception constructors, we can update this property centrally. It's niche enough< so my thinking is we don't need to expose this parameter in exception constructor. If someone complains, we can do it later anyway.

Contributes to #2346 and #2345

cc @peter-csala, @martincostello 

Your thoughts?

# Pull Request

## Confirm the following

### DIFF
--- a/src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
+++ b/src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
@@ -95,7 +95,7 @@ internal sealed class CircuitStateController<T> : IDisposable
 
         lock (_lock)
         {
-            SetLastHandledOutcome_NeedsLock(Outcome.FromException<T>(new IsolatedCircuitException()));
+            SetLastHandledOutcome_NeedsLock(Outcome.FromException<T>(CreateIsolatedCircuitException()));
             OpenCircuitFor_NeedsLock(Outcome.FromResult<T>(default), TimeSpan.MaxValue, manual: true, context, out task);
             _circuitState = CircuitState.Isolated;
         }
@@ -143,7 +143,7 @@ internal sealed class CircuitStateController<T> : IDisposable
             {
                 CircuitState.Open => CreateBrokenCircuitException(),
                 CircuitState.HalfOpen when !isHalfOpen => CreateBrokenCircuitException(),
-                CircuitState.Isolated => new IsolatedCircuitException(),
+                CircuitState.Isolated => CreateIsolatedCircuitException(),
                 _ => null
             };
 
@@ -308,11 +308,21 @@ internal sealed class CircuitStateController<T> : IDisposable
     private BrokenCircuitException CreateBrokenCircuitException()
     {
         TimeSpan retryAfter = _blockedUntil - _timeProvider.GetUtcNow();
-        return _breakingException switch
+        var ex = _breakingException switch
         {
             Exception exception => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter, exception),
             _ => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter)
         };
+
+        _telemetry.UpdateTelemetrySource(ex);
+        return ex;
+    }
+
+    private IsolatedCircuitException CreateIsolatedCircuitException()
+    {
+        var ex = new IsolatedCircuitException();
+        _telemetry.UpdateTelemetrySource(ex);
+        return ex;
     }
 
     private void OpenCircuit_NeedsLock(Outcome<T> outcome, bool manual, ResilienceContext context, out Task? scheduledTask)
--- a/src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
+++ b/src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
@@ -60,6 +60,17 @@ public sealed class ResilienceStrategyTelemetry
         Listener.Write<TResult, TArgs>(new(TelemetrySource, resilienceEvent, context, args, outcome));
     }
 
+    /// <summary>
+    /// Updates the <see cref="ExecutionRejectedException.TelemetrySource"/> property of the exception.
+    /// </summary>
+    /// <param name="exception">The exception to update.</param>
+    public void UpdateTelemetrySource(ExecutionRejectedException exception)
+    {
+        Guard.NotNull(exception);
+
+        exception.TelemetrySource = TelemetrySource;
+    }
+
     internal void Report<TArgs, TResult>(ResilienceEvent resilienceEvent, TArgs args)
         where TArgs : IOutcomeArguments<TResult>
     {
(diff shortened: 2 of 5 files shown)


src/Polly.Core/Telemetry/ResilienceStrategyTelemetry.cs
```diff
@@ -60,6 +60,17 @@ public void Report<TArgs, TResult>(ResilienceEvent resilienceEvent, ResilienceCo
         Listener.Write<TResult, TArgs>(new(TelemetrySource, resilienceEvent, context, args, outcome));
     }
 
+    /// <summary>
+    /// Updates the <see cref="ExecutionRejectedException.TelemetrySource"/> property of the exception.
+    /// </summary>
+    /// <param name="exception">The exception to update.</param>
+    public void UpdateTelemetrySource(ExecutionRejectedException exception)
```
martintmk, line 67: We can even hide this initially with `[EditorBrowsable(EditorBrowsableState.Never)]` ?

src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
```diff
@@ -308,11 +308,21 @@ private void SetLastHandledOutcome_NeedsLock(Outcome<T> outcome)
...
     {
         TimeSpan retryAfter = _blockedUntil - _timeProvider.GetUtcNow();
-        return _breakingException switch
+        var ex = _breakingException switch
         {
             Exception exception => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter, exception),
             _ => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter)
         };
+
+        _telemetry.UpdateTelemetrySource(ex);
```
peter-csala, line 317: IMHO this method call is super easy to forget. Also if someone wants to construct any of the Polly exception (for whatever reason) then she has to be aware of this feature and has to use it.

peter-csala:
If we go with this approach to _decorate_ all Polly exceptions (that derives from the `ExecutionRejectedException`) then that we would generate consistent behavior across the strategies. BUT for those exceptions that are not Polly exceptions we don't have this. 

So, if we extend the Polly exceptions with `TelemetrySource` then IMHO we should also provide a workaround for non-Polly exceptions as well to be able to use the telemetry source regardless the thrown exception.

martintmk:
> If we go with this approach to _decorate_ all Polly exceptions (that derives from the `ExecutionRejectedException`) then that we would generate consistent behavior across the strategies. BUT for those exceptions that are not Polly exceptions we don't have this.
> 
> So, if we extend the Polly exceptions with `TelemetrySource` then IMHO we should also provide a workaround for non-Polly exceptions as well to be able to use the telemetry source regardless the thrown exception.



> If we go with this approach to _decorate_ all Polly exceptions (that derives from the `ExecutionRejectedException`) then that we would generate consistent behavior across the strategies. BUT for those exceptions that are not Polly exceptions we don't have this.
> 
> So, if we extend the Polly exceptions with `TelemetrySource` then IMHO we should also provide a workaround for non-Polly exceptions as well to be able to use the telemetry source regardless the thrown exception.

How about we treat the `UpdateTelemetrySource` as an pub-internal API (albeit hidden, so it can be used in `Polly.RateLimiting`). We will add the `TelemetrySource` property to all our exceptions. 

If someone complains that they want to attach this property when constructing the exceptions, we can do the follow-up where we expose a new set of constructors for all relevant exceptions?

Alternatively, we can do it now. One thing I feel quite convinced about is to expose the full `ResilienceTelemetrySource` property on the exception, as having just string identifier is ambiguous and cannot identify the resilience strategy properly.

peter-csala:
> How about we treat the `UpdateTelemetrySource` as an pub-internal API (albeit hidden, so it can be used in `Polly.RateLimiting`). We will add the `TelemetrySource` property to all our exceptions.
> 
> If someone complains that they want to attach this property when constructing the exceptions, we can do the follow-up where we expose a new set of constructors for all relevant exceptions?
> 
> Alternatively, we can do it now. One thing I feel quite convinced about is to expose the full `ResilienceTelemetrySource` property on the exception, as having just string identifier is ambiguous and cannot identify the resilience strategy properly.

I'm okay with this approach. Do you want to finish this PR or shall I incorporate the idea into my PR?

martintmk:
> > How about we treat the `UpdateTelemetrySource` as an pub-internal API (albeit hidden, so it can be used in `Polly.RateLimiting`). We will add the `TelemetrySource` property to all our exceptions.
> > If someone complains that they want to attach this property when constructing the exceptions, we can do the follow-up where we expose a new set of constructors for all relevant exceptions?
> > Alternatively, we can do it now. One thing I feel quite convinced about is to expose the full `ResilienceTelemetrySource` property on the exception, as having just string identifier is ambiguous and cannot identify the resilience strategy properly.
> 
> I'm okay with this approach. Do you want to finish this PR or shall I incorporate the idea into my PR?

If you are OK with incorporating this, please continue in your PR :)

martincostello:
I'm not against the idea of the pub-ternal API. I think if we go down that route it's better to do the minimal amount and do more later if asked for, otherwise it's going to end up being quite a big change in terms of additional API surface area.

martintmk:
> I'm not against the idea of the pub-ternal API. I think if we go down that route it's better to do the minimal amount and do more later if asked for, otherwise it's going to end up being quite a big change in terms of additional API surface area.

this proposal really only adds one new pub API - i.e. `ExecutionRejectedException.TelemetrySource` so that works perfectly with your comment :) 


@peter-csala please go ahead!

peter-csala:
> > I'm not against the idea of the pub-ternal API. I think if we go down that route it's better to do the minimal amount and do more later if asked for, otherwise it's going to end up being quite a big change in terms of additional API surface area.
> 
> this proposal really only adds one new pub API - i.e. `ExecutionRejectedException.TelemetrySource` so that works perfectly with your comment :)
> 
> @peter-csala please go ahead!

Unfortunately no :( The `ExecutionRejectedException` is used by the Polly V7 API as well... So,  the old `RateLimitRejectedException`, and `BulkheadRejectedException` would be impacted as well.

I try to extend only the Polly.Core and Polly.RateLimiting exceptions.

martintmk:
> > > I'm not against the idea of the pub-ternal API. I think if we go down that route it's better to do the minimal amount and do more later if asked for, otherwise it's going to end up being quite a big change in terms of additional API surface area.
> > 
> > 
> > this proposal really only adds one new pub API - i.e. `ExecutionRejectedException.TelemetrySource` so that works perfectly with your comment :)
> > @peter-csala please go ahead!
> 
> Unfortunately no :( The `ExecutionRejectedException` is used by the Polly V7 API as well... So, the old `RateLimitRejectedException`, and `BulkheadRejectedException` would be impacted as well.
> 
> I try to extend only the Polly.Core and Polly.RateLimiting exceptions.

I would say it's ok, it's nullable anyway so on V7 this property will never be filled.

martintmk:
Duplicate of #2346

PR #2374, opened 2024-11-11
Increase test timeout

martincostello:
Increase the timeout on the `InfiniteRetry_Delay_Does_Not_Overflow_2163` test to (hopefully) reduce test flakiness in CI.


martincostello:
Well that didn't work 😅 - will just close rather than make bigger and bigger.

PR #2395, opened 2024-11-15
Parallelise mutation tests

martincostello:
- Run each mutation test project in parallel in its own GitHub Actions job.
- Install the .NET 8 SDK when required now we only explicitly install .NET 6 and 9.
- Minimise redundant work when running mutation tests.


martincostello:
Full end-to-end mutation tests job down from 45 minutes to 25 minutes.

martintmk (approved):

joelhulen:
@martincostello , I've updated the rules. Please let me know if you need anything else.

martincostello:
@joelhulen Could you change it to `mutation-tests` instead of `mutations-testing` please? The former is the aggregate job, the latter is just one of the legs.

joelhulen:
I'm sorry, @martincostello. These two were too similarly named! I've removed the wrong one and added the correct one.

Issue #2375, opened 2024-11-11
[Bug]: Inconsistent behavior when canceling a retry


kmcclellan:
### Describe the bug

If I cancel an execution that has been handled by a retry policy, it may or may not return the last executed outcome.

### Expected behavior

There is some question as to what should happen when you trigger cancellation for a resilience strategy that has already invoked the callback but has more work to do. On the one hand, the caller may want to know that execution did not complete per configuration (esp. if the intent is to retry indefinitely). On the other, the invocation may have had side effects, and we do have an outcome that the caller might be interested in.

In my opinion, callbacks passed to resilience pipelines shouldn't make assumptions about how they will be executed and should include a `catch` to clean up any side-effects. Therefore, a retry strategy that _would have_ attempted additional invocations should acknowledge cancellation by throwing `OperationCanceledException` regardless of whether it has executed the callback or what delay configuration is in place.

However, if instead we want the retry strategy to return the last executed outcome, it should do so consistently and refrain from acknowledging cancellation if invocation has completed at least once.

### Actual behavior

If the cancellation token is in the requested state at the time an invocation of the callback completes, [that outcome is returned to the caller](https://github.com/App-vNext/Polly/blob/7567acc330c9f74d6c7b14897a8ef4d15e9d475a/src/Polly.Core/Retry/RetryResilienceStrategy.cs#L72). If the token is instead triggered during the subsequent delay between attempts, [an `OperationCanceledException` outcome is returned](https://github.com/App-vNext/Polly/blob/7567acc330c9f74d6c7b14897a8ef4d15e9d475a/src/Polly.Core/Retry/RetryResilienceStrategy.cs#L110).

### Steps to reproduce

```csharp
using Polly;

var pipeline = new ResiliencePipelineBuilder<int>()
    .AddRetry(
        new()
        {
            ShouldHandle = x => ValueTask.FromResult(x.Outcome.Result < 10),
            MaxRetryAttempts = 10,
            Delay = TimeSpan.FromSeconds(2),
        })
    .Build();

await TestCancel(TimeSpan.FromSeconds(1));
await TestCancel(TimeSpan.FromSeconds(3));
await TestCancel(TimeSpan.FromSeconds(5));
await TestCancel(TimeSpan.FromSeconds(7));

async Task TestCancel(TimeSpan duration)
{
    try
    {
        using var cancellation = new CancellationTokenSource(duration);
        var n = 0;

        var result = await pipeline.ExecuteAsync(
[...]

PR #2396, opened 2024-11-18
Handle CancellationToken for Retry

peter-csala:
# Pull Request

#2375 

- Retry: handle cancelled `CancellationToken` as `OperationCanceledException`
- Assessment
  - Hedging: handles cancelled `CancellationToken` as `OperationCanceledException`
  - Timeout: does not handle cancelled `CancellationToken` as `OperationCanceledException`
    - We might want to tackle that as a separate PR 

## Confirm the following

### DIFF
--- a/src/Polly.Core/Retry/RetryResilienceStrategy.cs
+++ b/src/Polly.Core/Retry/RetryResilienceStrategy.cs
@@ -53,13 +53,9 @@ internal sealed class RetryResilienceStrategy<T> : ResilienceStrategy<T>
         {
             var startTimestamp = _timeProvider.GetTimestamp();
             var outcome = await StrategyHelper.ExecuteCallbackSafeAsync(callback, context, state).ConfigureAwait(context.ContinueOnCapturedContext);
-            try
+            if (context.CancellationToken.IsCancellationRequested)
             {
-                context.CancellationToken.ThrowIfCancellationRequested();
-            }
-            catch (OperationCanceledException e)
-            {
-                outcome = Outcome.FromException<T>(e);
+                outcome = Outcome.FromException<T>(new OperationCanceledException(context.CancellationToken));
             }
 
             var shouldRetryArgs = new RetryPredicateArguments<T>(context, outcome, attempt);
--- a/src/Polly.Core/Retry/RetryResilienceStrategy.cs
+++ b/src/Polly.Core/Retry/RetryResilienceStrategy.cs
@@ -76,7 +76,7 @@ internal sealed class RetryResilienceStrategy<T> : ResilienceStrategy<T>
                 TelemetryUtil.ReportExecutionAttempt(_telemetry, context, outcome, attempt, executionTime, handle);
             }
 
-            if (isLastAttempt || !handle)
+            if (context.CancellationToken.IsCancellationRequested || isLastAttempt || !handle)
             {
                 return outcome;
             }
--- a/src/Polly.Core/Retry/RetryResilienceStrategy.cs
+++ b/src/Polly.Core/Retry/RetryResilienceStrategy.cs
@@ -53,6 +53,15 @@ internal sealed class RetryResilienceStrategy<T> : ResilienceStrategy<T>
         {
             var startTimestamp = _timeProvider.GetTimestamp();
             var outcome = await StrategyHelper.ExecuteCallbackSafeAsync(callback, context, state).ConfigureAwait(context.ContinueOnCapturedContext);
+            try
+            {
+                context.CancellationToken.ThrowIfCancellationRequested();
+            }
+            catch (OperationCanceledException e)
+            {
+                outcome = Outcome.FromException<T>(e);
+            }
+
             var shouldRetryArgs = new RetryPredicateArguments<T>(context, outcome, attempt);
             var handle = await ShouldHandle(shouldRetryArgs).ConfigureAwait(context.ContinueOnCapturedContext);
             var executionTime = _timeProvider.GetElapsedTime(startTimestamp);
@@ -67,7 +76,7 @@ internal sealed class RetryResilienceStrategy<T> : ResilienceStrategy<T>
                 TelemetryUtil.ReportExecutionAttempt(_telemetry, context, outcome, attempt, executionTime, handle);
             }
 
-            if (context.CancellationToken.IsCancellationRequested || isLastAttempt || !handle)
+            if (isLastAttempt || !handle)
             {
                 return outcome;
             }
--- a/test/Polly.Core.Tests/Retry/RetryResilienceStrategyTests.cs
+++ b/test/Polly.Core.Tests/Retry/RetryResilienceStrategyTests.cs
@@ -105,6 +105,37 @@ public class RetryResilienceStrategyTests
     }
 
     [Fact]
+    public async Task ExecuteAsync_CancellationRequestedDuringDelay_EnsureNotRetried()
+    {
+        using var cancellationToken = new CancellationTokenSource();
+        _options.Delay = TimeSpan.FromMilliseconds(100);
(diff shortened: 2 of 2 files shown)


peter-csala:
It's kinda weird why it is complaining now about this line:

Screenshot 2024-11-18 at 12 42 09

peter-csala:
It seems like the timeout strategy also does not handle the linked CancellationToken's cancellation as `OperationCanceledException`

The following code prints `1` as a result
```cs
var cts = new CancellationTokenSource();
var p = new ResiliencePipelineBuilder<int>().AddTimeout(TimeSpan.FromSeconds(1)).Build();
var r = await p.ExecuteAsync(async token =>
{
	await Task.Delay(500, token);
	cts.Cancel();
	return 1;
}, cts.Token);

Console.WriteLine(r); // 1
```

peter-csala:
@martincostello Could you please rerun the `windows-latest` step? It seems to me it is just a [transient issue](https://github.com/App-vNext/Polly/actions/runs/11894363722/job/33141430063?pr=2396).

martincostello (approved):

src/Polly.Core/Retry/RetryResilienceStrategy.cs
```diff
@@ -53,6 +53,15 @@ protected internal override async ValueTask<Outcome<T>> ExecuteCore<TState>(Func
         {
             var startTimestamp = _timeProvider.GetTimestamp();
             var outcome = await StrategyHelper.ExecuteCallbackSafeAsync(callback, context, state).ConfigureAwait(context.ContinueOnCapturedContext);
+            try
+            {
+                context.CancellationToken.ThrowIfCancellationRequested();
```
martintmk, line 58: How about just checking for `context.CancellationToken.IsCancellationRequested()` and then constructing the exception manually?

Alternatively, you can do the same check inside `ExecuteCallbackSafeAsync` and it will apply everywhere where it is used.

peter-csala: I tried to follow the same pattern what we already have for the `_timeProvider.DelayAsync`:

https://github.com/App-vNext/Polly/blob/4a44d0e79d75a203f6ec5ef3d0887c196d8fe110/src/Polly.Core/Retry/RetryResilienceStrategy.cs#L106-L113

But yes for sure, I can rewrite the code to avoid try catch. Do you want?

martintmk: But yes for sure, I can rewrite the code to avoid try catch. Do you want?

Let's try please. I distinctly remember that any try-catch block had some impact on the micro benchmarks.

src/Polly.Core/Retry/RetryResilienceStrategy.cs
```diff
@@ -53,6 +53,11 @@ protected internal override async ValueTask<Outcome<T>> ExecuteCore<TState>(Func
         {
             var startTimestamp = _timeProvider.GetTimestamp();
             var outcome = await StrategyHelper.ExecuteCallbackSafeAsync(callback, context, state).ConfigureAwait(context.ContinueOnCapturedContext);
+            if (context.CancellationToken.IsCancellationRequested)
+            {
+                outcome = Outcome.FromException<T>(new OperationCanceledException(context.CancellationToken));
```
kmcclellan, line 58: I think it would be better to acknowledge cancellation _only when_ the retry policy would handle the outcome. It's important that a _valid_ outcome be surfaced by the policy, since the choice to continue processing may well have been deliberate (cf. item three of [MS recommendations](https://devblogs.microsoft.com/premier-developer/recommended-patterns-for-cancellationtoken/)).

peter-csala: Before this PR:
- If the cancellation is requested before the execution of a given attempt (either the original or any handled retry attempt) then it will short cut the execution with an OCE.
- If the cancellation is requested during the `OnRetry` then it will short cut the execution with an OCE.
- If the cancellation is requested during the waiting of the retry delay then it will short cut the execution with an OCE.

This PR handles the following case:
- If the cancellation is requested during the execution of the user callback then regardless of the outcome then it will short cut the execution with an OCE.

Just to clarify: are you asking to
- return the outcome of the user callback regardless of the cancellation was requested (or not) if the strategy won't handle the outcome of the user callback
- return the outcome of the user callback if the strategy would handle the outcome and it was the last attempt
- short cut the execution with an OCE if the strategy would handle the outcome it was not the last attempt

Is my understanding correct?

src/Polly.Core/Retry/RetryResilienceStrategy.cs
```diff
@@ -53,6 +53,11 @@ protected internal override async ValueTask<Outcome<T>> ExecuteCore<TState>(Func
         {
             var startTimestamp = _timeProvider.GetTimestamp();
             var outcome = await StrategyHelper.ExecuteCallbackSafeAsync(callback, context, state).ConfigureAwait(context.ContinueOnCapturedContext);
```
martintmk, line 55: One thing to be aware. If the outcome with disposable result is produced and then you are replacing it with exception, it might lead to a memory leak. 

I think it might be good idea to encapsulate all this handling into `ExecuteCallbackSafeAsync` method.

peter-csala: If we would move the code inside the `ExecuteCallbackSafeAsync` then we don't have access to the `isLastAttempt`. @kmcclellan suggested to respect the cancellation request only under certain circumstances. 

Please see the other comment section for more details.

martintmk: > If we would move the code inside the ExecuteCallbackSafeAsync then we don't have access to the isLastAttempt.

The `ExecuteCallbackSafeAsync` is internal, we could pass that information to the function. Wdyt?

peter-csala: That function is used by almost all strategies, where the "retry" concept may or may not make any sense. Of cource we can pass true as a default `isLastAttempt` but it feels a bit awkward to me.

peter-csala: Are you still suggesting to add a new `isLastAttempt` parameter with default value? I'm fine with either approach, I just want to close this PR this year if possible 😁

martintmk: Aaah sorry about this, got busy with other stuff. I would say meh, we can add it, it's internal detail anyway. At least this logic will be encapsulated there.

So a new default value with `isLastAttempt  = true`. Retry strategy will provide its own value.

kmcclellan: Hi! I am the reporter of the bug. Hoping I can help move this forward to the right solution!

The concern about disposable outcomes being thrown away is not really relevant to this change. Abiding by the rules of cancellation, we should _not_ replace an outcome unless we have more work to do (in which case it was probably an exception and won't have a result to dispose).

> Don’t throw OperationCanceledException after you’ve completed the work, just because the token was signaled

It is only _handled outcomes that are not the last outcome_ which could have a disposable result thrown away. This is true whether we attempt another execution or acknowledge cancellation. It's fair to say that retry policies currently don't support handling disposable outcomes.

I'm not even sure what `ExecuteCallbackSafeAsync` would supposedly do with `isLastAttempt`. In order to dispose, it would also need to know that the strategy would handle the outcome.

peter-csala:
It seems like the conversation get stuck. If we can't find a solution which is acceptable for everyone then I will close this PR at the end of this week and let others give it a try to tackle the issue

peter-csala:
The discussion got stuck so, I'm closing this PR to let others give a try.

PR #2456, opened 2025-01-18
Fix for retry cancellation

kmcclellan:
# Pull Request

[[Bug]: Inconsistent behavior when canceling a retry](https://github.com/App-vNext/Polly/issues/2375)

Despite how the conversations around [the previous attempt](https://github.com/App-vNext/Polly/pull/2396), only a small change is needed to make this work in a way that is coherent.

Note my choice to avoid throwing until _after_ delay is calculated and the retry event is logged. I think this is most likely what a user will expect, since every failed attempt currently generates such a log, and the outcome was "handled" in the sense that it was prevented from surfacing.

This also takes care of @martintmk's concern about disposables, since `DisposeHelper` is also called before cancellation is acknowledged, just as though we were continuing with the retry. I've added an extra test case for this too.

## Confirm the following

### DIFF
--- a/src/Polly.Core/Retry/RetryResilienceStrategy.cs
+++ b/src/Polly.Core/Retry/RetryResilienceStrategy.cs
@@ -67,7 +67,7 @@ internal sealed class RetryResilienceStrategy<T> : ResilienceStrategy<T>
                 TelemetryUtil.ReportExecutionAttempt(_telemetry, context, outcome, attempt, executionTime, handle);
             }
 
-            if (context.CancellationToken.IsCancellationRequested || isLastAttempt || !handle)
+            if (isLastAttempt || !handle)
             {
                 return outcome;
             }
@@ -100,17 +100,20 @@ internal sealed class RetryResilienceStrategy<T> : ResilienceStrategy<T>
                 await DisposeHelper.TryDisposeSafeAsync(resultValue, context.IsSynchronous).ConfigureAwait(context.ContinueOnCapturedContext);
             }
 
-            // stryker disable once all : no means to test this
-            if (delay > TimeSpan.Zero)
+            try
             {
-                try
+                context.CancellationToken.ThrowIfCancellationRequested();
+
+                // stryker disable once all : no means to test this
+                if (delay > TimeSpan.Zero)
                 {
                     await _timeProvider.DelayAsync(delay, context).ConfigureAwait(context.ContinueOnCapturedContext);
                 }
-                catch (OperationCanceledException e)
-                {
-                    return Outcome.FromException<T>(e);
-                }
+
+            }
+            catch (OperationCanceledException e)
+            {
+                return Outcome.FromException<T>(e);
             }
 
             if (incrementAttempts)
--- a/test/Polly.Core.Tests/Retry/RetryResilienceStrategyTests.cs
+++ b/test/Polly.Core.Tests/Retry/RetryResilienceStrategyTests.cs
@@ -67,6 +67,51 @@ public class RetryResilienceStrategyTests
     }
 
     [Fact]
+    public async Task ExecuteAsync_CancellationRequestedDuringCallback_EnsureNotRetried()
+    {
+        using var cts = new CancellationTokenSource();
+        _options.ShouldHandle = _ => PredicateResult.True();
+
+        var sut = CreateSut(TimeProvider.System);
+        var context = ResilienceContextPool.Shared.Get(cts.Token);
+
+        var result = await sut.ExecuteOutcomeAsync(
+            (_, _) =>
+            {
+                cts.Cancel();
+                return Outcome.FromResultAsValueTask(new object());
+            },
+            context,
+            default(object));
+
+        result.Exception.Should().BeOfType<OperationCanceledException>();
+    }
+
+    [Fact]
+    public async Task ExecuteAsync_CancellationRequestedDuringCallback_EnsureResultDisposed()
+    {
+        using var cts = new CancellationTokenSource();
+        _options.ShouldHandle = _ => PredicateResult.True();
+
+        var sut = CreateSut(TimeProvider.System);
+        var context = ResilienceContextPool.Shared.Get(cts.Token);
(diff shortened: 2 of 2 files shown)


kmcclellan:
Strange test failure from the build. No changes were made to this code, and I am unable to reproduce locally...

```
Polly.Specs.Utilities.SystemClockSpecs.Sleep_ShouldThrow_WhenCancellationRequested [FAIL]

Error: Expected a <System.OperationCanceledException> to be thrown, but no exception was thrown.

  Failed Polly.Specs.Utilities.SystemClockSpecs.Sleep_ShouldThrow_WhenCancellationRequested [89 ms]
  Error Message:
   Expected a <System.OperationCanceledException> to be thrown, but no exception was thrown.
  Stack Trace:
     at FluentAssertions.Execution.XUnit2TestFramework.Throw(String message)
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
   at FluentAssertions.Specialized.DelegateAssertionsBase`2.ThrowInternal[TException](Exception exception, String because, Object[] becauseArgs)
   at FluentAssertions.Specialized.DelegateAssertions`2.Throw[TException](String because, Object[] becauseArgs)
```

src/Polly.Core/Retry/RetryResilienceStrategy.cs
```diff
@@ -100,17 +100,20 @@ protected internal override async ValueTask<Outcome<T>> ExecuteCore<TState>(Func
                 await DisposeHelper.TryDisposeSafeAsync(resultValue, context.IsSynchronous).ConfigureAwait(context.ContinueOnCapturedContext);
             }
 
-            // stryker disable once all : no means to test this
-            if (delay > TimeSpan.Zero)
+            try
             {
-                try
+                context.CancellationToken.ThrowIfCancellationRequested();
```
peter-csala, line 105: Calling the `ThrowIfCancellationRequested` here might be **too late**.  The `OnRetry` telemetry event is reported and the `OnRetry` callback is already executed. But there will be no new retry attempt if cancellation is requested.

test/Polly.Core.Tests/Retry/RetryResilienceStrategyTests.cs
```diff
@@ -66,6 +66,51 @@ public async Task ExecuteAsync_CancellationRequestedAfterCallback_EnsureNotRetri
         executed.Should().BeTrue();
     }
 
+    [Fact]
+    public async Task ExecuteAsync_CancellationRequestedDuringCallback_EnsureNotRetried()
```
peter-csala, line 70: As in my PR I think we should cover all cases whenever we are dealing with cancellation token:
- Before the callback execution
- During the callback execution
- After the callback execution
  - During the OnRetry execution
  - During the sleep

src/Polly.Core/Retry/RetryResilienceStrategy.cs
kmcclellan: I considered this. The event may be named "on retry," but the actual meaning is more like "on handled by the retry policy." Currently, any outcome that is not returned to the caller is considered "handled" and triggers this event. Users don't want to wait for the next attempt (which may or may not happen). We want to know the strategy was triggered since this tells us the callback completed and why the outcome was not returned.

Cancellation in .NET is cooperative. It's normal for code to finish doing certain important tasks until it comes to a better "stopping place" to acknowledge the cancellation. Logging/telemetry for what has just happened I think falls into this category. I would be open to skipping the "delay" calculation and logging a zero if cancellation is triggered, but I'm also not convinced this adds much value.

peter-csala: As of now the strategy works like this on high level for a retry attempt (happy path):
- user provided callback is executed
- telemetry event is reported
- delay is calculated
- onretry is executed
- delay is waited

The execution can be stopped due to the following circumstances: the outcome is not handled by the strategy, the attempts are exhausted. From one side treating the cancellation in a different way feels a bit odd. But I agree that if the user provided callback executed then the telemetry and `OnRetry` hook should be performed as well because they allow the consumers to get insights what happened.

The `OnRetryArguments` serves multiple purposes. It tells about the past (`outcome`, `duration`, etc.) but also shares some information about the future (`delay`). You can access the information whether the cancellation was requested via the context (`context.CancellationToken.IsCancellationRequested`) but since it is not a top-level field I have doubts that anyone has ever checked it. IMHO making this information as a top-level field would make the `0`/`-1` delay more meaningful by providing contextual information.

IMHO the best would be to have something like this:
```mermaid
flowchart TD
    Args[OnRetryAgruments]
    Current[CurrentAttempt]
    Next[NextAttempt]

    Args --> Current
    Args --> Next

    Current --> Duration
    Current --> Outcome
    Current --> A[etc.]

    Next --> IsCancelled
    Next --> Delay
    Next --> b[etc.]
```
Maybe in V9 😛

martintmk (commented):
When considering this particular fix, the things I am looking for are:

- The old outcome is correctly disposed if the new one is produced.
- The retry outcome (either success, error, cancellation) is properly reported and no telemetry associated with produced outcome is lost. 

If the cancellation happens even after we said we will do next attempt, that's ok with me. That's the nature of cancellation actually :)

test/Polly.Core.Tests/Retry/RetryResilienceStrategyTests.cs
martintmk: @kmcclellan Let's extend our tests with these cases and we should be fine to merge this.

kmcclellan: This is done, @peter-csala look good?

peter-csala (approved):

test/Polly.Core.Tests/Retry/RetryResilienceStrategyTests.cs
peter-csala: Yepp, thank you.

martintmk (approved):

PR #2459, opened 2025-01-19
Remove FluentAssertions

### LINKED ISSUE #2450: Migrate to Shouldly
In light of the recent license changes to FluentAssertions, as well as the fact that we're now two versions behind due to breaking changes in v7, for future maintainability and to avoid licensing questions based on our usage of it, we should look to move to an alternative library providing similar/equivalent functionality that uses a standard open source license.

For example, Microsoft open source repositories are electing to not use v8 despite them themselves being open source. I'm not sure why specifically, other than an abundance of caution, but I've already seen examples of users asking questions to _other_ open source repositories using FluentAssertions regarding licensing implications etc. The "easiest" thing to do to avoid such questions ourselves would be to just not use it.

A good candidate replacement for this is [Shouldly](https://github.com/shouldly/shouldly), which I happen to already use as my assertion library of choice anyway.

I already took at look at doing this yesterday, but it was a **big** change (mostly in Polly.Specs), so it's best done on a per-project basis (or smaller) scope to reduce the size of change as FluentAssertions is removed.

The most impactful difference between the two seems to be the use of the `Invoking()` extension method.

martincostello:
Complete migration from FluentAssertions to Shouldly.

Resolves #2450.

### DIFF
--- a/eng/Test.targets
+++ b/eng/Test.targets
@@ -6,6 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Using Include="Shouldly" />
     <Using Include="Xunit" />
   </ItemGroup>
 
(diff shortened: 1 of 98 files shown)


adamralph:
@martincostello it looks like there are still some dangling references that should be removed — https://github.com/search?q=repo%3AApp-vNext%2FPolly%20fluentassertions&type=code

martincostello:
Thanks - I'll remove two of them, but the third is in the release notes so is still historically accurate.

PR #2496, opened 2025-02-14
Investigate macOS issue with dotnet-validate

martincostello:
Investigate issue with restoring dotnet-validate on macOS. Working in some of my own projects, but why isn't it working here? Something to do with cake?

https://github.com/NuGetPackageExplorer/NuGetPackageExplorer/issues/1698

No issue here for instance, despite using the same tool: https://github.com/martincostello/openapi-extensions/pull/252


martincostello:
> Something to do with cake?

Nothing to do with cake. `dotnet tool restore` immediately after installing the SDK has the same issue.

PR #2500, opened 2025-02-14
Use Stryker Mutator Dashboard

martincostello:
Upload mutation test results to Stryker Mutator dashboard.


martincostello:
Not sure what's going on - there's no evidence of reports being uploaded, but if I try and do it locally with the API key I get an HTTP 401 error...

PR #2510, opened 2025-02-18
Debugging Stryker Performance

slang25:
Obviously not a real suggestion, just trying to figure something out...

### DIFF
--- a/eng/stryker-config.json
+++ b/eng/stryker-config.json
@@ -20,6 +20,7 @@
       "statement"
     ],
     "configuration": "Debug",
+    "disable-mix-mutants": true,
     "target-framework": "net8.0",
     "thresholds": {
       "high": 100,
--- a/eng/stryker-config.json
+++ b/eng/stryker-config.json
@@ -20,7 +20,7 @@
       "statement"
     ],
     "configuration": "Debug",
-    "disable-mix-mutants": true,
+    "disable-mix-mutants": false,
     "target-framework": "net8.0",
     "thresholds": {
       "high": 100,
--- a/test/Polly.Core.Tests/CircuitBreaker/CircuitBreakerResiliencePipelineBuilderTests.cs
+++ b/test/Polly.Core.Tests/CircuitBreaker/CircuitBreakerResiliencePipelineBuilderTests.cs
@@ -8,7 +8,7 @@ namespace Polly.Core.Tests.CircuitBreaker;
 public class CircuitBreakerResiliencePipelineBuilderTests
 {
 #pragma warning disable IDE0028
-    public static TheoryData<Action<ResiliencePipelineBuilder>> ConfigureData = new()
+    public static List<Action<ResiliencePipelineBuilder>> ConfigureData = new()
     {
         builder => builder.AddCircuitBreaker(new CircuitBreakerStrategyOptions
         {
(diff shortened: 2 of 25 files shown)


slang25:
@martincostello I'm done with looking into this, I'm not 100% sure what's going on but my suspicion is that it's partly down to [this issue here](https://github.com/stryker-mutator/stryker-net/issues/2005).

So with the way things are in the `main` branch, many of the core project tests use `[MemberData(...)]`, which is dynamic and causes Stryker to get confused and decide to run all tests for the related mutations. If we work around this, like in this PR by using `[InlineData(...)]`, then Stryker will have a better a trimmer mapping of tests to mutants.

_However_, this is then even slower 🙃  I think this is because running the mutants in mixed mode makes things faster in principle, but I _think_ that performance gain is being drowned by the occurrence of timeouts. The timeouts are set to the sum of baseline test durations of the tests in the batch, plus a little leeway, so mixed mutant mode can raise this timeout duration from something like 10 seconds to a minute. This is why I think it takes close to 30 minute for the overall run.

I'd suggest waiting for the issue above to be addressed in Stryker and then try disabling mixed mutant mode again.

cc @dupdob - in case you are interested or have a better idea of whats going on here. The TL;DR; is that the mutation tests for the `Polly.Core.Tests` project is taking roughly 30 minutes in CI, whereas locally I can get it down to 4 minutes with this branch.

dupdob:
Hi
I will try and help as much as I can.
> So with the way things are in the main branch, many of the core project tests use [MemberData(...)], which is dynamic and causes Stryker to get confused and decide to run all tests for the related mutations. If we work around this, like in this PR by using [InlineData(...)], then Stryker will have a better a trimmer mapping of tests to mutants.

I am afraid this does not work around the issue you are mentioning (see below for full explanation). This work around another kind of issue which is more rarely encountered:  `MemberData` can lead to tests being skipped if the mutation results in an exception during the test setup phase.

> The timeouts are set to the sum of baseline test durations of the tests in the batch, plus a little leeway, so mixed mutant mode can raise this timeout duration from something like 10 seconds to a minute. 

Actually, disabling mixed mutant will always result in longer running time:
1. Disabling mixed mutant result in one batch per mutation, while mutant mixing typically divide this number by 5-10.
1. running a batch incurs a fixed time (Vstest start and setup time which is a fraction of second), so this will result in 5-10 times more time spent in this set up time
1. Timeout computation is based on the precise list of executed tests for a given set of mutants, so the total will be exactly the same with our without mixed mutant, expect for the fixed/setup part (which will multiply).

The only valid/known reason s for having different running time between Ci and local is:
1. using different parameters (this one is obvious, it does not need further explanation)
2. having different hardware:  Stryker runs test sessions in parallel, and often, CI pipeline runs on VMs with a low core count, as the build is seen as a very sequential process. 

xUnit issue:
The issue you are mentioning refers to a general problem with the xUnit test runner that does not respect VsTest contract by running extra test cases after signaling that the test ended.
Stryker relies on VsTest data collector architecture to track test executions; but while data collectors get notifications on test start and test end events, there is nothing for unitary test results. Stryker track which mutations have been covered by some code continuously, so it can associate mutants to a test by tracking what happens between test start and test end events.
But Stryker cannot identify what is the exact cause of the mutant being executed between tests, because it can happen for manby reasons :
1. Some async code that kept running after some previous tests. (note this should be avoided as it will often result in erroneous result with Stryker)
1. Some testfixture/assembly setup code that executed some mutations
1. Some data generation code (such as `MemberData` and others) that executed some mutations.

As Stryker cannot identify the cause, any such mutations will be flagged as to be tested against all tests. Otherwise, you could get false negative results: a mutant flag as survived while it is actually killed by some test that Stryker fails to identify as covering said mutant.

slang25:
Thanks for the details @dupdob, a lot of leanings in there for me.

Regarding the mixed mutants though, with this branch you can with and without mixed mutants and see for yourself. In CI here we can see:
* [disabled mixed mutants - 16:51](https://github.com/App-vNext/Polly/actions/runs/13400438020/job/37429736146)
* [with mixed mutants - 26:57](https://github.com/App-vNext/Polly/actions/runs/13401359261/job/37432732592)

PR #2512, opened 2025-02-19
Remove ignored mutations

martincostello:
Just curious as to the impact on the mutation tests if we remove this, as Polly.Testing has to manually remove it to get any results at all anyway:

https://github.com/App-vNext/Polly/blob/4469d8392076e4025157f437dc5a9f7d8634bc7c/build.cake#L315-L318

### DIFF
--- a/eng/stryker-config.json
+++ b/eng/stryker-config.json
@@ -16,8 +16,6 @@
       "LogInformation"
     ],
     "ignore-mutations": [
-      "block",
-      "statement"
     ],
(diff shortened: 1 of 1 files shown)


martincostello:
The coverage we lose with the current settings appear to fall mainly into these buckets:

- Argument validation
- Argument validation where removing it falls through to another overload that has the same checks
- Checking that something that should have been disposed was disposed

Have opened #2514 to fix some of the coverage gaps.

PR #2520, opened 2025-02-21
Improve mutation test perf

slang25:
# Pull Request

These settings bring the runtime down to about 16 minutes for `Polly.Core.Tests`.

## Confirm the following

### DIFF
--- a/.github/workflows/mutation-tests.yml
+++ b/.github/workflows/mutation-tests.yml
@@ -22,7 +22,7 @@ permissions:
 jobs:
   mutations:
     name: 'mutations-${{ matrix.name }}'
-    runs-on: windows-latest
+    runs-on: ubuntu-latest
     timeout-minutes: 60
 
     strategy:
--- a/eng/stryker-config.json
+++ b/eng/stryker-config.json
@@ -19,6 +19,7 @@
       "block",
       "statement"
     ],
+    "concurrency": 4,
     "configuration": "Debug",
     "target-framework": "net8.0",
     "thresholds": {


.github/workflows/mutation-tests.yml
```diff
@@ -22,7 +22,7 @@ permissions:
 jobs:
   mutations:
     name: 'mutations-${{ matrix.name }}'
-    runs-on: windows-latest
+    runs-on: ubuntu-latest
```
martincostello, line 25: I see the build is green, but I'm going to have to go spelunking through the Git history on this.

We did try using Ubuntu in the past but it didn't work correctly, so we had to go with Windows for functional correctness even though it was slower.

If whatever issue that was has been since fixed though then great.

martincostello: I found https://github.com/stryker-mutator/stryker-net/issues/2741#issuecomment-1783828972, and that points to an open issue relating to Ubuntu not working.

Maybe that's fixed now too, but still worth checking a little deeper.

martincostello: So https://github.com/App-vNext/Polly/commit/41e06e47aaf773110702349bd3c058d44e68ea5b was when we stopped using Linux because it wasn't working with .NET 8.

Either I tried it again at some point and it still didn't work, or I didn't bother because https://github.com/stryker-mutator/stryker-net/issues/2585 was still open.

eng/stryker-config.json
```diff
@@ -19,6 +19,7 @@
       "block",
       "statement"
     ],
+    "concurrency": 4,
```
martincostello, line 22: Based on [`concurrency`](https://stryker-mutator.io/docs/stryker-js/configuration/#concurrency-number) and [Standard GitHub-hosted runners for public repositories](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories), looks like the default would have been `4` anyway?

> Set the concurrency of workers. This defaults to `n-1` where `n` is the number of logical CPU cores available on your machine, unless `n <= 4`, in that case it uses `n`.

martincostello: I think [this](https://github.com/stryker-mutator/stryker-net/blob/b873b6268c7bdd9ca7529ac50e6329d05327f389/src/Stryker.Options/Options/Inputs/ConcurrencyInput.cs#L18) solves the mystery.

```csharp
public override int? Default => Math.Max(Environment.ProcessorCount / 2, 1);
```

The documentation is wrong, so it's actually using a concurrency of `2` by default.

martincostello: Nope, turns out _actually_ that was the documentation for stryker-js. 🤦

[The docs for .NET are correct](https://github.com/stryker-mutator/stryker-net/blob/b873b6268c7bdd9ca7529ac50e6329d05327f389/docs/configuration.md?plain=1#L303), they're just not what Google found as the top result when I searched for the Stryker docs.

martincostello:
For posterity, these are the numbers for different configurations at the time of writing:

| **Commit** | **Core** | **Extensions** | **Legacy** | **RateLimiting** | **Tests** |
|:-----------|:---------|:---------------|:-----------|:-----------------|:----------|
| https://github.com/App-vNext/Polly/commit/6cd95ba6654756a1162f445c194249bef3dcc8a2 | 29m 50s | 5m 54s | 10m 38s | 4m 41s | 4m 47s |
| https://github.com/App-vNext/Polly/pull/2520/commits/df1c1ff0e73eaf727a7335c22a025b397aa61589 | 19m 19s | 2m 54s | 6m 29s | 2m 35s | 2m 55s |
| https://github.com/App-vNext/Polly/pull/2522/commits/c55e90ad360d9236c1965b88b107a333e2664483 | ⌛ | 3m 16s | 7m 30s | 2m 33s | 2m 50s |

martincostello (approved):

PR #2522, opened 2025-02-21
Run mutations on Ubuntu

martincostello:
Just seeing whether the speed is actually just switching OS.

Either way #2520 will be the one that gets merged.


martincostello:
EBCAK: https://github.com/App-vNext/Polly/pull/2520#discussion_r1965104205

PR #2589, opened 2025-04-14
Improve mutation tests performance by ensuring solution file isn't discovered

slang25:
I was doing some testing locally and I've noticed that what is happening with the Cake build script is that when mutation tests are triggered, because they are launched from the root of the repo, the solution file is automatically detected which causes all tests to run.

This should limit each mutation test to only run the tests from the matching project.


martincostello (approved):

martincostello:
Nice find - feels like a bit of a footgun on the part of Stryker to do that given that it was given a specific project, but 10 minutes less runtime is a definite improvement.

PR #2649, opened 2025-06-16
Fix permissions

martincostello:
Fix [permissions to generate an SBOM](https://github.com/anchore/sbom-action?tab=readme-ov-file#permissions).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,9 @@ jobs:
       package-version: ${{ steps.build.outputs.package-version }}
 
     permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -39,7 +39,9 @@ jobs:
       package-version: ${{ steps.build.outputs.package-version }}
 
     permissions:
+      attestations: write
       contents: read
+      id-token: write
```
martincostello:
```suggestion
    permissions:
      actions: read
      contents: write
```

PR #2440, opened 2025-01-10
Update to xunit v3

martincostello:
- Update to xunit v3.
- Fix some IDE0055 warnings.
- Invert some pre-processor directives.
- Use generic assertions where relevant.
- Use collection expressions where relevant.
- Remove redundant using statements.
- Improve some assertions.
- Avoid infinite loop and throw if test times out.
- Improve condition.

Resolves #2439.

PR #2664, opened 2025-07-04
Reduce async overhead

pentp:
Remove many unnecessary async awaits from the pipeline hot path.
Also cleaned up some excessive use of lambdas.


martincostello:
Thanks for your PR.

I'm not against any improvements that don't impact the public API surface, _but_ `ValueTask` was specifically adopted during the development of Polly v8 to reduce memory allocations so those parts of the changes raise some questions.

Have you run the latest benchmarks before and after these changes? We'd want to see some concrete numbers on any improvements and trade-offs that these changes might make.

Some parts could potentially be moved out into separate smaller PRs as they're pretty uncontroversial (e.g. using `ArgumentNullException.ThrowIfNull()`, the unused `Stopwatch`) so would get merged sooner.

martincostello:
I've cherry-picked some of the changes here into #2667, which has now been merged.

pentp:
The changes from `ValueTask` to `Task` are improvements - removing an `async ValueTask` wrapping method around a `Task`.
The other changes are just reducing async overhead - `ValueTask` is allocation free only for synchronously completed methods (or if using pooling) and even then the async machinery has overhead.

martincostello:
> Have you run the latest benchmarks before and after these changes? We'd want to see some concrete numbers on any improvements and trade-offs that these changes might make.

martincostello:
Thanks for taking the time to run the benchmarks - it looks like all the numbers have improved for both duration and memory consumption.

src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
```diff
@@ -285,9 +296,13 @@ private void CloseCircuit_NeedsLock(Outcome<T> outcome, bool manual, ResilienceC
...
             if (_onClosed is not null)
             {
-                _executor.ScheduleTask(() => _onClosed(args).AsTask(), context, out scheduledTask);
+                return _executor.ScheduleTask(() => _onClosed(args).AsTask());
             }
         }
+
+#pragma warning disable S4586
+        return null;
+#pragma warning restore S4586
```
martincostello: Rather than need to suppress things like this, could we just return `Task.CompletedTask` and do `!ReferenceEquals(task, Task.CompletedTask)` instead of `task != null` in the relevant places?

src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
```diff
@@ -311,21 +326,13 @@ private void SetLastHandledOutcome_NeedsLock(Outcome<T> outcome)
     private BrokenCircuitException CreateBrokenCircuitException()
     {
         TimeSpan retryAfter = _blockedUntil - _timeProvider.GetUtcNow();
-        var exception = _breakingException switch
-        {
-            Exception ex => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter, ex),
-            _ => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter)
-        };
+        var exception = new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter, _breakingException!);
```
martincostello: I think I'd prefer this the way it was, but we could pass the exception in and make it nullable to have it rely less on what the state is (or require the `!` in the caller):

```csharp
var inner = _breakingException;
exception = _circuitState switch
{
    CircuitState.Open => CreateBrokenCircuitException(inner),
    CircuitState.HalfOpen when !isHalfOpen => CreateBrokenCircuitException(inner),
    CircuitState.Isolated => new IsolatedCircuitException(),
    _ => null
};


private BrokenCircuitException CreateBrokenCircuitException(Exception? breakingException)
{
    var exception = breakingException switch
    {
        Exception ex => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter, ex),
        _ => new BrokenCircuitException(BrokenCircuitException.DefaultMessage, retryAfter)
    };
    _telemetry.SetTelemetrySource(exception);
    return exception;
}
```

src/Polly.Core/Hedging/Controller/TaskExecution.cs
```diff
@@ -99,34 +99,43 @@ public async ValueTask<bool> InitializeAsync<TState>(
         _activeContext = _cachedContext;
         _activeContext.InitializeFrom(primaryContext, _cancellationSource!.Token);
 
-        if (primaryContext.CancellationToken.CanBeCanceled)
-        {
-            _cancellationRegistration = primaryContext.CancellationToken.Register(o => ((CancellationTokenSource)o!).Cancel(), _cancellationSource);
-        }
+#if NET
+        _cancellationRegistration = primaryContext.CancellationToken.UnsafeRegister(static o => ((CancellationTokenSource)o!).Cancel(), _cancellationSource);
```
martincostello, line 103: Why the `Unsafe` version, and what makes it safe to use?

src/Polly.Core/Registry/ResiliencePipelineRegistry.TResult.cs
```diff
@@ -47,33 +47,31 @@ public bool TryGet(TKey key, [NotNullWhen(true)] out ResiliencePipeline<TResult>
...
+                key,
+                _builderNameFormatter(key),
+                _instanceNameFormatter?.Invoke(key),
+                configure);
 
-                return new ResiliencePipeline<TResult>(component, DisposeBehavior.Reject, contextPool);
-            });
+            (var contextPool, var component) = componentBuilder.CreateComponent();
+
+            return _pipelines.GetOrAdd(key, new ResiliencePipeline<TResult>(component, DisposeBehavior.Reject, contextPool));
```
martincostello, line 64: Is the intention here trying to avoid the cost of the closure?

If so, can it be refactored to use a local method and then just do the single `GetOrAdd()` rather than `TryGetValue()` and then `GetOrAdd()`?

src/Polly.Core/Registry/ResiliencePipelineRegistry.TResult.cs
```diff
@@ -47,33 +47,31 @@ public bool TryGet(TKey key, [NotNullWhen(true)] out ResiliencePipeline<TResult>
...
         public async ValueTask DisposeAsync()
         {
-            foreach (var strategy in _pipelines.Values)
+            foreach (var kv in _pipelines)
             {
-                await strategy.DisposeHelper.ForceDisposeAsync().ConfigureAwait(false);
+                await kv.Value.DisposeHelper.ForceDisposeAsync().ConfigureAwait(false);
             }
-
-            _pipelines.Clear();
```
martincostello, line 76: Why remove this line?

src/Polly.Core/Registry/ResiliencePipelineRegistry.cs
```diff
@@ -122,20 +122,16 @@ public ResiliencePipeline GetOrAddPipeline(TKey key, Action<ResiliencePipelineBu
...
+        var componentBuilder = new RegistryPipelineComponentBuilder<ResiliencePipelineBuilder, TKey>(
+            _activator,
+            key,
+            _builderNameFormatter(key),
+            _instanceNameFormatter?.Invoke(key),
+            configure);
+
+        (var contextPool, var component) = componentBuilder.CreateComponent();
+
+        return _pipelines.GetOrAdd(key, new ResiliencePipeline(component, DisposeBehavior.Reject, contextPool));
```
martincostello, line 134: Similar question here - this doesn't seem better at first glance as we're always creating the `componentBuilder` even if it doesn't need to be used.

src/Polly.Core/Registry/ResiliencePipelineRegistry.cs
```diff
@@ -232,22 +228,21 @@ public bool TryAddBuilder<TResult>(TKey key, Action<ResiliencePipelineBuilder<TR
...
         {
-            await pipeline.DisposeHelper.ForceDisposeAsync().ConfigureAwait(false);
+            await kv.Value.DisposeHelper.ForceDisposeAsync().ConfigureAwait(false);
         }
 
-        foreach (var disposable in registries)
+        foreach (var kv in _genericRegistry)
         {
-            await disposable.DisposeAsync().ConfigureAwait(false);
+            await ((IAsyncDisposable)kv.Value).DisposeAsync().ConfigureAwait(false);
```
martincostello, line 245:
```suggestion
            if (kv.Value is IAsyncDisposable asyncDisposable)
            {
                await asyncDisposable.DisposeAsync().ConfigureAwait(false);
            }
```

src/Polly.Core/Registry/ResiliencePipelineRegistry.cs
```diff
@@ -258,15 +253,12 @@ private GenericRegistry<TResult> GetGenericRegistry<TResult>()
...
-                _pipelineComparer,
-                _builderNameFormatter,
-                _instanceNameFormatter);
-        });
+        return (GenericRegistry<TResult>)_genericRegistry.GetOrAdd(typeof(TResult), new GenericRegistry<TResult>(
+            () => new ResiliencePipelineBuilder<TResult>(_activator()),
+            _builderComparer,
+            _pipelineComparer,
+            _builderNameFormatter,
+            _instanceNameFormatter));
```
martincostello, line 261: Same here regarding always allocating.

src/Polly.Core/Timeout/TimeoutResilienceStrategy.cs
```diff
@@ -83,11 +94,10 @@ protected internal override async ValueTask<Outcome<TResult>> ExecuteCore<TResul
...
     private static CancellationTokenRegistration CreateRegistration(CancellationTokenSource cancellationSource, CancellationToken previousToken)
     {
-        if (previousToken.CanBeCanceled)
-        {
-            return previousToken.Register(static state => ((CancellationTokenSource)state!).Cancel(), cancellationSource, useSynchronizationContext: false);
-        }
-
-        return default;
+#if NET
+        return previousToken.UnsafeRegister(static state => ((CancellationTokenSource)state!).Cancel(), cancellationSource);
```
martincostello, line 98: Same question here about `Unsafe` variant.

src/Polly.Core/Utils/Pipeline/PipelineComponent.cs
```diff
@@ -17,15 +17,23 @@ internal abstract ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState>(
...
-        .GetResult();
+        => ExecuteCore([DebuggerDisableUserUnhandledExceptions] static (context, state) =>
+        {
+            try
+            {
+                return new ValueTask<Outcome<TResult>>(new Outcome<TResult>(state.callback(context, state.state)));
+            }
+#pragma warning disable CA1031 // Do not catch general exception types
+            catch (Exception e)
+            {
```
martincostello:
```suggestion
#pragma warning disable CA1031
            catch (Exception e)
#pragma warning restore CA1031
            {
```

src/Polly.Core/Utils/Pipeline/ReloadableComponent.cs
```diff
@@ -38,64 +35,62 @@ internal override ValueTask<Outcome<TResult>> ExecuteCore<TResult, TState>(
...
-            }
-
-            DisposeRegistration();
-            TryRegisterOnReload();
-
-            _ = DisposeDiscardedComponentSafeAsync(previousComponent);
-        });
+        _tokenSource = CancellationTokenSource.CreateLinkedTokenSource([.. reloadTokens]);
+#if NET
+        _ = _tokenSource.Token.UnsafeRegister(static s => ((ReloadableComponent)s!).Reload(), this);
```
martincostello, line 51: +1

test/Polly.Core.Tests/CircuitBreaker/Controller/ScheduledTaskExecutorTests.cs
```diff
@@ -56,22 +48,20 @@ public async Task ScheduleTask_Multiple_EnsureExecutionSerialized()
...
-            out var task);
+            });
 
         executing.WaitOne();
 
-        scheduler.ScheduleTask(() => Task.CompletedTask, ResilienceContextPool.Shared.Get(CancellationToken), out var otherTask);
+        var otherTask = scheduler.ScheduleTask(() => Task.CompletedTask);
 
 #pragma warning disable xUnit1031 // Do not use blocking task operations in test method
-        otherTask.Wait(50, CancellationToken).ShouldBeFalse();
```
martincostello, line 74: Why revert these cancellation token usages? The field was added intentionally to reduce churn when we can adopt xunit v3 (#2440) as `CancellationToken.None` will be replaced with `TestContext.Current.CancellationToken`.

test/Polly.Core.Tests/Utils/Pipeline/PipelineComponentFactoryTests.cs
```diff
@@ -29,7 +29,7 @@ public class PipelineComponentFactoryTests
     public void WithDisposableCallbacks_NoCallbacks_ReturnsOriginalComponent(IEnumerable<Action> callbacks)
     {
         var component = Substitute.For<PipelineComponent>();
-        var result = PipelineComponentFactory.WithDisposableCallbacks(component, callbacks);
+        var result = PipelineComponentFactory.WithDisposableCallbacks(component, callbacks.ToList());
```
martincostello: I think you could use `[.. callbacks]` in these two places.

src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
pentp: Changing the inner exception parameter to nullable would be the correct solution, but unfortunately it means public API changes (though it's a safe change) and the build gives tons of errors about API changes, so I'd leave it for a separate PR.

src/Polly.Core/Hedging/Controller/TaskExecution.cs
pentp: The unsafe version doesn't capture any async state for the registration. It's safe to use because the callback doesn't do anything other than call `CancellationTokenSource.Cancel()` (any registrations on that 2nd CTS would have their own async state, if they need it).

src/Polly.Core/Registry/ResiliencePipelineRegistry.TResult.cs
pentp: Yes, this avoids the closure and delegate allocation. On .NET 8+ it could be made to work using `GetOrAdd` with a delegate also, but it's not worth the effort (it only saves a `GetHashCode` computation if the entry wasn't already in the dictionary, but has its own additional overhead also).

src/Polly.Core/Registry/ResiliencePipelineRegistry.TResult.cs
pentp: `ConcurrentDictionary.Clear` is expensive - it's basically allocating a whole new `ConcurrentDictionary` internally.

src/Polly.Core/Timeout/TimeoutResilienceStrategy.cs
pentp: This is also just avoiding any potential async state capture which isn't needed for the registration callback.

src/Polly.Core/Utils/Pipeline/ReloadableComponent.cs
pentp: This also avoids capturing any async state potentially indefinitely.

src/Polly.Core/Registry/ResiliencePipelineRegistry.cs
pentp: This only allocates if the key wasn't already in the registry so no change from previous behavior. The delegate based `GetOrAdd` doesn't provide any additional protection against race conditions, it just saves a bit on `GetHashCode` calculations (but not in this case as `TryGetValue` was called earlier anyway).

src/Polly.Core/Registry/ResiliencePipelineRegistry.cs
pentp: This always avoids the closure and delegate allocation, but only allocates a `componentBuilder` if it wasn't in the dictionary already (`TryGetValue` call above checks for that).

src/Polly.Core/Registry/ResiliencePipelineRegistry.cs
pentp: That doesn't match the previous behavior (unconditional cast) and the entires are all `GenericRegistry` instances, so always implement `IAsyncDisposable`.

src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
martincostello: I wasn't suggesting that, just to effectively do `if (inner is not null) { inner .ctor } else { the other .ctor }`.

martincostello:
Just one last comment and fixing the mutation tests and this looks good to merge.

src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
pentp: I still think the correct change is to change the inner exception parameter to be nullable, but I'll leave it for a separate PR since a public API change.

martincostello (approved):

PR #2676, opened 2025-07-21
Migrate from Cake .NET tool to Cake .NET SDK

devlead:
Sorry for the drive-by PR out of the blue - no hard feelings at all if it's closed, but happy to assist and support otherwise 😎

Recently, we launched the preview of [Cake.Sdk](https://cakebuild.net/blog/2025/07/dotnet-cake-cs), which enables Cake to be used with the new [dotnet run app.cs](https://devblogs.microsoft.com/dotnet/announcing-dotnet-run-app/) coming with .NET 10.

I noticed that you're starting to target .NET 10 in your `dotnet-vnext` branch, which this PR targets. I took a quick look at what would be needed to migrate your Cake script from the .NET tool to the SDK, and it turned out to be just a few lines that need to be changed:

- Rename `build.cake` to `cake.cs` and add SDK directive
- Update build.ps1 to use dotnet cake.cs instead of dotnet cake
- Add *Cake.Sdk* to global.json `msbuild-sdks` section
- Move *Cake.FileHelpers* and *Newtonsoft.Json* from `#addin` to `#package` directives
- Add `ProjectType=Test` property for analyzer support

This is just the things needed to work with the new runner, there's potential for refactoring and improvements, but wanted to have this PR as focused has possible.


Directory.Packages.props
```diff
@@ -54,5 +54,7 @@
     <PackageVersion Include="System.ValueTuple" Version="4.5.0" />
     <PackageVersion Include="xunit" Version="2.9.3" />
     <PackageVersion Include="xunit.runner.visualstudio" Version="3.1.3" />
+    <PackageVersion Include="Cake.FileHelpers" Version="7.0.0" />
+    <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
```
martincostello: Nit: sorting

cake.cs
```diff
@@ -1,16 +1,15 @@
+#:sdk Cake.Sdk
+#:package Cake.FileHelpers
+#:package Newtonsoft.Json
```
martincostello, line 3: This is great if it ties in with Directory.Packages.props as it means we'll get automated PRs to keep them up to date.

cake.cs
```diff
@@ -1,16 +1,15 @@
...
+#:property ProjectType=Test
```
martincostello, line 4: This seems a bit weird to me. I know that it's _running_ tests, but it doesn't feel like tests.

global.json
```diff
@@ -3,5 +3,8 @@
     "version": "10.0.100-preview.6.25358.103",
     "allowPrerelease": false,
     "rollForward": "latestMajor"
+  },
+  "msbuild-sdks": {
+    "Cake.Sdk": "5.0.25198.49-beta"
```
martincostello, line 8: I assume 5.0 will ship in November?

cake.cs
devlead: Available were ( https://github.com/App-vNext/Polly/tree/dotnet-vnext/eng/analyzers )
- Library
- Benchmark
- Test

Library didn't work as it requires a NuGet package with the same name as the project to exist. An alternative would be to create a Build config.

cake.cs
devlead: Yes, it ties in with CPM if a `Directory.Packages.prop` exists.

global.json
devlead: 6.0.x will ship in November. Major & Minor is aligned with Cake.Core, 5.0.x uses Cake Core/Common 5.0.
image

Directory.Packages.props
devlead: sure sorted

martincostello (approved):

PR #2717, opened 2025-08-22
Fix flaky mutants

martincostello:
Fix flaky mutants after task refactoring in #2664.

### DIFF
--- a/src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
+++ b/src/Polly.Core/CircuitBreaker/Controller/CircuitStateController.cs
@@ -162,6 +162,8 @@ internal sealed class CircuitStateController<T> : IDisposable
         }
 
         task = ExecuteScheduledTaskAsync(task, context);
+
+        // stryker disable once all : no means to test this
         if (!task.IsCompleted)
         {
             return WaitHalfOpenTask(task, context.ContinueOnCapturedContext);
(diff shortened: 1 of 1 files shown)


pentp:
Isn't the `Debug.Assert(task.IsCompleted)` assert below those conditions enough to fix the mutation tests? IIRC mutation tests always run in debug mode and those asserts should ensure that the opposite condition will fail, no?

martincostello:
They are, but I've seen two different failing mutations since the merge.

https://github.com/App-vNext/Polly/blob/015914566660f8a92ae7a200f6349d533ad7c8b9/eng/stryker-config.json#L23

pentp:
Oh, I think the reason is because `Debug.Assert` also gets mutated (removed) and if done together with the other mutation, then both will "survive". The reason I thought asserts would be enough is that in our codebase we have `"ignore-methods": [ "Assert" ]` in Stryker config and I had forgotten it's not the default.

martincostello:
I'd be happy to accept a PR to tweak our configuration to do the same and then revert these changes so the conditions are mutated again.

I'm going to merge this now though as 4-in-a-row failures in main and 3-in-a-row successes here _suggests_ the flaky CI is at least resolved for now (🤞).

martincostello:
Just had [another one](https://github.com/App-vNext/Polly/actions/runs/17159883406/job/48686139751?pr=2722) happen here:

https://github.com/App-vNext/Polly/blob/bb71bbed84ec885d456bd39c3b46d659e307feca/src/Polly.Core/Hedging/Controller/HedgingExecutionContext.cs#L132

> LogicalNotExpression to un-LogicalNotExpression mutation Survived (132:13)

PR #2725, opened 2025-08-27
Add moderation workflow

martincostello:
Trial the use of [github/ai-moderator](https://github.com/github/ai-moderator) to detect spam.


.github/actionlint.yml
martincostello, line 1: Workaround for https://github.com/rhysd/actionlint/issues/548.

PR #2740, opened 2025-09-06
Simplify code

martincostello:
Apply `AD0001` and `AD0002` code fixes.


martincostello:
Coverage fall suggests not all of these are strictly correct and change the semantics.

PR #3065, opened 2026-05-10
Pin runner images

martincostello:
Pin runner images to a specific OS, rather than using `latest`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,13 +42,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ windows-latest ]
         include:
-          - os: macos-latest
+          - os: macos-26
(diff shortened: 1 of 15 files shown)


.github/workflows/build.yml
```diff
@@ -42,14 +42,16 @@ jobs:
...
       matrix:
-        os: [ windows-latest ]
         include:
-          - os: macos-latest
-            os_name: macos
-          - os: ubuntu-latest
-            os_name: linux
-          - os: windows-latest
-            os_name: windows
+          - job-name: macos-latest
```
Frulfump, line 46: Should the job name change as well?

martincostello: No because otherwise the required statuses won't match and I can't update them so I've intentionally left them as latest.

PR #3074, opened 2026-05-19
Restrict System.Threading.RateLimiting dependency to pre .net 8

### LINKED ISSUE #3073: [Feature request]: Remove direct System.Threading.RateLimiting dependency on frameworks where not required

We've hit obscure and intermittent  'could not load file or assembly' issues loading the System.Threading.RateLimiting dll at runtime when using the .net package pruning feature (I suspect).

Our .net 10 app ends up with a dependency on v8.0 of System.Threading.RateLimiting due to the direct dependency on `>= 8.0` listed on [Polly.RateLimiting](https://www.nuget.org/packages/Polly.RateLimiting/#dependencies-body-tab) for all .net versions:

Image

I think the dependency is unnecessary for .net 8+ and removing it would mean that consumers just pulled in the framework version automatically

The dependency on System.Threading.RateLimiting and any other System.* dependencies would ideally be optional for versions of .net where they are provided by the framework already.

### Additional context

_No response_

mungojam:
# Pull Request

Closes #3073

I've followed the same convention as used for other framework specific dependencies

## Confirm the following

~- [ ]  I have included unit tests for the issue/feature~ n/a

### DIFF
--- a/src/Polly.RateLimiting/Polly.RateLimiting.csproj
+++ b/src/Polly.RateLimiting/Polly.RateLimiting.csproj
@@ -27,6 +27,6 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Polly.Core\Polly.Core.csproj" />
-    <PackageReference Include="System.Threading.RateLimiting" />
+    <PackageReference Include="System.Threading.RateLimiting" Condition="!$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))" />
   </ItemGroup>
 </Project>


mungojam:
@dotnet-policy-service agree

mungojam:
Perhaps I was wrong in my assumption that it's built into .net or perhaps it comes in a later version than 8. I'll do more digging and experimenting later on to confirm.

mungojam:
This isn't the right fix, I think it needs a .net 10 target for polly adding instead

PR #3094, opened 2026-06-01
Adds caller cancellation token propagation in hedging and timeout strategies

### LINKED ISSUE #3086: [Bug]: Timeout strategy does not propagate correct CancellationToken
### Describe the bug

This is similar to the stale issue #934, but since it's been 4 years and we are now talking about v8 resilience pipelines, I am choosing to file a new issue.

When a timeout strategy is present in a pipeline, Polly replaces the `ResilienceContext.CancellationToken` with an internal timeout token before invoking the user callback. It also registers the caller's original token so that cancelling the caller token cancels the internal timeout token. If the caller token is cancelled, downstream code observes cancellation on Polly's internal token and throws an `OperationCanceledException` carrying that internal token. Polly then propagates that original exception without restoring or translating the token back to the caller's original token.

This makes it difficult for calling code to distinguish caller-initiated cancellation from other failures (for example timeouts) by inspecting `OperationCanceledException.CancellationToken`.

Consider the following very common pattern:

```csharp
try
{
	await pipeline.ExecuteAsync(callback, callerCancellationToken);
}
catch (OperationCanceledException ex) when (ex.CancellationToken == callerCancellationToken)
{
	throw;
}
catch (Exception ex)
{
	throw new ApplicationException("Operation failed.", ex);
}
```

In this case the cancellation really did come from `callerCancellationToken`, but the exception filter fails because the exception carries Polly's internal token instead. This behavior leaks an implementation detail of the timeout strategy into the exception observed by the caller. The caller's cancellation token caused the operation to cancel, but the escaping exception does not carry that token.

That is especially problematic for libraries that need to distinguish cancellation from genuine operation failures. A very common pattern is to let caller cancellation pass through unchanged while wrapping other exceptions with domain-specific context. If token identity is not preserved across Polly's timeout strategy, that pattern becomes unreliable even though the cancellation originated upstream.

The correct behavior would be for timeout strategies to translate upstream-triggered internal-token cancellation back into an `OperationCanceledException` associated with the previous/caller token, while continuing to throw `TimeoutRejectedException` when the timeout itself fires.

### Expected behavior

If cancellation originates from the token passed to `ExecuteAsync(..., cancellationToken)`, the `OperationCanceledException` escaping `ExecuteAsync` should carry that same caller token, or Polly should otherwise document that callers must not rely on `OperationCanceledException.CancellationToken` identity after executing through strategies such as timeout.

Note: Timeouts internal to Polly are different. If the timeout strategy itself causes cancellation while the caller token is not cancelled, the current behavior of throwing `TimeoutRejectedException` makes sense. The surprising case is caller-token cancellation routed through Polly's internal linked timeout token.

### Actual behavior

When caller cancellation causes Polly's internal timeout token to be cancelled, the exception escaping `ExecuteAsync` incorrectly carries the internal callback token rather than the caller token passed to `ExecuteAsync`.

### Steps to reproduce

Here's a minimal test (using TUnit) that reproduces the issue.

```csharp
using Polly;

public class PollyCancellationTokenTests
{
	[Test]
	public async Task CallerCancellationThroughTimeoutStrategyPropagatesOriginalTokenAsync()
	{
		var pipeline =
			new ResiliencePipelineBuilder()
				.AddTimeout(TimeSpan.FromMinutes(1))
				.Build();

		using var cts = new CancellationTokenSource();

		var ex = await Assert
			.That(async () =>
			{
				await pipeline.ExecuteAsync(async cancellationToken =>
				{
					await cts.CancelAsync(); // Simulate cancellation request from upstream caller
					cancellationToken.ThrowIfCancellationRequested(); // Simulate cancellation response from downstream code
				}, cts.Token);
			})
			.Throws<OperationCanceledException>();

		await Assert
			.That(ex.CancellationToken)
			.IsEqualTo(cts.Token)
			.Because("CancellationToken that triggered the cancellation should be on the resulting exception.");
	}
}
```

The final assertion fails because `ex.CancellationToken` is not equal to `cts.Token`.

```text
Expected to be equal to System.Threading.CancellationToken
but received System.Threading.CancellationToken

at Assert.That(ex.CancellationToken).IsEqualTo(cts.Token)
```

### Polly version

8.6.6

### .NET Version

net10.0

DaRosenberg:
# Pull Request

Fixes #3086 (Timeout strategy does not propagate the caller's `CancellationToken`)

When a resilience pipeline contains a strategy that substitutes the execution `CancellationToken` with an internal one (timeout, and also hedging), a caller-initiated cancellation surfaces an `OperationCanceledException` whose `CancellationToken` is Polly's *internal* token rather than the caller's. This breaks the common pattern of letting caller cancellation pass through unchanged while wrapping other failures, because callers cannot reliably compare `OperationCanceledException.CancellationToken` to their own token.

### Goal

Polly should throw an `OperationCanceledException` carrying the caller's token **if and only if** the cancellation was actually caused by a cancellation request on that token — for any pipeline, regardless of which strategies it is composed of or how they are nested.

### Approach

A repo-wide audit shows that within `Polly.Core` exactly two strategies substitute the execution token: **timeout** (`TimeoutResilienceStrategy`) and **hedging** (`TaskExecution` via `ResilienceContext.InitializeFrom`). Every other strategy and the pipeline plumbing only *read* `context.CancellationToken`, so they already emit the correct token at their own level.

The fix therefore lives in those two strategies, via a small shared helper:

```csharp
// Polly.Core/Utils/OutcomeUtilities.cs
public static Outcome<T> WithCallerCancellation<T>(this Outcome<T> outcome, CancellationToken callerToken)
{
    if (callerToken.IsCancellationRequested
        && outcome.Exception is OperationCanceledException oce
        && oce.CancellationToken != callerToken)
    {
        return Outcome.FromException<T>(new OperationCanceledException(callerToken).TrySetStackTrace());
    }

    return outcome;
}
```

- **Timeout** applies it to the outcome it returns, using the token that was on the context *before* it substituted its own. The existing timeout-detection branch (which produces `TimeoutRejectedException`) is unchanged.
- **Hedging** applies it at its accepted-outcome return sites, using the upstream token it already captures for the duration of the hedged execution.

Because each substituting strategy normalizes back to *its own* previous token, the behavior composes correctly through arbitrary nesting: an inner timeout rewrites to the mid-level token, the outer timeout rewrites that to the caller's token, and so on. The simplest case (`AddTimeout` only) and deeply nested cases both end up with the caller's token.

### Design decisions and trade-offs

- **Surgical fix** We considered normalizing the exception once at the outermost pipeline execution boundary. That would also cover hypothetical third-party strategies that substitute the token, but it adds work to the universal execution hot path. We chose the surgical approach because it adds **zero overhead** to the common path (work happens only inside the two strategies, only on the cancellation branch) and provably covers every composition of the built-in strategies. The trade-off is that a custom strategy that substitutes the token would remain responsible for its own token handling.
- **Exception shape** When the token must be corrected we create a bare `new OperationCanceledException(callerToken).TrySetStackTrace()`, matching the existing convention in `DelegatingComponent`, `CompositeComponent`, and hedging's pre-execution cancellation check. We deliberately did not chain the original exception as an `InnerException`; it keeps the behavior consistent with the rest of the codebase, at the cost of not preserving the original deep stack trace in this specific path.
- **Scope: v8 (`Polly.Core`) only.** The legacy v7 `Policy` API uses a combined linked token and already documents (in `AsyncTimeoutEngine`/`TimeoutEngine`) that the token on the exception is not reliable for this determination. We left v7 untouched to avoid changing long-stable behavior.
- **Deliberately ignoring a benign race** Detection relies on `callerToken.IsCancellationRequested`, read after the exception has been produced. Because a `CancellationToken` is a monotonic latch with no record of *when* or *why* it fired, there is an inherent, unavoidable race: if a non-caller cause produces the exception and the caller then cancels within the small window before we inspect the token, the cancellation is attributed to the caller. We chose not to add machinery to fight this, for three reasons:
  1. The timeout-vs-caller variant of this race is **pre-existing** — Polly's existing timeout classification already uses the same `previousToken.IsCancellationRequested` proxy, so this change does not make it worse (it only makes the resulting token more coherent).
  2. It cannot be fully eliminated: the token model exposes no causal/temporal information after the fact.
  3. In every misfire the caller's token genuinely *is* cancelled, so attributing the cancellation to the caller is defensible rather than harmful. The "only if" guarantee that matters in practice — *not* claiming caller cancellation when the caller's token was never cancelled — is fully preserved.

### Tests

- A regression test file under `Issues` (`IssuesTests.CancellationTokenPropagation_3086.cs`) with end-to-end coverage: the exact issue repro, timeout±retry in both orders, hedging, nested timeouts, a no-substitution baseline, and two "only if" guards (a real timeout still throws `TimeoutRejectedException`; an unrelated `OperationCanceledException` is preserved when the caller token is not cancelled)
- Unit tests for the new helper covering all branches
- Focused token-identity assertions added to the timeout and hedging strategy tests

All `Polly.Core.Tests` pass; branch and method coverage remain at 100% and the new code is fully covered.

## Confirm the following

### DIFF
--- a/src/Polly.Core/Utils/OutcomeUtilities.cs
+++ b/src/Polly.Core/Utils/OutcomeUtilities.cs
@@ -12,7 +12,8 @@ internal static class OutcomeUtilities
     /// <param name="callerToken">The cancellation token that was associated with the execution before the strategy substituted it.</param>
     /// <returns>
     /// An outcome whose <see cref="OperationCanceledException"/> carries <paramref name="callerToken"/> when the caller
-    /// requested cancellation; otherwise the original <paramref name="outcome"/> unchanged.
+    /// requested cancellation, preserving the original exception as its <see cref="Exception.InnerException"/>;
+    /// otherwise the original <paramref name="outcome"/> unchanged.
     /// </returns>
     /// <remarks>
     /// The rewrite happens only when <paramref name="callerToken"/> actually requested cancellation. This preserves
@@ -23,7 +24,7 @@ internal static class OutcomeUtilities
     {
         if (callerToken.IsCancellationRequested && outcome.Exception is OperationCanceledException oce && oce.CancellationToken != callerToken)
         {
-            return Outcome.FromException<T>(new OperationCanceledException(callerToken).TrySetStackTrace());
+            return Outcome.FromException<T>(new OperationCanceledException(oce.Message, oce, callerToken).TrySetStackTrace());
         }
 
         return outcome;
--- a/src/Polly.Core/Utils/OutcomeUtilities.cs
+++ b/src/Polly.Core/Utils/OutcomeUtilities.cs
@@ -1,0 +1,31 @@
+namespace Polly.Utils;
+
+internal static class OutcomeUtilities
+{
+    /// <summary>
+    /// Ensures that an <see cref="OperationCanceledException"/> escaping a strategy that substituted the
+    /// execution <see cref="CancellationToken"/> (e.g. timeout or hedging) carries the caller's token when
+    /// the cancellation was caused by that token.
+    /// </summary>
+    /// <typeparam name="T">The result type of the outcome.</typeparam>
+    /// <param name="outcome">The outcome produced by the strategy.</param>
+    /// <param name="callerToken">The cancellation token that was associated with the execution before the strategy substituted it.</param>
+    /// <returns>
+    /// An outcome whose <see cref="OperationCanceledException"/> carries <paramref name="callerToken"/> when the caller
+    /// requested cancellation; otherwise the original <paramref name="outcome"/> unchanged.
+    /// </returns>
+    /// <remarks>
+    /// The rewrite happens only when <paramref name="callerToken"/> actually requested cancellation. This preserves
+    /// the contract that a real timeout, or an unrelated <see cref="OperationCanceledException"/> thrown while the
+    /// caller's token was not cancelled, is left untouched.
+    /// </remarks>
+    public static Outcome<T> WithCallerCancellationToken<T>(this Outcome<T> outcome, CancellationToken callerToken)
+    {
+        if (callerToken.IsCancellationRequested && outcome.Exception is OperationCanceledException oce && oce.CancellationToken != callerToken)
+        {
+            return Outcome.FromException<T>(new OperationCanceledException(callerToken).TrySetStackTrace());
+        }
+
+        return outcome;
+    }
+}
--- a/test/Polly.Core.Tests/Issues/IssuesTests.CancellationTokenPropagation_3086.cs
+++ b/test/Polly.Core.Tests/Issues/IssuesTests.CancellationTokenPropagation_3086.cs
@@ -77,12 +77,17 @@ public partial class IssuesTests
     [Fact]
     public async Task NestedTimeouts_CallerCancellation_ExceptionCarriesCallerToken_3086()
     {
-        var inner = new ResiliencePipelineBuilder()
+        var innermost = new ResiliencePipelineBuilder()
             .AddTimeout(TimeSpan.FromMinutes(1))
             .Build();
 
-        var pipeline = new ResiliencePipelineBuilder()
+        var inner = new ResiliencePipelineBuilder()
             .AddTimeout(TimeSpan.FromMinutes(2))
+            .AddPipeline(innermost)
+            .Build();
+
+        var pipeline = new ResiliencePipelineBuilder()
+            .AddTimeout(TimeSpan.FromMinutes(3))
             .AddPipeline(inner)
             .Build();
 
--- a/test/Polly.Core.Tests/Issues/IssuesTests.CancellationTokenPropagation_3086.cs
+++ b/test/Polly.Core.Tests/Issues/IssuesTests.CancellationTokenPropagation_3086.cs
@@ -22,7 +22,7 @@ public partial class IssuesTests
             pipeline.ExecuteAsync(
                 async token =>
                 {
-                    await cts.CancelAsync(); // simulate cancellation request from upstream caller
+                    cts.Cancel(); // simulate cancellation request from upstream caller
                     token.ThrowIfCancellationRequested(); // simulate cancellation response from downstream code
                 },
                 cts.Token).AsTask());
@@ -44,7 +44,7 @@ public partial class IssuesTests
             pipeline.ExecuteAsync(
                 async token =>
                 {
-                    await cts.CancelAsync();
+                    cts.Cancel();
                     token.ThrowIfCancellationRequested();
                 },
                 cts.Token).AsTask());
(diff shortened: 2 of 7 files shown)


src/Polly.Core/Utils/OutcomeUtilities.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// <remarks>
+    /// The rewrite happens only when <paramref name="callerToken"/> actually requested cancellation. This preserves
+    /// the contract that a real timeout, or an unrelated <see cref="OperationCanceledException"/> thrown while the
+    /// caller's token was not cancelled, is left untouched.
+    /// </remarks>
+    public static Outcome<T> WithCallerCancellationToken<T>(this Outcome<T> outcome, CancellationToken callerToken)
+    {
+        if (callerToken.IsCancellationRequested && outcome.Exception is OperationCanceledException oce && oce.CancellationToken != callerToken)
+        {
+            return Outcome.FromException<T>(new OperationCanceledException(callerToken).TrySetStackTrace());
```
martincostello: I realise this was called out in the PR description, but I wonder if its still better to contain the inner exception and preserve it with [`.ctor(String, Exception, CancellationToken)`](https://learn.microsoft.com/dotnet/api/system.operationcanceledexception.-ctor#system-operationcanceledexception-ctor(system-string-system-exception-system-threading-cancellationtoken)):

```suggestion
            return Outcome.FromException<T>(new OperationCanceledException(outcome.Message, outcome.Exception, callerToken).TrySetStackTrace());
```

test/Polly.Core.Tests/Issues/IssuesTests.CancellationTokenPropagation_3086.cs
```diff
@@ -0,0 +1,197 @@
...
+                    await cts.CancelAsync();
+                    token.ThrowIfCancellationRequested();
+                },
+                cts.Token).AsTask());
+
+        exception.CancellationToken.ShouldBe(cts.Token);
+    }
+
+    [Fact]
+    public async Task NestedTimeouts_CallerCancellation_ExceptionCarriesCallerToken_3086()
```
martincostello, line 78: Just to ensure it's "arbitrary" nesting rather than just one level, could we add one more level of nesting to this test?

martincostello (commented):
This looks a lot less involved than I thought it might have been.

Just a few comments.

martincostello:
CI failure is fixed by #3096.

martincostello:
Some test fixes needed for `net481` that doesn't have `CancelAsync()`.

DaRosenberg:
Yeah just noticed, didn't know that was still a target — will fix.

DaRosenberg:
Didn't think it was worth doing the whole `#if NET8_0_OR_GREATER` in test code — let me know if you think that's worth it and I can add it.

martincostello (approved):

PR #3135, opened 2026-07-02
Add support for getting/setting the retry attempt number on the context

### LINKED ISSUE #3095: [Feature request]: Allow setting the retry attempt

The problem I'm facing is that I'm using a Polly retry strategy for running a retriable action, but in a durable way: I'm storing the number of previous attempts, and the process can exit in the middle of the Nth attempt. When that happens, and execution resumes in a different process, I want to be able to run the same resiliency pipeline, but tell it to start from attempt N so that a recurring error that might crash the process doesn't make the pipeline start from the first attempt every time and go into an infinite retry loop.

The cheapest solution would be to add an optional property to `RetryStrategyOptions`:
```cs
    public ResiliencePropertyKey<uint>? CurrentAttemptPropertyKey { get; set; }
```

When this property is set on the context, `RetryResilienceStrategy` should read the attempt number from this property instead of [initializing it to `0`](https://github.com/App-vNext/Polly/blob/8d978129b6569b69241b32f90d39cd97f822d7ce/src/Polly.Core/Retry/RetryResilienceStrategy.cs#L50).

### Additional context

_No response_

BalassaMarton:
Added `RetryResilienceContextExtensions` with methods for getting/setting the current retry attempt number.
Changed `RetryResilienceStrategy` to respect and update this property on the context.
Closes #3095

### DIFF
--- a/src/Polly.Core/Retry/RetryResilienceContextExtensions.cs
+++ b/src/Polly.Core/Retry/RetryResilienceContextExtensions.cs
@@ -1,0 +1,45 @@
+namespace Polly;
+
+/// <summary>
+///   Extension methods for <see cref="ResilienceContext"/> to store and retrieve properties related to the Retry strategy.
+/// </summary>
+public static class RetryResilienceContextExtensions
+{
+    internal static readonly ResiliencePropertyKey<int> AttemptNumberPropertyKey = new("Retry.AttemptNumber");
+
+    /// <summary>
+    /// Gets the current attempt number from the <see cref="ResilienceContext"/> (starting with <c>0</c> for the first attempt).
+    /// </summary>
+    /// <param name="context">The <see cref="ResilienceContext"/> instance.</param>
+    /// <returns>The current attempt number.</returns>
+    public static int GetRetryAttemptNumber(this ResilienceContext context)
+    {
+        Guard.NotNull(context);
+
+        return context.Properties.TryGetValue(AttemptNumberPropertyKey, out var value)
+            ? value
+            : 0;
+    }
+
+    /// <summary>
+    /// Sets the attempt number in the <see cref="ResilienceContext"/>.
+    /// </summary>
+    /// <param name="context">The <see cref="ResilienceContext"/> instance.</param>
+    /// <param name="attemptNumber">The attempt number to set.</param>
+    /// <remarks>
+    /// Application code can use this method to set the initial attempt number when executing a retry strategy.
+    /// The strategy will update the attempt number on each retry, and it can be retrieved using <see cref="GetRetryAttemptNumber"/>.
+    /// The strategy will overwrite the attempt number on each retry, so any value set inside callbacks may be lost or cause inconsistent behavior.
+    /// </remarks>
+    public static void SetRetryAttemptNumber(this ResilienceContext context, int attemptNumber)
+    {
+        Guard.NotNull(context);
+
+        if (attemptNumber < 0)
+        {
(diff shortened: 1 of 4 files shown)


src/Polly.Core/Retry/RetryResilienceContextExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// <param name="context">The <see cref="ResilienceContext"/> instance.</param>
+    /// <param name="attemptNumber">The attempt number to set.</param>
+    /// <remarks>
+    /// Application code can use this method to set the initial attempt number when executing a retry strategy.
+    /// The strategy will update the attempt number on each retry, and it can be retrieved using <see cref="GetRetryAttemptNumber"/>.
+    /// The strategy will overwrite the attempt number on each retry, so any value set inside callbacks may be lost or cause inconsistent behavior.
+    /// </remarks>
+    public static void SetRetryAttemptNumber(this ResilienceContext context, int attemptNumber)
+    {
+        Guard.NotNull(context);
```
martincostello, line 36: We should validate `attemptNumber >= 0` and throw if not.

src/Polly.Core/Retry/RetryResilienceStrategy.cs
```diff
@@ -47,7 +47,7 @@ protected internal override async ValueTask<Outcome<T>> ExecuteCore<TState>(Func
     {
         double retryState = 0;
 
-        int attempt = 0;
+        var attempt = context.GetRetryAttemptNumber();
```
martincostello: Enforce non-negative values in case a user messes with the context.

```suggestion
        var attempt = Math.Max(0, context.GetRetryAttemptNumber());
```

src/Polly.Core/Retry/RetryResilienceStrategy.cs
```diff
@@ -131,6 +131,8 @@ protected internal override async ValueTask<Outcome<T>> ExecuteCore<TState>(Func
             {
                 attempt++;
```
martincostello, line 132: This could potentially overflow now as a user could set an arbitrarily high value.

BalassaMarton: That's already handled by `IsLastAttempt`.

src/Polly.Core/Retry/RetryResilienceContextExtensions.cs
BalassaMarton: Fixed in latest push

src/Polly.Core/Retry/RetryResilienceStrategy.cs
BalassaMarton: Fixed in latest push

src/Polly.Core/Retry/RetryResilienceContextExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+        return context.Properties.TryGetValue(AttemptNumberPropertyKey, out var value)
+            ? value
+            : 0;
+    }
+
+    /// <summary>
+    /// Sets the attempt number in the <see cref="ResilienceContext"/>.
+    /// </summary>
+    /// <param name="context">The <see cref="ResilienceContext"/> instance.</param>
+    /// <param name="attemptNumber">The attempt number to set.</param>
```
martincostello:
```suggestion
    /// <param name="attemptNumber">The non-negative attempt number to set.</param>
```

martincostello:
https://github.com/App-vNext/Polly/issues/3095#issuecomment-5507142811

PR #3140, opened 2026-07-03
feat(extensions): add EnableReloads overload accepting custom IOptionsMonitor

### LINKED ISSUE #3050: [Feature request]: Expose `EnableReloads` overload that accepts a custom `IOptionsMonitor<TOptions>` instance

Currently, `AddResiliencePipelineContext.EnableReloads<TOptions>()` only supports enabling reloads via an `IOptionsMonitor<TOptions>` that is **resolved from DI**:

```csharp
RegistryContext.EnableReloads(
    ServiceProvider.GetRequiredService<IOptionsMonitor<TOptions>>(),
    name);
```

This makes it impossible to enable reloads for **custom or external `IOptionsMonitor<TOptions>` implementations** that are **not registered in DI** as `IOptionsMonitor<TOptions>`.

This limitation prevents several legitimate and advanced scenarios, such as:

*   Using a custom options system not based on `Microsoft.Extensions.Options`
*   Wrapping another configuration source (feature flags, remote config, tenant-specific config, etc.)
*   Sharing a single monitor instance across multiple pipelines without DI registration
*   Avoiding DI pollution when the monitor is internal or pipeline-specific

Expose an additional public overload that allows explicitly supplying an `IOptionsMonitor<TOptions>` instance:

```csharp
public void EnableReloads<TOptions>(
    IOptionsMonitor<TOptions> monitor,
    string? name = null)
    where TOptions : class;
```

This would internally forward to the existing infrastructure currently wired via `RegistryContext.EnableReloads(...)`.

Expected behavior:

*   The pipeline is rebuilt when `monitor.OnChange` fires
*   `name` continues to behave the same way as in the existing overload
*   No change in behavior for existing users

This overload would be purely additive and backward-compatible.

### Additional context

_No response_

arashzjahangiri:
closes #3050 

Adds a new public overload of AddResiliencePipelineContext<TKey>.EnableReloads<TOptions> that accepts a caller-supplied IOptionsMonitor<TOptions> directly, instead of resolving it from the DI container.

The underlying ConfigureBuilderContextExtensions.EnableReloads already accepts the monitor directly - this change surfaces that path through the public API.

Use cases enabled:
- Custom options systems outside Microsoft.Extensions.Options
- Wrapping alternative configuration sources (feature flags, remote configs)
- Sharing monitor instances across pipelines without DI registration

## Confirm the following

### DIFF
--- a/src/Polly.Extensions/DependencyInjection/AddResiliencePipelineContext.cs
+++ b/src/Polly.Extensions/DependencyInjection/AddResiliencePipelineContext.cs
@@ -2,6 +2,9 @@ using System.Diagnostics.CodeAnalysis;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 using Polly.Registry;
+using Polly.Utils;
+
+#pragma warning disable RS0026 // Do not add multiple public overloads with optional parameters
 
 namespace Polly.DependencyInjection;
 
@@ -47,6 +50,27 @@ public sealed class AddResiliencePipelineContext<TKey>
     /// </remarks>
     public void EnableReloads<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] TOptions>(string? name = null)
         => RegistryContext.EnableReloads(ServiceProvider.GetRequiredService<IOptionsMonitor<TOptions>>(), name);
+
+    /// <summary>
+    /// Enables dynamic reloading of the resilience pipeline whenever the options produced by <paramref name="monitor"/> are changed.
+    /// </summary>
+    /// <typeparam name="TOptions">The options type to listen to.</typeparam>
+    /// <param name="monitor">The options monitor to listen to for changes.</param>
+    /// <param name="name">The named options, if any.</param>
+    /// <remarks>
+    /// Use this overload when the <see cref="IOptionsMonitor{TOptions}"/> instance is not registered in the
+    /// dependency injection container (for example, when wrapping a custom configuration source or feature flag system).
+    /// <para>
+    /// You can listen for changes from multiple options by calling this method with different <typeparamref name="TOptions"/> types.
+    /// </para>
+    /// </remarks>
+    public void EnableReloads<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] TOptions>(
+        IOptionsMonitor<TOptions> monitor,
+        string? name = null)
+    {
+        Guard.NotNull(monitor);
+        RegistryContext.EnableReloads(monitor, name);
+    }
 
     /// <summary>
     /// Gets the options identified by <paramref name="name"/>.
--- a/src/Polly.Extensions/DependencyInjection/AddResiliencePipelineContext.cs
+++ b/src/Polly.Extensions/DependencyInjection/AddResiliencePipelineContext.cs
@@ -4,8 +4,6 @@ using Microsoft.Extensions.Options;
 using Polly.Registry;
 using Polly.Utils;
 
-#pragma warning disable RS0026 // Do not add multiple public overloads with optional parameters
-
 namespace Polly.DependencyInjection;
 
 /// <summary>
@@ -48,8 +46,26 @@ public sealed class AddResiliencePipelineContext<TKey>
     /// You can listen for changes from multiple options by calling this method with different <typeparamref name="TOptions"/> types.
     /// </para>
     /// </remarks>
+#pragma warning disable RS0027 // API with optional parameter(s) should have the most parameters amongst its public overloads
     public void EnableReloads<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] TOptions>(string? name = null)
         => RegistryContext.EnableReloads(ServiceProvider.GetRequiredService<IOptionsMonitor<TOptions>>(), name);
+#pragma warning restore RS0027
+
+    /// <summary>
+    /// Enables dynamic reloading of the resilience pipeline whenever the options produced by <paramref name="monitor"/> are changed.
+    /// </summary>
+    /// <typeparam name="TOptions">The options type to listen to.</typeparam>
+    /// <param name="monitor">The options monitor to listen to for changes.</param>
+    /// <remarks>
+    /// Use this overload when the <see cref="IOptionsMonitor{TOptions}"/> instance is not registered in the
+    /// dependency injection container (for example, when wrapping a custom configuration source or feature flag system).
+    /// <para>
+    /// You can listen for changes from multiple options by calling this method with different <typeparamref name="TOptions"/> types.
+    /// </para>
+    /// </remarks>
+    public void EnableReloads<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] TOptions>(
+        IOptionsMonitor<TOptions> monitor)
+        => EnableReloads(monitor, name: null);
 
     /// <summary>
     /// Enables dynamic reloading of the resilience pipeline whenever the options produced by <paramref name="monitor"/> are changed.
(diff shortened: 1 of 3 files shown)


src/Polly.Extensions/DependencyInjection/AddResiliencePipelineContext.cs
```diff
@@ -2,6 +2,9 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 using Polly.Registry;
+using Polly.Utils;
+
+#pragma warning disable RS0026 // Do not add multiple public overloads with optional parameters
```
martincostello: Please rework the API to remove the need to suppress this warning.

src/Polly.Extensions/.PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -1,7 +1,6 @@
 ﻿#nullable enable
 abstract Polly.Telemetry.MeteringEnricher.Enrich<TResult, TArgs>(in Polly.Telemetry.EnrichmentContext<TResult, TArgs> context) -> void
 Polly.DependencyInjection.AddResiliencePipelineContext<TKey>
-Polly.DependencyInjection.AddResiliencePipelineContext<TKey>.EnableReloads<TOptions>(string? name = null) -> void
```
martincostello, line 4: This is a breaking change.

src/Polly.Extensions/DependencyInjection/AddResiliencePipelineContext.cs
arashzjahangiri: Reworked the API to avoid the RS0026 suppression entirely. The new `IOptionsMonitor` overload no longer has an optional `name` parameter — instead there are two explicit overloads: one with just the monitor, and one with both monitor and name. The original `EnableReloads<TOptions>(string? name = null)` remains unchanged (no breaking change). The RS0027 that appears on the original overload (a consequence of adding a 2-param sibling) is suppressed with a targeted pragma/restore pair on that method only, consistent with the pattern already used in Polly.Core.

src/Polly.Extensions/.PublicAPI/PublicAPI.Shipped.txt
arashzjahangiri: I reverted this change.

src/Polly.Extensions/DependencyInjection/AddResiliencePipelineContext.cs
```diff
@@ -45,8 +46,47 @@ internal AddResiliencePipelineContext(ConfigureBuilderContext<TKey> registryCont
     /// You can listen for changes from multiple options by calling this method with different <typeparamref name="TOptions"/> types.
     /// </para>
     /// </remarks>
+#pragma warning disable RS0027 // API with optional parameter(s) should have the most parameters amongst its public overloads
```
martincostello: Is there a way you can add this new API without _any_ warning suppressions being required?

arashzjahangiri: The issue is RS0027: the analyzer requires that if an overload has optional parameters, it must be the one with the most parameters across all overloads. 


  There are two options:

  Option 1 — keep the targeted suppression (current state)
  No breaking change, matches the pattern already used in Polly.Core.

  Option 2 — zero suppressions, binary-breaking
  Remove = null from the existing overload, add an explicit no-arg overload to keep source compatibility, and move the optional param to the new 2-param overload:


```
  // Explicit no-arg overload so existing callers with no arguments still compile
  public void EnableReloads<TOptions>()

  // Was (string? name = null) — now required param, binary-breaking but source-compatible
  public void EnableReloads<TOptions>(string? name)

  // Optional param lives here, on the overload with the most parameters — satisfies RS0026 and RS0027
  public void EnableReloads<TOptions>(IOptionsMonitor<TOptions> monitor, string? name = null)

````

What do you think?

Comment (martincostello) [reply src/Polly.Extensions/DependencyInjection/AddResiliencePipelineContext.cs]: Option 3 - maybe a different method name to avoid the conflict?

Comment (arashzjahangiri) [reply src/Polly.Extensions/DependencyInjection/AddResiliencePipelineContext.cs]: Done!
Went with `EnableReloadsWithMonitor<TOptions>`. Single method, no overloads, no suppressions needed. 

Any other concerns?

File: src/Polly.Extensions/DependencyInjection/AddResiliencePipelineContext.cs [hunk 4]
```diff
@@ -48,6 +49,27 @@ internal AddResiliencePipelineContext(ConfigureBuilderContext<TKey> registryCont
     public void EnableReloads<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)] TOptions>(string? name = null)
         => RegistryContext.EnableReloads(ServiceProvider.GetRequiredService<IOptionsMonitor<TOptions>>(), name);
```
martincostello: This could call the new method.

test/Polly.Extensions.Tests/ReloadableResiliencePipelineTests.cs
```diff
@@ -95,6 +96,66 @@ public void AddResiliencePipeline_EnsureReloadable(string? name)
         }
     }
 
+    [InlineData(null)]
+    [InlineData("custom-name")]
+    [InlineData("")]
+    [Theory]
```
martincostello:
```suggestion
    [Theory]
    [InlineData(null)]
    [InlineData("")]
    [InlineData("custom-name")]
```

test/Polly.Extensions.Tests/ReloadableResiliencePipelineTests.cs
```diff
@@ -95,6 +96,66 @@ public void AddResiliencePipeline_EnsureReloadable(string? name)
...
+
+        pipeline.Execute(_ => "dummy", ctx);
+        ctx.Properties.GetValue(TagKey, string.Empty).ShouldBe("reloaded-tag");
+
+        resList.Count.ShouldBe(2);
+        resList[0].Received(1).Dispose();
+        resList[1].Received(0).Dispose();
+
+        serviceProvider.Dispose();
+        resList[1].Received(1).Dispose();
```
martincostello, line 142: Verify that `[0]` doesn't?

test/Polly.Extensions.Tests/ReloadableResiliencePipelineTests.cs
```diff
@@ -95,6 +96,66 @@ public void AddResiliencePipeline_EnsureReloadable(string? name)
...
+
+    [Fact]
+    public void EnableReloadsWithMonitor_NullMonitor_Throws()
+    {
+        var services = new ServiceCollection();
+        services.AddResiliencePipeline("my-pipeline", (_, context) =>
+        {
+            Assert.Throws<ArgumentNullException>("monitor",
+                () => context.EnableReloadsWithMonitor((IOptionsMonitor<ReloadableStrategyOptions>)null!));
+        });
```
martincostello, line 155: Nothing here asserts that this callback was invoked.

src/Polly.Extensions/DependencyInjection/AddResiliencePipelineContext.cs
arashzjahangiri: Done!

test/Polly.Extensions.Tests/ReloadableResiliencePipelineTests.cs
arashzjahangiri: Added!

test/Polly.Extensions.Tests/ReloadableResiliencePipelineTests.cs
arashzjahangiri: Added a `called` flag, asserted `called.ShouldBeTrue()` after `GetPipeline`.

martincostello (approved):

PR #3192, opened 2026-08-11
Update to .NET 11 preview 7

martincostello:
Update to preview 7 of .NET 11.


martincostello:
Looks like the mutation tests are extra-busted with 11.0-preview.7 - will merge as-is and investigate tomorrow.

